PR #14, opened 2018-08-09
Adds a default Make task that builds locally in release mode

### LINKED ISSUE #13: Makefile default should build for my platform
When I ran `make`, it started running a release build for Linux on my Mac ⁉️ 

It's because that task is simply the first in the Makefile.

I'd expect `make` alone to build whatever is appropriate for my system.

colindean:
Fixes #13

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,15 @@
+.PHONY: all
+local:
+	cargo build --release
+	strip target/release/miniserve
+	upx target/release/miniserve
+
+run:
+ifndef ARGS
+	@echo Run "make run" with ARGS set to pass arguments…
+endif
+	cargo run --release -- $(ARGS)
+
 .PHONY: build-linux
 build-linux:
 	cargo build --target x86_64-unknown-linux-musl --release
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
-.PHONY: all
+.PHONY: local
 local:
 	cargo build --release
 
+.PHONY: run
 run:
 ifndef ARGS
 	@echo Run "make run" with ARGS set to pass arguments…
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,6 @@
 .PHONY: all
 local:
 	cargo build --release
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,3 +1,15 @@
+.PHONY: all
+local:
+	cargo build --release
+	strip target/release/miniserve
+	upx target/release/miniserve
```
svenstaro: I suppose we shouldn't have that for local stuff and neither the strip.

colindean: That's fair. Stripping would be a logical part of a `make install`, though. I don't want to get into setting up the stuff for `$PREFIX` to do that.

Makefile
```diff
@@ -1,3 +1,13 @@
+.PHONY: all
```
svenstaro: Shouldn't the `.PHONY` be `local` instead of `all`?

colindean: Yes, but it's also better to have all phony tasks under one `.PHONY`, so I did that, too.

Makefile
```diff
@@ -1,17 +1,25 @@
...
 	strip target/x86_64-pc-windows-gnu/release/miniserve.exe
 	upx target/x86_64-pc-windows-gnu/release/miniserve.exe
 
-.PHONY: build-apple
 build-apple:
 	cargo build --target x86_64-apple-darwin --release
 	strip target/x86_64-apple-darwin/release/miniserve
 	upx target/x86_64-apple-darwin/release/miniserve
+
+.PHONY: local run build-apple build-win build-linux
```
svenstaro: I try to conform to Clark Grubb's Makefile style guide. See [here](http://clarkgrubb.com/makefile-style-guide#phony-targets).

colindean: Ah, OK. I'll revert my commit but I'll have to wait until I'm done with some other work to do tonight. I wish GitHub's interface had a revert 😩

colindean:
Fixed.

svenstaro:
Great stuff, thanks!

Issue #4, opened 2018-05-17
Add option to not follow symlinks


svenstaro:
Currently, miniserve follows symlinks. There should be an option to not have it do that.

PR #19, opened 2018-11-06
add option to ignore symlinks

pebbleKite:
This is an attempt to address #4


pebbleKite:
#4

svenstaro:
Hi, thanks a lot! I had a look and a think and I now think that this is going the right way. However, we might want to go all the way with this. You implemented a custom directory rendering function for the version without symlinks. How about you restructure this in such a way that we always use a custom rendering function (as this would also allow us to make other changes such as changing the styling).

This way, you could then easily add a condition in that rendering function in case symlinks should be excluded.

Makes sense?

pebbleKite:
The last commit (8034d4b2604dd3f5dc97c95e7da8e302e0044068) should address the single renderer.

src/main.rs
```diff
@@ -127,9 +133,15 @@ pub fn parse_args() -> MiniserveConfig {
                 .validator(is_valid_auth)
                 .help("Set authentication (username:password)")
                 .takes_value(true),
+        ).arg(
+            Arg::with_name("no-symlinks")
+                .short("P")
+                .long("no-symlinks")
+                .help("Do not follow symbolic symbolic links"),
```
svenstaro: Double `symbolic`

pebbleKite: Sorry about that. fc46689 should fix that.

svenstaro:
Thanks for the clean-up, will take a look soon!

svenstaro:
Looks neat, thanks a lot.

PR #26, opened 2019-02-02
Support for random URL

boastful-squirrel:
Sometimes, I think it's quite convenient to generate random URL, so only people with the right one can crawl the server. Let's be honest, it's very cheap security-wise, but I think there could be some use-cases where people would want this feature.

I'm talking about something like this:

``` 
miniserve -i 192.168.0.1 -r /tmp/myshare
# Serving path /tmp/myshare at http://192.168.0.1:8080/ibe1CYIYqIOuCQ_0lkpWM
```

Currently, the option is `-r  / --random-route`, maybe there is a better choice than this, as `-r` makes me think of "recursive", I might be wrong though.

If this feature isn't wanted, I'll close the PR.


boastful-squirrel:
On it :)

boastful-squirrel:
I also removed the `-r` short option

README.md
```diff
@@ -25,6 +25,11 @@ Sometimes this is just a more practical and quick way than doing things properly
 
     miniserve --auth joe:123 unreleased-linux-distros/
 
+### Generate a random route so only people with the right URL can crawl the server:
+
+    miniserve -i 192.168.0.1 -r /tmp/myshare
```
svenstaro: Gotta fix this invocation.

README.md
```diff
@@ -25,6 +25,11 @@ Sometimes this is just a more practical and quick way than doing things properly
...
+    # Serving path /tmp/myshare at http://192.168.0.1:8080/ibe1CYIYqIOuCQ_0lkpWM
```
svenstaro: Also this.

mglax:
Yep, I'll do it  asap. Also, I failed with Git it seems, as all changes done on the fancy-listing branches also appear here. I'll fix this asap

src/main.rs
```diff
@@ -164,6 +177,12 @@ pub fn parse_args() -> MiniserveConfig {
         None
     };
 
+    let mut random_route = None;
```
svenstaro: Can you make this a bit like https://github.com/svenstaro/miniserve/pull/26/files#diff-639fbc4ef05b315af92b4d836c31b023R158?

This way, you don't have the mutable variable around.

src/main.rs
```diff
@@ -136,12 +142,19 @@ pub fn parse_args() -> MiniserveConfig {
                 .long("no-symlinks")
                 .help("Do not follow symbolic links"),
         )
+        .arg(
+            Arg::with_name("random-route")
```
svenstaro: This should conflict with an explicitly chosen route.

src/main.rs
```diff
@@ -198,10 +218,14 @@ fn configure_app(app: App<MiniserveConfig>) -> App<MiniserveConfig> {
         }
     };
 
+    let route_url = "/".to_string();
```
svenstaro: Maybe call this `base_route` to communicate the intention?

svenstaro: And in fact, you can probably just remove this as you don't really need it. Just put it directly in your format string at `let full_route`.

boastful-squirrel:
I'll look into that, thanks for the advice :)

boastful-squirrel:
I'm closing this PR as I messed up hard with git, opening a new one with all requestes changes

Issue #28, opened 2019-02-02
Allow different sort modes for file listing


svenstaro:
It would be nice to be able to sort the files differently. Currently, the ordering feels kind of wrong as the dirs are usually followed by the files but currently everything is only ordered alphabetically. It would be nice to get different order types. Maybe these order modes:

**dirsfiles**
```
dir1/
dir2/
dir3/
file1
file2
file3
```

**alpha**
```
A/
a
B
b/
C
c
```

**natural**
```
A/
B
C
a
b
c
```

boastful-squirrel:
I'll be on it. Should we keep alphabetical order as default ?

svenstaro:
I think it makes the most sense to order it like other web server listings
do: folders first and first, both sorted in alpha.

boastful-squirrel:
Got it

boastful-squirrel:
What should I do for the CLI ? Different options conflicting with eachothers, or a `sort` option that can take 3 values ? 

I mean, something like this:


```
miniserve --sort-foobar /tmp
```

or 
```
miniserve --sort=foobar /tmp
```

boastful-squirrel:
So I implemented dirsfiles and alpha, what's "natural" exactly ? In your example, it looks like dirsfiles :P

svenstaro:
Yeah I suppose that's a bit misleading what I wrote there. So the general
idea was to have either ASCII value based ordering or alphabetical ordering
whereby you would get ABCabc (ASCII) as opposed to AaBbCc (alpha).

Natural sorting actually just changes how you treat numbers:
https://en.m.wikipedia.org/wiki/Natural_sort_order

PR #30, opened 2019-02-04
Implemented sorting

boastful-squirrel:
Here is my attempt to solve #28. I didn't know if I should let a possibility for the user not to sort the result, though it didn't really make sense to me. By default, the sort is set to `natural`. 

If there's anything to improve the quality of the code, I'm open. Cheers !


svenstaro:
Honestly, this is pretty sweet. However, I'm thinking perhaps it's time to break up this program into some more files as the main.rs is getting pretty big. It's certainly not part of this PR but we should think about that.

PR #32, opened 2019-02-12
Split project into multiple files

boastful-squirrel:
Here is my attempt to split the project into multiple files. If you're not OK with the file hierarchy, please tell me, I can adapt it to your needs :)


src/config.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub interfaces: Vec<IpAddr>,
+    pub auth: Option<auth::BasicAuthParams>,
+    pub path_explicitly_chosen: bool,
+    pub no_symlinks: bool,
+    pub random_route: Option<String>,
+    pub sort_method: listing::SortingMethods,
+    pub reverse_sort: bool,
+}
+
+pub fn configure_app(app: App<MiniserveConfig>) -> App<MiniserveConfig> {
```
svenstaro: I think this particular function should stay in main.rs because main.rs essentially configures the whole actix app, right?

src/listing.rs
```diff
@@ -0,0 +1,252 @@
...
+                    entry.link,
+                    entry.name,
+                    entry.size.unwrap()
+                );
+            }
+        }
+    }
+
+    let html = format!(
+        "<html>\
```
svenstaro: Side note: What do you think about using `typed-html` here instead of the giant raw string? Not that we _need_ JSX but the typed stuff would be nice.

svenstaro:
Overall this looks pretty good with some minor changes!

Since we're already at this, would you mind adding proper docstrings to everything? :P

boastful-squirrel:
Sure, I'll do the changes accordingly. Should I also include #33 or do you prefer a separate PR?

src/listing.rs
boastful-squirrel: That's a cool idea, would be easier to maintain. Should I include this on this PR ?

src/config.rs
boastful-squirrel: Sure, I can do that. In this case, I think I will remove config.rs, otherwise it would just contain the `MiniserveConfig` struct (which will I would put back in main.rs), what do you think ?

boastful-squirrel:
I warn you, I'm bad at writing docstrings :)

src/listing.rs
svenstaro: Nah, let's make it a new one. Keep this one nice and tight.

PR #35, opened 2019-02-14
Switched to structopt

boastful-squirrel:
So here's my attempt at switching from `clap` to `structopt` :)


README.md
```diff
@@ -27,7 +27,7 @@ Sometimes this is just a more practical and quick way than doing things properly
 
 ### Generate random 6-hexdigit URL:
 
-    miniserve --random-route -i 192.168.0.1 /tmp
+    miniserve --random-route -i 192.168.0.1 -- /tmp
```
svenstaro: Aw, this is ugly, can we get rid of the `--` in this case somehow?

src/args.rs
```diff
@@ -1,173 +1,106 @@
...
-    ))
-}
-
-/// Checks wether a port is valid
-fn is_valid_port(port: String) -> Result<(), String> {
-    port.parse::<u16>()
-        .and(Ok(()))
-        .or_else(|e| Err(e.to_string()))
+#[derive(StructOpt, Debug)]
+#[structopt(name = "miniserve")]
```
svenstaro: I suggest adding `raw(global_settings = "&[AppSettings::ColoredHelp]")` to make the help nicer.

src/args.rs
```diff
@@ -1,173 +1,106 @@
...
+struct CLIArgs {
+    /// Be verbose, includes emitting access logs
+    #[structopt(short = "v", long = "verbose")]
+    verbose: bool,
+    /// Which path to serve
```
svenstaro: Let's space the arguments vertically to make them neater to read. Currently, this is like a big block. Let's have
```
/// docstring
#[structopt(stuff)]
verbose: bool,

/// docstring
...
```

src/main.rs
```diff
@@ -1,3 +1,7 @@
+extern crate structopt;
+#[macro_use]
+extern crate clap;
```
svenstaro: I think we can eliminate these stay rust 2018 by importing the macros directly.

README.md
svenstaro: It should be possible as their example has a similar case and they don't need `--`: https://github.com/TeXitoi/structopt#example

src/listing.rs
```diff
@@ -6,26 +6,19 @@ use std::cmp::Ordering;
 use std::fmt::Write as FmtWrite;
 use std::io;
 use std::path::Path;
-use std::str::FromStr;
 
-#[derive(Clone, Copy, Debug)]
-/// Available sorting methods
-pub enum SortingMethods {
-    /// Natural sorting method 
-    /// 1 -> 2 -> 3 -> 11
```
svenstaro: I liked the docstrings. Can we have them back?

svenstaro:
All in all I really like this as it removes about as much boilerplate as I hoped it would. There are some minor remarks.

boastful-squirrel:
So, I still need to solve 2 problems you raised:

- use of extern crate, right now I'm struggling with it, seems like it's not that simple to remove
- the `--` ugly duck, I don't know why `structOpt` tries to parse the final argument (the path) as interface...

src/main.rs
svenstaro: I solved this here already: https://github.com/svenstaro/overpower/blob/master/src/main.rs

svenstaro:
By the way, feel free to add yourself to the authors in Cargo.toml and such.

boastful-squirrel:
Thanks a lot, that's kind of you. 

I fixed the extern crate thingie, now let's find out why this works:

```
miniserve -i=127.0.0.1  -i=192.168.0.1 /
```

but this fails (need to precede the path with `--`):
```
miniserve -i 127.0.0.1  -i 192.168.0.1 /
```

boastful-squirrel:
Note that the example given in `structopt` documentation is a bit different than the bug we're observing. The example in the documentation uses this command:
```
./basic -o foo.txt -dvvvs 1337 -l alice -l bob --nb-cars 4 bar.txt baz.txt
```
so the `-l` option is followed by another option.

In our case, if I do that, it works too:
```
miniserve -i 127.0.0.1 -i 127.0.0.1 --sort=natural /tmp
```

However, this won't work:
```
miniserve -i 127.0.0.1 -i 127.0.0.1 /tmp
```

and I'm pretty sure it wouldn't work in their example either

svenstaro:
Oh that's an interesting observation. I didn't think of that. Perhaps we
should report that upstream as a bug and for time being work around it by
reordering the order of arguments in the readme? :P

boastful-squirrel:
I'll try to create a minimal failing example and report the issue... Maybe they have a solution for it :)

svenstaro:
Turns out we already have another `--` example in the REAMDE already. Oh well!

svenstaro:
It would generally be a really useful if we could tell structopt to only accept multiple things in the format `-i blah -i blah2` as opposed to `-i blah blah2`.

boastful-squirrel:
So we would need this https://docs.rs/clap/2.29.0/clap/struct.Arg.html#method.number_of_values

boastful-squirrel:
Oh, maybe I found it :)

svenstaro:
Excellent!

Issue #36, opened 2019-02-15
Show last modified date in listing


svenstaro:
I actually had this requirement a few minutes ago that I wanted to see the modified date in the file listing.

boastful-squirrel:
Should be pretty quick to do, since we already retrieve metadata of the file :) The only thing is: what format for the date should we use ?

boastful-squirrel:
Also, I think we should add a sorting option based on modification date too, while we are at it.

svenstaro:
Let's allow for setting a custom date formatting specifier. I'm sure
whatever Rust library we use to output the date can just directly eat a
format string and will also have some sane defaults such as ISO 8601 or RFC
3339 so we don't have to reinvent the wheel. Let's make it ISO by default
though since it's easy to read when it's sorted.

Also I like the idea of being able to sort by date, of course. :)

boastful-squirrel:
What do you think about using a crate for “humanizing” the duration ? I mean, instead of showing a date, something like “4 days ago”, “3 minutes and 12 seconds ago” would be great, I think

boastful-squirrel:
<img width="1339" alt="screenshot 2019-02-16 at 15 09 21" src="https://user-images.githubusercontent.com/47121073/52900925-debda180-31fc-11e9-9b47-9051d880a894.png">

Something like this :)

PR #38, opened 2019-02-16
Display last modification date in listing + provides a way to sort by modification date

boastful-squirrel:
This addresses #36 

If there's something I should improve, please tell me :)


boastful-squirrel:
screenshot 2019-02-17 at 00 05 20

boastful-squirrel:
screenshot 2019-02-17 at 09 18 31

I tried to mimick `ls -l`'s result when it comes to the format.

boastful-squirrel:
I'll update the README tonight ;)

src/listing.rs
```diff
@@ -260,3 +306,39 @@ pub fn directory_listing<S>(
...
+fn convert_to_utc(src_time: Option<SystemTime>) -> (String, String) {
+    match src_time {
+        Some(time) => {
+            let date_time = DateTime::<Utc>::from(time);
+            (
+                date_time.format("%e %b").to_string(),
+                date_time.format("%R").to_string(),
+            )
+        }
+        None => ("".to_string(), "".to_string()),
```
svenstaro: Can't this be done more simply using `unwrap_or_default()` or such?

src/listing.rs
```diff
@@ -260,3 +306,39 @@ pub fn directory_listing<S>(
...
+fn humanize_duration(src_time: Option<SystemTime>) -> String {
+    match src_time {
+        Some(std_time) => match SystemTime::now().duration_since(std_time) {
+            Ok(from_now) => match Duration::from_std(from_now) {
+                Ok(duration) => HumanTime::from(duration).to_text_en(Accuracy::Rough, Tense::Past),
+                Err(_) => "".to_string(),
+            },
+            Err(_) => "".to_string(),
+        },
+        None => "".to_string(),
```
svenstaro: Same as above, can't we do all of these default values using `unwrap_or_default()`?

boastful-squirrel: I'll try :)

src/listing.rs
boastful-squirrel: I'll try :)

src/listing.rs
```diff
@@ -260,3 +306,39 @@ pub fn directory_listing<S>(
...
+        None => ("".to_string(), "".to_string()),
+    }
+}
+
+/// Converts a SystemTime to a string readable by a human,
+/// i.e. calculates the duration between now() and the given SystemTime,
+/// and gives a rough approximation of the elapsed time since
+///
+/// If no SystemTime was given, returns an empty string
+fn humanize_duration(src_time: Option<SystemTime>) -> String {
```
svenstaro: Wrong name, I think, as this is not a duration but a point in time. So humanize_systemtime?

On a side note, wouldn't it be cool actually put a function like this into chrono_humanize itself?

boastful-squirrel: True, that would be a good addition to the crate

svenstaro:
Excellent work. The conversion still looks a little bit weird but I think it's better than before. I'll take it!

Issue #34, opened 2019-02-13
Break out HTML and use a templating engine


boastful-squirrel:
_Originally posted by @svenstaro in https://github.com/svenstaro/miniserve/pull/32_

Just creating the issue so we can remember it. I'll probably do it after #32 is merged and #33 is done

boastful-squirrel:
I also think we should split HTML/CSS and js. CSS is a static file, JS should be static too, so I suggest we put them in a separate file? Then we just call them from the HTML.

svenstaro:
Yes, I think you're right. It would be better to just break out the HTML and save it as normal HTML as opposed to using typed-html. That's not the right tool for this job at this point. Just separate out everything and use normal templating.

boastful-squirrel:
Do you have any recommendation for the template engine crate we should use ? I know `aksama` but it might be too big ? `maud` seems very promising too, though it requires nightly, `horrorshow`, despite its name, seems cool too

svenstaro:
I think `handlebars` does everything we need. I don't think we need logic inside of our templates.

boastful-squirrel:
Right, I'll take a look at it.

boastful-squirrel:
I'm starting this today.

boastful-squirrel:
I'm having quite good results with Maud at the moment, it's blazingly fast ! I'll keep in touch

PR #39, opened 2019-02-18
Improved mobile view

boastful-squirrel:
So I just had a use case where I had to use my phone to browse files, and it was quite painful to have to zoom every time. I tweaked the CSS a little bit to offer a more condensed view when on small screens.

Here is a screenshot:

screenshot 2019-02-18 at 22 15 46

This is getting quite messy while #34 is not done (and that we don't split the CSS in a separate file), but that's an attempt, please tell me what you think :)


boastful-squirrel:
Note that it's a proof of concept based on #38, I'll do a proper PR when we're done with it.

svenstaro:
Should be rebased.

PR #41, opened 2019-02-21
WIP: working example of miniserve using Handlebars

boastful-squirrel:
Still a work in progress.

Right now, I think this version is equivalent to the current release, feature-wise. I'm still unhappy with some parts of the code, especially the fact that I store the datetime both in String format and DateTime format, because I fail to call `convert_to_utc` from within the template. I'll read the documentation of Handlebars, I think they have a mechanism of `handlers`, maybe they're here for that...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,7 +87,7 @@ dependencies = [
  "percent-encoding 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "rand 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.87 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.88 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.38 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_urlencoded 0.5.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "sha1 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


svenstaro:
Well we don't have to use handlebars. There are so many rust templating libraries to choose from and I don't particularly care as long as it's very simple and doesn't require too much code and doesn't bloat the binary too much.

boastful-squirrel:
Oh i'm pretty sure it's simple to do :) It's just that the guy who did the crate decided that the documentation of the Javascript version would be enough, but in this case it's not ahah :D

boastful-squirrel:
Or maybe I could create a "ReprEntry" struct, which would contain the fields as wanted for display, keep the Entry struct untouched, and implement the From trait to convert an Entry into a ReprEntry. It would definitely be cleaner

boastful-squirrel:
So, `Entry` represents an entry "logic-wise", sort operations are done on it, etc. `ReprEntry` is a representation of an `Entry` in the `Handlebar` template, so we can decide on how to display things nicely. I implemented `From<Entry> for ReprEntry` so it's painless to use. 

Just a quick question about errors: I had to use `io::Error` (`io::ErrorKind::Other` to be precise) in the renderer because `directory_listing()` _needs_ to return an `io::Error` (because `files_listing_renderer()` wants an `io::Error`)... I tried to create a custom error type (an enum of `io::Error` and `RendererError`), but then it didn't compile because it didn't match `files_listing_renderer()`'s signature.

I'm quite new to Rust, so errors are a bit obscure to me, if you think it's worth digging deep to implement our own error, I'll do it. If the way I did it is enough (for now ?), then :)

svenstaro:
So first of all, for exactly this purpose there is the [failure](https://github.com/rust-lang-nursery/failure) crate which makes this kind of thing really easy.

Next, I don't really like the complexity introduced by the templating like this, especially with the duplicated struct. Can you take a look at maud? I've been meaning to use that for a while now and it seems pretty cool and ergonomic, though it requires nightly but that's not a dealbreaker for me.

boastful-squirrel:
Thanks for the input on failure, I didn't know it :)

I looked at Maud, it seems very powerful indeed. Since you don't mind using the nightly, then let's go, the resulting code should indeed be less complex.

boastful-squirrel:
I'm closing this PR, see you soon using Maud :)

Issue #37, opened 2019-02-15
Allow clicking the headers to change sort order and sort column


svenstaro:
I suppose this should ignore the command-line provided sort options. Also this should be appended to the URL like so: `localhost:8080?sort=date&order=desc`.

boastful-squirrel:
Good idea ! I think we should solve #34 first, for clarity sake :P

svenstaro:
In fact, now that I think about it, sorting really should be a representational problem and if we went ahead and implemented this then we'd have a sort of duality where you can sort using launch params and by clicking. I think we should only probably allow for letting the end user sort using url parameters and remove the CLI options again (and instead parse the url stuff).

The HTML could just feature some links to click on to change the sort type as well (dirs first vs strictly alphabetical, for instance). Clicking a column header would sort on that and clicking it again would sort the opposite way.

I think that's more intuitive and removes some complexity from the command line options while giving the user more control.

What do you think about that?

boastful-squirrel:
I thought about that too yes, CLI arguments and Javascript-based sorting kind of clash with eachothers. 

The only corner case I see is for people who disable Javascript/use js blocking extensions/use terminal browsers, for those ones, CLI sorting can have a point, but on a more general point of view, I think you're right, it's better to let people sort files the way they want.

Whilst we are at it, a search bar could also be helpful, even if ctrl+f should do the trick since we don't paginate :)

svenstaro:
I'm not talking about JavaScript and I want none in here. The page shall be
regenerated once a column is clicked for sorting.

boastful-squirrel:
Oh, I should learn to read :p In this case, I don't see any corner case so that's probably the best way to handle sorting. Very cool idea!

PR #42, opened 2019-02-23
Renders file listing using maud

boastful-squirrel:
So, this is my WIP using `maud`.

Feature-wise, this is equivalent to `master`. I was thinking, should I load the CSS from an external file in the `css()` method ? Or is the big string fine ? Anyways, I think you were right about using `maud`: it's amazingly fast (I tried to render a very huge directory, it did render in a blink) , and I think the code became less complex than before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,8 @@ addons:
 
 matrix:
   allow_failures:
-    - rust:
-      - stable
-      - beta
+    - rust: stable
+    - rust: beta
   include:
     - env:
       - TARGET=x86_64-unknown-linux-musl
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,18 +20,20 @@ matrix:
     - rust: beta
   include:
     - rust: nightly
-    - env:
+      env:
       - TARGET=x86_64-unknown-linux-musl
       - BIN_NAME=miniserve
       - PROPER_NAME=miniserve-linux
       os: linux
-    - env:
+    - rust: nightly
+      env:
       - TARGET=x86_64-pc-windows-gnu
       - BIN_NAME=miniserve.exe
       - PROPER_NAME=miniserve-win.exe
       - RUSTFLAGS="-C linker=x86_64-w64-mingw32-gcc"
       os: linux
-    - env:
+    - rust: nightly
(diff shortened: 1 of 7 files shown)


svenstaro:
Not a thorough review but keep in mind to also change .travis-ci.yml to only use nightly and set the stable and beta to allow failure.

svenstaro:
The css is fine the way it is. Maybe at some point I'll get annoyed and move it but for now let's keep it.

boastful-squirrel:
I'll do it asap. I don't know Travis so I will take a look at the syntax before doing anything.

Gotcha for the css ;)

boastful-squirrel:
I took the liberty to modify `convert_utc` and `humanize_systemtime`, so that if `None` is given as their parameter, it returns `None` instead of empty strings. `maud` allows the `if let` syntax, so there was no reason to return empty strings anymore. 

Also, I hope I didn't mess with the `travis.yml` file...

svenstaro:
From the looks of it you need to explicitly tell travis to use nightly in the include matrix thing.

boastful-squirrel:
It seems 6aff9bc fixed it

svenstaro:
Well, you still have to do what I said because otherwise the release builds will be done using stable which is travis default as opposed to nightly.

boastful-squirrel:
I hope this builds the matrix as expected. Otherwise, I could write this:
```
- rust: nightly
  env:
      - TARGET=x86_64-unknown-linux-musl
      - BIN_NAME=miniserve
      - PROPER_NAME=miniserve-linux
  os: linux
- rust: nightly
   env:
       - TARGET=x86_64-pc-windows-gnu
       - BIN_NAME=miniserve.exe
       - PROPER_NAME=miniserve-win.exe
       - RUSTFLAGS="-C linker=x86_64-w64-mingw32-gcc"
  os: linux
[...]
```

I hope both notations are equivalent...

svenstaro:
What you did last doesn't work. What you wrote in the comment should work.

boastful-squirrel:
OK cool ! Heading to #37 :)

svenstaro:
Merging. Good job.

PR #44, opened 2019-02-25
Removed sorting from CLI + added sorting from HTML

boastful-squirrel:
So this is my WIP for solving #37 

Right now, I can sort by name/size/date, in ascending and descending order by clicking the corresponding header. I'll clean up some parts of the code which, I think, could be simplified. Sorting by name is done using natural sort, does that seem fine to you ?

### DIFF
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -134,13 +134,11 @@ pub fn directory_listing<S>(
     let is_root = base.parent().is_none() || req.path() == random_route;
     let page_parent = base.parent().map(|p| p.display().to_string());
 
-    let mut sort_method: Option<SortingMethod> = None;
-    let mut sort_order: Option<SortingOrder> = None;
-
-    if let Ok(query) = Query::<QueryParameters>::extract(req) {
-        sort_method = query.sort.clone();
-        sort_order = query.order.clone();
-    }
+    let (sort_method, sort_order) = if let Ok(query) = Query::<QueryParameters>::extract(req) {
+        (query.sort.clone(), query.order.clone())
+    } else {
+        (None, None)
+    };
 
     let mut entries: Vec<Entry> = Vec::new();
 
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -1,36 +1,38 @@
-use actix_web::{fs, HttpRequest, HttpResponse, Result};
+use actix_web::{fs, FromRequest, HttpRequest, HttpResponse, Query, Result};
 use bytesize::ByteSize;
 use htmlescape::encode_minimal as escape_html_entity;
 use percent_encoding::{utf8_percent_encode, DEFAULT_ENCODE_SET};
+use serde::Deserialize;
 use std::io;
 use std::path::Path;
 use std::time::SystemTime;
 
 use crate::renderer;
 
+/// Query parameters
+#[derive(Debug, Deserialize)]
+struct SortingQueryParameters {
+    sort: Option<SortingMethod>,
+    order: Option<SortingOrder>,
+}
+
 /// Available sorting methods
-#[derive(Debug)]
+#[derive(Debug, Deserialize, Clone)]
 pub enum SortingMethod {
     /// Sort by name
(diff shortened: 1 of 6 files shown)


boastful-squirrel:
screenshot 2019-02-26 at 23 27 26

boastful-squirrel:
Note that the SVGs (chevrons) are taken front FontAwesome. FontAwesome icons are licence CC BY 4.0 so I think I have to include the license in the project for these icons. Ir that a problem? If so, I'll draw those icons myself, or maybe use Unicode characters.

svenstaro:
It would be optimal if you could try using unicode chars first. Then it will also look fine in lynx. :)

boastful-squirrel:
So I switched to unicode characters and improved the look and feel a bit, now selected column title is dark gray. If Lynx is an actual use-case for you, I might tweak some things to it looks nicer

svenstaro:
Once you're happy with your code, remove WIP and I'll do a review.

boastful-squirrel:
Sure ;)

boastful-squirrel:
So I made various improvements to the CSS and added a back to top button: I missed it when using my mobile phone... I'll remove the WIP so you can review it. Again, if terminal browsers are an actual use case (Lynx for instance) and want me to rethink the structure of the HTML/CSS so it is displayed perfectly in this kind of browsers, please tell me: with bbc7ef7, it should be better than before, but it's not perfect yet.

boastful-squirrel:
screenshot 2019-03-02 at 21 29 02

svenstaro:
The default order should be to implicitly sort by name, right? It doesn't seem to be the case for me. I honestly can't which order the default is. It seems random.

svenstaro:
Apart from that, this actually seems really good and it's super fast. Great job!

BTW what do you think about highlighting the currently hovered line? Should be a breeze in CSS.

svenstaro:
Also I just took a look and turns out actix-web now has typed URL params: https://docs.rs/actix-web/0.7.18/actix_web/struct.Query.html

So we should probably use those.

boastful-squirrel:
Oh yeah, right now it's not sorted by default. I'll correct that. I'll also use Queries, on my way.

boastful-squirrel:
I think that's it. Please tell me if you see issues with the code. Except the conflicts, which I'll resolve eventually

src/listing.rs
```diff
@@ -104,15 +127,21 @@ pub fn directory_listing<S>(
...
     let title = format!("Index of {}", req.path());
     let base = Path::new(req.path());
     let random_route = format!("/{}", random_route.unwrap_or_default());
     let is_root = base.parent().is_none() || req.path() == random_route;
     let page_parent = base.parent().map(|p| p.display().to_string());
 
+    let mut sort_method: Option<SortingMethod> = None;
+    let mut sort_order: Option<SortingOrder> = None;
+
+    if let Ok(query) = Query::<QueryParameters>::extract(req) {
```
svenstaro: Could you do  this without the `mut` with `let (sort_method, sort_order) = if let ...`?

svenstaro: And, in fact, couldn't you put this into the handler signature like in the *Query* example [here](https://actix.rs/docs/extractors/)?

boastful-squirrel: I'll try, it didn't seem that easy when I first tried. `files_listing_renderer` takes an `HttpRequest` as parameter, so I was quite stuck with that...

boastful-squirrel:
I made the requested change for the query extraction.

About using https://actix.rs/docs/handlers/, I tried, though our use-case seems a bite more complex than what's being presented here. `StaticFiles::files_listing_renderer()` expects an `HttpRequest` (and it seems I can't create an `HttpRequest` from a `Query`, only the opposite, which is what I do). Maybe it's doable, though I can't see how. If you want me to dig deeper into it, I will, though I think I tried everything I could :P - Or we could postpone it to a future PR.

svenstaro:
Merging as is. Great work!

Issue #2, opened 2018-05-17
Streaming zip/tar of directories


svenstaro:
It would be cool and convenient to be able to stream directories on the host as archives on the fly if you want to offer many individual files for download.

boastful-squirrel:
I might take a look at this, it's a cool feature :)

svenstaro:
Yeah, so the way I imagine it to work is to have a link on every page that says "Download as tar | zip" and then it would redirect to `<host>/some_dir/.zip` or `<host>/some_dir/.tar`. Then we just need to hope that there is no `.zip` (and others) file in there which I suppose is a fair assumption to make. We should probably also allow for compression when users choose `<host>/some_dir/.tar.(gz,bz,xz,etc)`.

svenstaro:
We'd have to do some streaming. See [here](https://actix.rs/docs/response) at the very bottom.

Also, I think we'd best use a query param like `?download=targz`, `?download=zip`, etc. I'd like to stay Rust native all in all so we'd have to use a compression library that is pure Rust. There is [libflate](https://crates.io/crates/libflate) and [tar](https://github.com/alexcrichton/tar-rs) as well as [zip-rs](https://github.com/mvdnes/zip-rs).

boastful-squirrel:
Using query parameters is probably better yes, when #44 is merged we'll have the struct to parse this.

boastful-squirrel:
I'm back on this one. Let's see :)

svenstaro:
We now have _downloading_ of archives but not streaming as originally intended. I'm going to leave this open for now.

boastful-squirrel:
We really gotta find a crate that allows streaming. That's the last big chunk

svenstaro:
Yes, definitely. Perhaps even add that to `tar-rs`.

boastful-squirrel:
So, it seems it's doable with tar-rs, according to a comment on Reddit (thanks a lot to him). I could not try it yet, but here is the link of the a tool which is quite similar to miniserve, but has streaming support for folder downloads: https://github.com/lnicola/rusty-share

I must admit I'm not very experienced with async stuff, so it's not guaranteed I understand how he did it :p

cc @vojta7 (you, on the other hand, seems to understand async very well ;))

vojta7:
I can try to look into it, but I wouldn't call my understanding of async good.

gyscos:
[...]

PR #48, opened 2019-03-08
Download folders in .tar.gz format

boastful-squirrel:
This is my WIP for solving #2 (only .tar.gz at the moment).

The code isn't quite clean, the error handling needs some more work, and I have a bug on Mac OS X because user folders are wierd symlinks, and it causes the current code to fail. I'm investigating this issue. 

Otherwise, it works pretty well :)


boastful-squirrel:
So, I had to disable symlinks following strategy in the `TarBuilder` object because of these issues:

- https://github.com/alexcrichton/tar-rs/issues/147
- https://github.com/alexcrichton/tar-rs/issues/174

That's temporary, hopefully !

src/listing.rs
```diff
@@ -218,17 +227,43 @@ pub fn directory_listing<S>(
...
+                .body(Body::Streaming(Box::new(once(Ok(content)))))),
+            Err(_) => {
+                println!(
+                    "{error} an error occured while compressing {folder}",
+                    error = Paint::red("error:").bold(),
+                    folder = dir.path.display(),
+                );
+                Ok(HttpResponse::Ok()
+                    .status(http::StatusCode::INTERNAL_SERVER_ERROR)
+                    .body(""))
```
svenstaro: A helpful message here wouldn't hurt.

boastful-squirrel: Yup, that's in my to-do list :)

boastful-squirrel: I'm wondering, how precise do you want the messages to be ? Most descriptions from `std::io::Error` are not really helpful on the user-end, for instance `Is a directory` doesn't says much. Should I then go deeper than that ?

I know there's the `common_failures` crate that might help us in this case, at least for the `std::io::Error`...

boastful-squirrel:
So, I started adding helpful messages for errors which might occur during the archiving of a folder. It's not done, I still miss some details on some errors, however I can print the full chain of errors which could be helpful for debugging if a user has a problem. I'm not quite sure about the messages I wrote, I'll try to improve them. Oh, and I'll add the missing details for some errors :)

boastful-squirrel:
Solving the tar-rs' issue made me realize that I don't have to actually create the temporary tar file, I can store the bytes in a buffer, just like I do with `gzip`. I'll implement that tonight :)

svenstaro:
I just had a super cursory view and it kind of appears like you're rebuilding pretty-env-logger. I think we should use standard Rust logging.

boastful-squirrel:
Right, I'll be on it ;)

boastful-squirrel:
I switched to `log`, it's clearer indeed. I set the level to `Error` if not in `verbose` mode because I think it's still important for a user to know what happened. 

Also, I changed the _"Info: miniserve has been invoked [...]"_ to an orange notice (kind of a warning, but not on the `log` stream), because I think it can be a security issue in some cases, so better really warn the user. I even thought about putting the whole message in orange, I don't know.

Also, my PR (https://github.com/alexcrichton/tar-rs/pull/187) is going good, I have to add some comments and we should be fine, then we can wait for the new release of `tar-rs`, and eventually we'll be able to set the `follow_symlink` option to `true` (`false` if `-P`) in `archive.rs` 👍

svenstaro:
Yes, it's pretty big already! Let's keep it .tar.gz.

boastful-squirrel:
Great! I'll keep you posted when it comes to the fix of tar-rs.

svenstaro:
It's amazing that you're fixing this issue and the two over at tar-rs in
one go! Keep it up.

src/listing.rs
svenstaro: Well, they don't have to be too precise on the client side but it'd be helpful to have something more than an empty text. They should be pretty accurate on the server side of things, though.

boastful-squirrel:
:)

So my PR has been merged, hopefully a new version will be released

boastful-squirrel:
Here we go ;)

src/renderer.rs
```diff
@@ -259,6 +275,25 @@ fn css() -> Markup {
         color: #3498db;
         text-decoration: none;
     }
+    .download {
+        display: flex;
```
svenstaro: I kind of don't like how this pushes the rest of the content away quite a bit. How about instead position the download button right of the header where it tells you the current directory?

src/listing.rs
svenstaro: I don't think you ever gave this any text. It should have at least a somewhat rough message for what's wrong. Even if you just print the name of the bottom-most exception.

src/errors.rs
```diff
@@ -0,0 +1,90 @@
+use failure::{Backtrace, Context, Fail};
+use std::fmt::{self, Debug, Display};
+
+/// Kinds of errors which might happen during the generation of an archive
+#[derive(Debug, Fail)]
+pub enum CompressionErrorKind {
```
svenstaro: While this seems to have a lot of boilerplate, I'm afraid it's the correct choice for this application.

src/archive.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    pub fn content_encoding(&self) -> ContentEncoding {
+        match &self {
+            CompressionMethod::TarGz => ContentEncoding::Gzip,
+        }
+    }
+}
+
+pub fn create_archive_file(
```
svenstaro: Document this.

src/archive.rs
```diff
@@ -0,0 +1,138 @@
...
+    method: &CompressionMethod,
+    dir: &PathBuf,
+    skip_symlinks: bool,
+) -> Result<(String, Bytes), errors::CompressionError> {
+    match method {
+        CompressionMethod::TarGz => tgz_compress(&dir, skip_symlinks),
+    }
+}
+
+/// Compresses a given folder in .tar.gz format
```
svenstaro: Same, please document inputs/outputs.

src/renderer.rs
boastful-squirrel: You're right. I did some experiments and it's kind of hard to make it look right with very long path though. How about just putting on a new line, but align to the left ? Here's a screenshot, with 1 and then several download buttons

Screenshot 2019-03-15 at 17 22 21

Screenshot 2019-03-15 at 17 23 02

boastful-squirrel:
I edited the documentation so it's a bit clearer, also I implemented the design idea I posted https://github.com/svenstaro/miniserve/pull/48#discussion_r266056153. I'm available for fixes if needed.

As for your last comment, I'll take a look at this, I have no idea wether it's possible, but that sounds like a cool challenge. I didn't really see the problem, since the compression job does not block the user from browsing the server.

svenstaro:
The web server should start sending bytes as soon as we give it bytes but I don't we give it bytes early enough. I'm pretty sure it should work. I'm unsure that what you're doing right now is even supposed to work: Setting the `Content-Length` and `chunked` at the same time. See https://gist.github.com/CMCDragonkai/6bfade6431e9ffb7fe88.

Basically, while we're archiving, we have no idea how long the entire file is going to be but this behavior is already supported by all browsers, thankfully.

boastful-squirrel:
Ok, I get what you mean. So, the first step will probably to make all the methods in `archive.rs` non blocking, they should stream the results of bytes, instead of waiting for the whole input to be consumed. Well, that's gotta be fun ! :)

svenstaro:
I never said it was gonna be easy but in the end I think the payoff is going to be worth it since it will be a lot more useful for sharing large directories! :)

svenstaro:
Keep at it, you've done a marvelous job so far!

boastful-squirrel:
I fear the tar-rs crate won't let us stream its result... All methods seem blocking, I'll dig deeper in the doc

svenstaro:
Ok that really is a bummer. Tell you what, I'll merge this as is without async but I really kinda want async in tbe long run because it does enable sending really large directories which currently is impossible.

boastful-squirrel:
I agree with you. Maybe we should add a little mention in the README, so people don't start trying to share huge folders

boastful-squirrel:
Maybe I could try to reimplement the method that browses a directory, and stream the result for each file...

svenstaro:
That seems like band-aid for what should really be an implementation detail of tar-rs. As I said, we'll need to fix this properly in tar-rs if anything. It shouldn't be _that_ hard.

svenstaro:
Please take a look at the remaining open comments.

src/errors.rs
boastful-squirrel: So should I fix something on that aspect ?

boastful-squirrel:
I think the only comment I didn't fix is https://github.com/svenstaro/miniserve/pull/48#discussion_r266048905, however I don't know if it requires a fix, or if it was just a comment

svenstaro:
Can you just https://github.com/svenstaro/miniserve/pull/48#discussion_r266044603 a bit of text and that's that?

src/errors.rs
svenstaro: Nah, let's leave this part as is.

boastful-squirrel:
No idea what I broke :/

boastful-squirrel:
I added a little warning in the readme about the compression not being async-ready. I hope that's fine with you.

svenstaro:
Ok, back from vacation and merging! Stellar work. Hopefully we can make this async at some point.

Issue #51, opened 2019-03-24
Rework colors


svenstaro:
Currently I'm not super happy with how everything looks. It has a pale color palette without good contrast. It should have better contrast.

boastful-squirrel:
Should be easy, do you have a colorscheme or an application you like design-wise ?

svenstaro:
I actually kinda like that colorscheme. Actualllly how about colorscheme support? We could provide a bunch of colorschemes. The few bytes won't make it too bloaty and another added CLI option won't kill us.

Yeah but generally I also want a more contrasted default color scheme that still looks clean but just has better contrast than the current one.

boastful-squirrel:
Sure, great idea. I'll be on it then ;)

svenstaro:
Yes, absolutely go for hard coding them.

svenstaro:
Maybe actually in the form of having a different CSS template files? I'm not sure which would be better.

boastful-squirrel:
If we only change colors, I think the first option is the easiest. If we want to have actual themes and not only color schemes, the latter option is the best.

I'll start with the first one, we'll see how it goes

svenstaro:
Sounds good.

boastful-squirrel:
I'm doing good progress. I'll create the PR once I created enough themes. 

Here's my version of Zenburn:
<img width="1139" alt="Screenshot 2019-03-26 at 23 02 22" src="https://user-images.githubusercontent.com/47121073/55036594-6a57f880-501b-11e9-8f31-0efba1dcb8af.png">

Everything can be customized if you don't like a color, I already parametrized all the CSS.

svenstaro:
While you're at it. I somehow really don't like the way the date is
formatted under Last modification. It seems weirdly spaced out. Can you
play around with that as well?

boastful-squirrel:
Sure, how would you want it to be ?

svenstaro:
I honestly don't know :P I'd probably play around with it a bit and look at how other software represents the dates.

boastful-squirrel:
OK, I'll look at this.

In the meantime, I came up with that:
<img width="318" alt="Screenshot 2019-03-26 at 23 28 34" src="https://user-images.githubusercontent.com/47121073/55037699-e6a00b00-501e-11e9-91e4-447d9312e0ca.png">

svenstaro:
I like the a lot more than what we have currently! Also try putting the
time just in front of the date and see how it looks.

[...]

PR #55, opened 2019-04-01
Color schemes

boastful-squirrel:
So, here is my WIP to solve #51, and in a more general way, have different colour schemes. At the moment, I implemented 3 dark themes since I'm more comfortable designing them, but I'll eventually come up with a light theme (or multiple ones, if needed), I just have to find a nice way to make the colours fit together...

I also did several improvements to the CSS in general (padding, back to top button bigger, and only visible on small screen (I can set it back if needed), etc). Also, I prepared the moment when we'll have several compression methods available. Download buttons should fit nicely next to each-other.

I am open to any changes when it comes to the colours, or the design in general. Let's use this PR to fix everything that you don't like design-wise :)


src/archive.rs
```diff
@@ -25,6 +25,11 @@ impl CompressionMethod {
         .to_string()
     }
 
+    /// Lists compression methods
+    pub fn get_compression_methods() -> Vec<Self> {
+        vec![CompressionMethod::TarGz]
```
svenstaro: Can't we use strum for this? Then we don't have to maintain this list.

svenstaro:
It would probably be prudent to have a boring, non-offensive default light theme with good contrast. Also, it's probably useful to have a CLI parameter for a default theme.

svenstaro:
Also, while not really related, one thing I noticed is that links to files are also generated to retain the params which is not really useful and actually quite annoying (ie `http://localhost:8080/src/renderer.rs?sort=name&order=asc&theme=archlinux`)
It would be much neater if it were just `http://localhost:8080/src/renderer.rs` since what do we even need the params for on files?

src/themes.rs
```diff
@@ -0,0 +1,193 @@
...
+            ColorScheme::ArchLinux => true,
+            ColorScheme::Zenburn => true,
+            ColorScheme::Monokai => true,
+        }
+    }
+
+    /// Returns the name of a color scheme
+    pub fn get_name(&self) -> String {
+        match &self {
+            ColorScheme::ArchLinux => "Archlinux",
```
svenstaro: There is no `Archlinux`. There is either `Arch Linux` or `archlinux`.

svenstaro:
Looking pretty good! The pink in Arch Linux, while really not at all related to Arch Linux does look pretty good.

src/archive.rs
boastful-squirrel: I was looking for a method for listing the possible variants of an enum but couldn't find one... I'll check

boastful-squirrel:
Got it for all your comments, I'll update the code ASAP :)

boastful-squirrel:
I'm just having a little issue with 

> it's probably useful to have a CLI parameter for a default theme.

I use `serde` on the `ColorScheme` enum, so I can't use `arg_enum!` on it, as it does not read field attributes.

This:
```
arg_enum! {
    #[derive(Debug, Deserialize, PartialEq, Clone)]
    pub enum Foo {
        #[serde(alias = "bar")]
        Bar
    }
}```

will cause this:
```
#[serde(alias = "bar")]
 ^ no rules expected this token in macro call
```

I'm trying to find a proper way of handling this limitation. Also, I'm starting to design a light color scheme.

**Comment by svenstaro** [comment]
Yeah, I've done that exact thing before. See here: https://github.com/svenstaro/overpower/blob/master/src/main.rs

**Comment by boastful-squirrel** [comment]
> Yeah, I've done that exact thing before. See here: https://github.com/svenstaro/overpower/blob/master/src/main.rs

In your example, you don't have `#[stuff_like_this]` on the attributes of your `HttpsMethods` enum. In our case, we have something like this:

```
#[derive(Debug, Deserialize, Clone)]
pub enum ColorScheme {
    #[serde(alias = "archlinux")]
    Archlinux,
    // ...
}
```

The `serde` part is not being parsed properly. Actually, reading at the source code of `arg_enum`, it doesn't expect such attributes in a struct.

**Comment by svenstaro** [comment]
Good point, maybe talk to the serde guys directly and figure that out?

**Comment by svenstaro** [comment]
It should be noted that due to the way the macro is implemented, you don't have to explicitly define a lower case version as the comparison ignores cases: https://kbknapp.github.io/clap-rs/clap/macro.arg_enum!.html

**Comment by boastful-squirrel** [comment]
I think I needed it for query parameters, but if you don't mind having something like ?theme=Archlinux (so, capital A), then we should be fine

**Comment by svenstaro** [comment]
Uh, well the query parameters are kind of neat lowercased.

**Comment by boastful-squirrel** [comment]
OK, then I'll talk to the serde guys ;)

**Comment by boastful-squirrel** [comment]
I found a way without talking to the `serde` guys, I'm quite happy with the result, as I think it makes the code even clearer.

So, this works:
```
arg_enum! {
    #[derive(Debug, Deserialize, Clone)]
    #[serde(rename_all = "lowercase")]
    pub enum ColorScheme {
        Archlinux,
        Zenburn,
        Monokai,
        Squirrel,
    }
}
```

I came up with the light theme myself, so if it lacks contrast or if you would like some color changes, please don't hesitate

**Comment by svenstaro** [comment]
Ok, that way is even neater!

**Comment by boastful-squirrel** [comment]
I removed the WIP label as I'm quite happy with the code. As usual, I'm ready for the review, feel free to also comment on the color schemes, if there are colors choice you don't like, I'm open to change them.

**Comment by boastful-squirrel** [comment]
I'll start working on the upload part, I'll keep you updated

**Comment by boastful-squirrel** [comment]
The Javascript part was really annoying, drag and drop really behaves differently depending on the browser, here's a solution I saw on StackOverflow and which seems to work.

**Comment by boastful-squirrel** [comment]
I think there was too much vertical space wasted with the download button + the upload form.

**Comment by boastful-squirrel** [comment]
I'm currently trying to make mobile info stand out a bit more

**Comment by boastful-squirrel** [comment]
So, to improve the mobile view I took the liberty to remove the last modification from the mobile view (it still exists in normal view, obviously). The problem I had is that the date is so "dense" that it was actually quite hard to spot the filename (which is the most important thing), with all those "Last modification: ...." on every line. 

I also stylised the size of a file so it looks like a "badge" with a highly contrasted background color.

**Comment by svenstaro** [comment]
I'm a super big fan of the squirrel theme! Looks friggen amazing.

File: src/archive.rs (hunk 1 above)
Comment (svenstaro) [reply src/archive.rs]: I still think you should use https://github.com/Peternator7/strum#EnumIter for this.

Comment (svenstaro) [reply src/archive.rs]: Or just remove it as we don't need it right now.

File: src/themes.rs [hunk 3]
```diff
@@ -0,0 +1,275 @@
...
+            ColorScheme::Archlinux => "Archlinux",
+            ColorScheme::Zenburn => "Zenburn",
+            ColorScheme::Monokai => "Monokai",
+            ColorScheme::Squirrel => "Squirrel",
+        }
+        .to_string()
+    }
+
+    /// Lists available color schemes
+    pub fn get_color_schemes() -> Vec<Self> {
```
svenstaro: Same as above, maybe we can generate this?

src/themes.rs
```diff
@@ -0,0 +1,275 @@
...
+        match &self {
+            ColorScheme::Archlinux => true,
+            ColorScheme::Zenburn => true,
+            ColorScheme::Monokai => true,
+            ColorScheme::Squirrel => false,
+        }
+    }
+
+    /// Returns the name of a color scheme
+    pub fn get_name(&self) -> String {
```
svenstaro: And here, maybe use https://github.com/Peternator7/strum#intostaticstr?

src/themes.rs
```diff
@@ -0,0 +1,275 @@
...
+        Archlinux,
+        Zenburn,
+        Monokai,
+        Squirrel,
+    }
+}
+
+impl ColorScheme {
+    /// Returns the URL-compatible name of a color scheme
+    pub fn to_string(&self) -> String {
```
svenstaro: Isn't it more modern to implement `Display`?

src/renderer.rs
```diff
@@ -14,51 +15,122 @@ pub fn page(
...
+    let link = parametrized_link("", &sort_method, &sort_order, &color_scheme);
+    let title = format!("Switch to {} theme", color_scheme.get_name());
+
+    html! {
+        a href=(link) title=(title) {
+            (color_scheme.get_name())
+            " "
+            @if color_scheme.is_dark() {
+                "(dark)"
+            } @ else {
```
svenstaro: Can you fix the spacing here and in other places to make it consistent? I think it should be `} @else {` as that seems to be what you've done in most places.

src/renderer.rs
```diff
@@ -104,48 +199,43 @@ fn build_link(
...
                         }
-                    }
-                }
-                @if !entry.is_dir() {
-                    @if let Some(size) = entry.size {
-                        span .mobile-info {
-                            strong { "Size: " }
-                            (size)
-                            (br())
+                    } @ else if entry.is_symlink() {
```
svenstaro: Spacing as above.

svenstaro:
Overall stellar quality as always and I can't wait to merge this but especially that uploading bug has to be fixed first. Also, I believe we can use strum a bit to cut down on the boilerplate.

src/themes.rs
boastful-squirrel: I'll check this crate :)

boastful-squirrel:
So, I couldn't use `strum` to its full power for the `ColorScheme` enum because `arg_enum` already defines `Display`. I reduced the boilerplate to its maximum, but I still need to keep that `to_slug()` method which is kind of ugly.

src/archive.rs
boastful-squirrel: Done :)

src/themes.rs
boastful-squirrel: Done :)

svenstaro:
Super nice!

PR #52, opened 2019-03-24
File upload

vojta7:
This is my WIP for file uploading #6 

I would like to clean up my error handling and discus some ideas.

- if we can upload files, maybe I should add option to delete files too
- Is it ok to override existing files?
- I think the ability to upload file should be behind CLI flag and not by default

I've tested my code only under linux, but it seems ok.

PR #58, opened 2019-04-04
File uploading

vojta7:
This is continuation of work on pull #52 

I am fairly happy with current state and in my point of view it could be merged.


boastful-squirrel:
I tested it a bit and had no problem with it feature-wise. The design is a bit rough, but I can help on this part. I think this should be merged before #55 so I can take those changes into account in the new design, but that's your wall @svenstaro

src/renderer.rs
```diff
@@ -14,12 +14,22 @@ pub fn page(
...
     html! {
         (page_header(page_title))
-        body {
+        body id="dropContainer" {
             span #top { }
-            h1.title { (page_title) }
+            h1 { (page_title) }
+            @if file_upload {
+                form id="file_submit" action={(upload_route) "?path=" (current_dir)} method="POST" enctype="multipart/form-data" {
+                    p { "Select file to upload" }
```
svenstaro: `p { "Select file to upload or drag it into the window" }`

vojta7: This definitely sounds cleaner and more useful :)

vojta7:
> I tested it a bit and had no problem with it feature-wise, even with random routes (which is a killer feature of `miniserve` to me), great job. The design is a bit rough, but I can help on this part. I think this should be merged before #55 (once the code has been reviewed) so I can take those changes into account in the new design, but that's your call @svenstaro

I am glad to hear, that it works well for you. I will appreciate any ideas on how to improve my design. I was thinking about making whole file upload asynchronous with js. It would definitely improve user experience, but you wouldn't be able to use it without js, which you can now. But maybe this is solvable, I will try to look more into this.

boastful-squirrel:
I think we need to keep a no-javascript compatible version. As for the design goes, there are several improvements I would make: I would probably align the upload firm to the right, I would use a label + hidden input instead of a select as it can be stylized more easily, and finally I would tweak the overlay a little bit, maybe use a big bold dashed border + quite transparent background instead of a box shadow. I don't know, I would just try out stuff.

If your PR is merged before mine, I'll adapt your work to the new design, and in the opposite case, I'll help you integrate this in the new design.

svenstaro:
I'm also cringing at having some JavaScript in here now but I'm afraid that
we can't do dragndrop without it. I'd like to keep the JavaScript to an
absolute minimum though.

I'm not entirely happy with how the overlay looks right now either but I
think we can indeed handle this as part of a later PR.

Once I get home I'll merge your PR and then we'll go from there. At any
rate, stellar work.

vojta7:
I agree, that my design is not good, but my design sense is not the best. So if @boastful-squirrel is willing to improve it as part of his PR, I would be glad.

As far as the JavaScript goes, I did some research about drag and drop and as you said it can't be done without it. But it took just a few lines and if JavaScript is disabled you can still upload, so I think it's worth it.

src/file_upload.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/// Handle incoming request to upload file.
+/// Target file path is expected as path parameter in URI and is interpreted as relative from
+/// server root directory. Any path which will go outside of this directory is considered
+/// invalid.
+/// This method returns future.
+pub fn upload_file(req: &HttpRequest<crate::MiniserveConfig>) -> FutureResponse<HttpResponse> {
+    let app_root_dir = req.state().path.clone().canonicalize().unwrap();
+    let path = match Query::<QueryParameters>::extract(req) {
```
svenstaro: Can't you extract this type safely?

vojta7: I don't think I can, but I could return some type of error if canonicalize() returns an error. My justification for this unwrap was, that server root path should be always valid and exactly same unwrap() happens in [main](https://github.com/svenstaro/miniserve/blob/master/src/main.rs#L116)

svenstaro:
Aaanyway, merging as is! Thanks! :D

PR #60, opened 2019-04-08
Bug fixes

vojta7:
Hi, these are fixes of the following two errors:

- When uploading took longer browser didn't redirect you back after successful upload, but instead displayed error - 'connection aborted'
- When invoked without an explicit path server started immediately even tho there was a countdown.

### DIFF
--- a/src/file_upload.rs
+++ b/src/file_upload.rs
@@ -1,8 +1,9 @@
 use crate::errors::FileUploadErrorKind;
 use crate::renderer::file_upload_error;
 use actix_web::{
-    dev, http::header, multipart, FromRequest, FutureResponse, HttpMessage, HttpRequest,
-    HttpResponse, Query,
+    dev,
+    http::{header, StatusCode},
+    multipart, FromRequest, FutureResponse, HttpMessage, HttpRequest, HttpResponse, Query,
 };
 use futures::{future, Future, Stream};
 use serde::Deserialize;
@@ -138,6 +139,7 @@ pub fn upload_file(req: &HttpRequest<crate::MiniserveConfig>) -> FutureResponse<
             .collect()
             .map(move |_| {
                 HttpResponse::TemporaryRedirect()
+                    .status(StatusCode::SEE_OTHER)
                     .header(header::LOCATION, return_path.to_string())
                     .finish()
             })
--- a/src/file_upload.rs
+++ b/src/file_upload.rs
@@ -144,10 +144,7 @@ pub fn upload_file(req: &HttpRequest<crate::MiniserveConfig>) -> FutureResponse<
                         .header(header::LOCATION, return_path.to_string())
                         .finish(),
                 ),
-                Err(e) => {
-                    let error_description = format!("{}", e);
-                    create_error_response(&error_description, &return_path)
-                }
+                Err(e) => create_error_response(&e.to_string(), &return_path),
             }),
     )
 }
--- a/src/file_upload.rs
+++ b/src/file_upload.rs
@@ -105,9 +105,10 @@ pub fn upload_file(req: &HttpRequest<crate::MiniserveConfig>) -> FutureResponse<
         .to_str()
         .unwrap_or("/")
         .to_owned();
-    let app_root_dir = match req.state().path.canonicalize() {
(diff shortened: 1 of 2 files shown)


svenstaro:
Good catch! Will review this in a bit.

svenstaro:
@boastful-squirrel Could you take a look? I'm a bit pressed for time and can't test out before/after right now.

boastful-squirrel:
I'll check it asap

vojta7:
I removed unwrap in file upload, cleaned up unnecessary clones and a bit refactored upload_file function.

boastful-squirrel:
I'll check it tonight ;)

boastful-squirrel:
Upload went fine on a 1GB file :)

src/file_upload.rs
```diff
@@ -136,17 +137,28 @@ pub fn upload_file(req: &HttpRequest<crate::MiniserveConfig>) -> FutureResponse<
...
+                ),
+                Err(e) => {
+                    let error_description = format!("{}", e);
+                    create_error_response(&error_description, &return_path)
+                }
             }),
     )
 }
+
+// Convenience method for creating response errors, when file upload fails.
```
svenstaro: `errors, when` to `errors if`

src/file_upload.rs
```diff
@@ -136,17 +137,28 @@ pub fn upload_file(req: &HttpRequest<crate::MiniserveConfig>) -> FutureResponse<
...
+fn create_error_response(
+    description: &str,
+    return_path: &str,
+) -> FutureResult<HttpResponse, actix_web::error::Error> {
+    future::ok(
+        HttpResponse::NotAcceptable()
```
svenstaro: This seems to be the wrong error code to use according to this: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406

src/file_upload.rs
```diff
@@ -111,23 +118,17 @@ pub fn upload_file(req: &HttpRequest<crate::MiniserveConfig>) -> FutureResponse<
             }
         }
         Err(_) => {
-            return Box::new(future::ok(
-                HttpResponse::BadRequest().body("Unspecified parameter path"),
```
svenstaro: I think we should keep this as a `BadRequest` as it seems to be the most semantically correct error here.

src/file_upload.rs
```diff
@@ -101,7 +101,14 @@ fn handle_multipart(
...
 /// This method returns future.
 pub fn upload_file(req: &HttpRequest<crate::MiniserveConfig>) -> FutureResponse<HttpResponse> {
-    let app_root_dir = req.state().path.clone().canonicalize().unwrap();
+    let return_path: String = req.headers()[header::REFERER]
+        .to_str()
+        .unwrap_or("/")
+        .to_owned();
+    let app_root_dir = match req.state().path.canonicalize() {
+        Ok(path) => path,
+        Err(_) => return Box::new(create_error_response("Internal server error", &return_path)),
```
svenstaro: Can't you get rid of the `match` here as you don't care about the `Err` anyway? You could use `let app_root_dir = if let ...`.

src/file_upload.rs
vojta7: Yes, it seems to be the best.

vojta7:
> Good fixes with some minor changes required. Next time it would be cool if you made this into separate PRs because then I could already have merged the second fix that you did.
I will keep that in mind.

src/file_upload.rs
```diff
@@ -136,17 +138,28 @@ pub fn upload_file(req: &HttpRequest<crate::MiniserveConfig>) -> FutureResponse<
...
-                        .body(file_upload_error(&error_description, &r_p2.clone()).into_string()),
-                )
+            .then(move |e| match e {
+                Ok(_) => future::ok(
+                    HttpResponse::SeeOther()
+                        .header(header::LOCATION, return_path.to_string())
+                        .finish(),
+                ),
+                Err(e) => {
+                    let error_description = format!("{}", e);
```
svenstaro: Can't you do `e.to_string()` instead?

svenstaro:
:+1:

PR #64, opened 2019-04-13
Fix auth check

boastful-squirrel:
That's the first of the many PRs I'm going to create. I did lots of changes on errors to make the code cleaner and remove every `unwrap/expect` in the `main()` method so the user gets nice error messages instead of panics. 

However, that would be so many changes in 1 PR that I thought I'll split it in several little PRs, so you can check and correct me. This PR and the next one is more about fixing bugs than actually rethinking the errors.

This PR is about auth, the way it was parsed/checked was not compliant with the RFC. This should do it. Note that this allows empty passwords because the RFC says nothing about it, we could totally forbid it if you want.

For now, I use error messages, but in a later PR, this will return a `ContextualError` with the right error kind.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -81,12 +81,20 @@ fn parse_auth(src: &str) -> Result<(String, String), String> {
 
     let username = match split.next() {
         Some(username) => username,
-        None => return Err("Invalid credentials string, expected format is username:password".to_owned())
+        None => {
+            return Err(
+                "Invalid credentials string, expected format is username:password".to_owned(),
+            )
+        }
     };
 
     let password = match split.next() {
         Some(password) => password,
-        None => return Err("Invalid credentials string, expected format is username:password".to_owned())
+        None => {
+            return Err(
+                "Invalid credentials string, expected format is username:password".to_owned(),
+            )
+        }
     };
     // Should we allow empty passwords ?
 
--- a/src/args.rs
+++ b/src/args.rs
@@ -77,13 +77,24 @@ fn parse_interface(src: &str) -> Result<IpAddr, std::net::AddrParseError> {
 
 /// Checks wether the auth string is valid, i.e. it follows the syntax username:password
 fn parse_auth(src: &str) -> Result<(String, String), String> {
-    match src.find(':') {
-        Some(_) => {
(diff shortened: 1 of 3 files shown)


src/args.rs
```diff
@@ -77,13 +77,32 @@ fn parse_interface(src: &str) -> Result<IpAddr, std::net::AddrParseError> {
...
+
+    let password = match split.next() {
+        Some(password) => password,
+        None => {
+            return Err(
+                "Invalid credentials string, expected format is username:password".to_owned(),
+            )
+        }
+    };
+    // Should we allow empty passwords ?
```
svenstaro: Yes, empty passwords are fine if they are alright with the spec. Rephrase the comment to state that fact.

src/args.rs
```diff
@@ -77,13 +77,32 @@ fn parse_interface(src: &str) -> Result<IpAddr, std::net::AddrParseError> {
...
+        Some(username) => username,
+        None => {
+            return Err(
+                "Invalid credentials string, expected format is username:password".to_owned(),
+            )
         }
-        None => Err("Correct format is username:password".to_owned()),
+    };
+
+    let password = match split.next() {
```
svenstaro: Since we can't unroll an iterator directly into a tuple I suppose we're sadly stuck doing it this duplicated way.

src/args.rs
boastful-squirrel: I'll do it asap

boastful-squirrel:
I'll have to add a limit to 255 characters to the password too, not a RFC stuff, just Windows truncating the password if it's longer than 255 chars.

boastful-squirrel:
It's a chance that you made me revise my code, as I was not RFC compliant, I misread some lines.

So, for future reference if anyone ever references this PR in a later stage, it should be noted that some well-know tools such as `httpd` do limit the length of the username to 255 characters too, as stated here: http://httpd.apache.org/docs/2.4/programs/htpasswd.html#restrictions - However, they fail to give any explanation on why it is that way. At least, it's not in respect of the spec, which does not limit the username length.

svenstaro:
Good stuff as always.

PR #65, opened 2019-04-13
Use parse() instead of to_socket_addrs()

boastful-squirrel:
We used to use `to_socket_addrs()` for parsing IP addresses, however this didn't make much sense as we were only accepting valid IP addresses, not domain names. The only exception was `localhost` to fix Windows's behaviour. 

So, I replaced `localhost` with `127.0.0.1` so we can use `parse()` instead of `to_socket_addrs()` which makes no sense anymore, and cost us 2 `unwrap()` instead of 1 (and soon, no `unwrap` at all)

Also, I extracted the interface parsing from the `bind()` call... You guessed it, it's an initial step (but necessary) towards proper error handling in the `main()` function. 

I will wait for your input on this PR and #64, the next one will be a huge chunk, so better take a good start.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -92,7 +92,7 @@ fn main() {
             if interface == IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)) {
                 // If the interface is 0.0.0.0, we'll change it to 127.0.0.1 so that clicking the link will
                 // also work on Windows. Why can't Windows interpret 0.0.0.0?
-                String::from("127.0.0.1")
+                "127.0.0.1".to_string()
             } else if interface.is_ipv6() {
                 // If the interface is IPv6 then we'll print it with brackets so that it is clickable.
                 format!("[{}]", interface)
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -85,18 +85,22 @@ fn main() {
...
-        }
-    });
+    let interfaces = miniserve_config
+        .interfaces
+        .iter()
+        .map(|&interface| {
+            if interface == IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)) {
+                // If the interface is 0.0.0.0, we'll change it to 127.0.0.1 so that clicking the link will
+                // also work on Windows. Why can't Windows interpret 0.0.0.0?
+                String::from("127.0.0.1")
```
svenstaro: I thought nowadays you're supposed to do `.to_string()` so let's do that here.

svenstaro:
Good change! Little nitpick but then merge.

src/main.rs
boastful-squirrel: I'll fix this asap

PR #69, opened 2019-04-16
Modify back to top button

wyhaya:
The style of the back button is irregular on some phones.

preview


svenstaro:
@boastful-squirrel any thoughts on this?

boastful-squirrel:
@svenstaro much better solution that the one I came up with :)

svenstaro:
Thanks!

Issue #70, opened 2019-04-16
It's impossible to switch back to the light theme


svenstaro:
Once you left the default theme, you can't get back into it using the link that's generated in the drop down as the link is wrong.

boastful-squirrel:
Oh dang. It should only happen if no other query parameter (sort order or sort method). I'll fix this ASAP. I forgot this case.

PR #71, opened 2019-04-16
Fix issue 70

boastful-squirrel:
This should fix issue #70, sorry for the oversight. Note that I could not test it extensively


svenstaro:
With random route, this shows `Index of /d35165`. I wonder whether that's considered to be correct? It should be `Index of /` right? It should be relative to the base I think.

Might help to make a test for that at this point! I set some testing infra the other day.

svenstaro:
Or actually it even showed that before so I suppose that's not a new issue.

svenstaro:
Well, the issue I reported is fixed so that's good then.

PR #74, opened 2019-04-18
Create .editorconfig

KSXGitHub:


svenstaro:
Personally not using editor config but if it helps people... Thanks

Issue #75, opened 2019-04-18
Proposal: Password Hashing


KSXGitHub:
## Problem

miniserve already has an `--auth` feature, the problem is, the password is publicly visible. This makes it impossible to expose the command line to the public.

## Solution

Support `--auth username:sha256:password_hash` syntax in addition to the old syntax.

---

P.S. I am trying to learn Rust by contributing to this repository, so don't implement this feature just yet.

svenstaro:
Alternatively, the normal auth option could be extended to accept a hash
prefix such as sha256 and then we don't have to add a parameter per
password hash type.

svenstaro:
To clarify, it would look like this: --auth hi:sha256:hash

Think about it.

KSXGitHub:
> --auth hi:sha256:hash

It looks cool.

KSXGitHub:
@svenstaro I find using `:` as separator to be pretty limited. What if we want to support other things (e.g. salt) in the future?

How about this?

```
--auth username%sha256#hash&salt
```

The idea is: Instead of using `:` as separator, I want to use `:`, `%`, `#`, `&` as leading indicators.

svenstaro:
I don't want to overcomplicate this too much. After all, this tool's
primary selling point is ease of use. I don't think we need a salt.

svenstaro:
That said' I'd be ok with using another separator.

KSXGitHub:
Well, I already edit `parse_base_auth` and I don't want to change it again because I'm lazy.

PR #76, opened 2019-04-19
Add support for hashed password (sha256 and sha512)

KSXGitHub:
Supported syntax:
  * `--auth user:password`
  * `--auth user:sha256:hash`
  * `--auth user:sha512:hash`

Fix https://github.com/svenstaro/miniserve/issues/75

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -37,7 +37,7 @@ struct CLIArgs {
     )]
     interfaces: Vec<IpAddr>,
 
-    /// Set authentication (username:password)
+    /// Set authentication (username:password, username:sha256:hash, or username:sha512:hash)
     #[structopt(short = "a", long = "auth", parse(try_from_str = "parse_auth"))]
     auth: Option<auth::RequiredAuth>,
 
--- a/src/args.rs
+++ b/src/args.rs
@@ -160,6 +160,7 @@ pub fn parse_args() -> crate::MiniserveConfig {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use rstest::rstest_parametrize;
 
     fn create_required_auth(username: &str, password: &str, encrypt: &str) -> auth::RequiredAuth {
         use auth::*;
@@ -176,67 +177,35 @@ mod tests {
         }
     }
 
-    #[test]
-    fn parse_auth_plain() -> Result<(), String> {
+    #[rstest_parametrize(
+        auth_string, username, password, encrypt,
+        case("username:password", "username", "password", "plain"),
+        case("username:sha256:abcd", "username", "abcd", "sha256"),
+        case("username:sha512:abcd", "username", "abcd", "sha512")
+    )]
+    fn parse_auth_valid(auth_string: &str, username: &str, password: &str, encrypt: &str) {
         assert_eq!(
-            parse_auth("username:password")?,
-            create_required_auth("username", "password", "plain")
-        );
-
-        Ok(())
-    }
-
-    #[test]
-    fn parse_auth_sha256() -> Result<(), String> {
-        assert_eq!(
-            parse_auth("username:sha256:abcd")?,
-            create_required_auth("username", "abcd", "sha256")
-        );
-
-        Ok(())
-    }
-
-    #[test]
-    fn parse_auth_sha512() -> Result<(), String> {
-        assert_eq!(
-            parse_auth("username:sha512:abcd")?,
-            create_required_auth("username", "abcd", "sha512")
-        );
-
-        Ok(())
-    }
-
-    #[test]
-    fn parse_auth_invalid_syntax() {
-        assert_eq!(
-            parse_auth("foo").unwrap_err(),
-            "Invalid credentials string, expected format is username:password".to_owned()
+            parse_auth(auth_string).unwrap(),
+            create_required_auth(username, password, encrypt),
         );
     }
 
-    #[test]
-    fn parse_auth_invalid_hash_method() {
-        assert_eq!(
-            parse_auth("username:blahblah:abcd").unwrap_err(),
-            "Invalid hash method, only accept either sha256 or sha512".to_owned()
-        );
-    }
-
-    #[test]
-    fn parse_auth_invalid_hash_string() {
-        assert_eq!(
-            parse_auth("username:sha256:invalid").unwrap_err(),
-            "Hash string is not a valid hex code".to_owned()
-        );
-    }
-
-    #[test]
-    fn parse_auth_excessive_length() {
-        let auth_string = format!("username:{}", "x".repeat(256));
-
-        assert_eq!(
-            parse_auth(&*auth_string).unwrap_err(),
-            "Password length cannot exceed 255 characters".to_owned()
-        );
+    #[rstest_parametrize(
+        auth_string, err_msg,
+        case(
+            "foo",
+            "Invalid credentials string, expected format is username:password"
+        ),
+        case(
+            "username:blahblah:abcd",
+            "Invalid hash method, only accept either sha256 or sha512"
+        ),
+        case(
+            "username:sha256:invalid",
+            "Hash string is not a valid hex code"
+        ),
+    )]
+    fn parse_auth_invalid(auth_string: &str, err_msg: &str) {
+        assert_eq!(parse_auth(auth_string).unwrap_err(), err_msg.to_owned(),);
     }
 }
--- a/src/args.rs
+++ b/src/args.rs
@@ -91,22 +91,21 @@ fn parse_auth(src: &str) -> Result<auth::RequiredAuth, String> {
         None => return Err(errmsg),
     };
 
-    let password = match split.next() {
-        Some(hash) => match second_part {
+    let password = if let Some(hash) = split.next() {
+        match second_part {
             "sha256" => auth::RequiredAuthPassword::Sha256(hash.to_owned()),
             "sha512" => auth::RequiredAuthPassword::Sha512(hash.to_owned()),
             _ => return Err("Invalid hash method, valid methods is sha256".to_owned())
-        },
-        None => {
-            // To make it Windows-compatible, the password needs to be shorter than 255 characters.
-            // After 255 characters, Windows will truncate the value.
-            // As for the username, the spec does not mention a limit in length
-            if second_part.len() > 255 {
-                return Err("Password length cannot exceed 255 characters".to_owned());
-            }
+        }
+    } else {
+        // To make it Windows-compatible, the password needs to be shorter than 255 characters.
+        // After 255 characters, Windows will truncate the value.
+        // As for the username, the spec does not mention a limit in length
+        if second_part.len() > 255 {
+            return Err("Password length cannot exceed 255 characters".to_owned());
+        }
 
-            auth::RequiredAuthPassword::Plain(second_part.to_owned())
-        },
+        auth::RequiredAuthPassword::Plain(second_part.to_owned())
     };
 
     Ok(auth::RequiredAuth {
--- a/src/args.rs
+++ b/src/args.rs
@@ -200,4 +200,31 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn parse_auth_invalid_syntax() {
+        assert_eq!(
+            parse_auth("foo").unwrap_err(),
+            "Invalid credentials string, expected format is username:password".to_owned()
+        );
+    }
+
+    #[test]
+    fn parse_auth_invalid_hash_method() {
+        assert_eq!(
+            parse_auth("username:blahblah:hash").unwrap_err(),
+            "Invalid hash method, only accept either sha256 or sha512".to_owned()
+        );
+    }
+
+    #[test]
+    fn parse_auth_excessive_length() {
+        let password = &"x".repeat(256);
+        let param = "username:".to_owned() + password;
+
+        assert_eq!(
+            parse_auth(&*param).unwrap_err(),
+            "Password length cannot exceed 255 characters".to_owned()
+        );
+    }
 }
--- a/src/args.rs
+++ b/src/args.rs
@@ -161,7 +161,7 @@ pub fn parse_args() -> crate::MiniserveConfig {
 mod tests {
     use super::*;
 
-    fn create_required_auth (username: &str, password: &str, encrypt: &str) -> auth::RequiredAuth {
+    fn create_required_auth(username: &str, password: &str, encrypt: &str) -> auth::RequiredAuth {
         use auth::*;
         use RequiredAuthPassword::*;
 
@@ -171,7 +171,7 @@ mod tests {
                 "plain" => Plain(password.to_owned()),
                 "sha256" => Sha256(hex::decode(password.to_owned()).unwrap()),
                 "sha512" => Sha512(hex::decode(password.to_owned()).unwrap()),
-                _ => panic!("Unknown encryption type")
+                _ => panic!("Unknown encryption type"),
             },
         }
     }
--- a/src/args.rs
+++ b/src/args.rs
@@ -172,26 +172,32 @@ mod tests {
     }
 
     #[test]
-    fn parse_auth_plain() {
+    fn parse_auth_plain() -> Result<(), String> {
         assert_eq!(
-            parse_auth("username:password").unwrap(),
+            parse_auth("username:password")?,
             create_required_auth("username", "password", "plain")
         );
+
+        Ok(())
     }
 
     #[test]
-    fn parse_auth_sha256() {
+    fn parse_auth_sha256() -> Result<(), String> {
         assert_eq!(
-            parse_auth("username:sha256:hash").unwrap(),
+            parse_auth("username:sha256:hash")?,
             create_required_auth("username", "hash", "sha256")
         );
+
+        Ok(())
     }
 
     #[test]
-    fn parse_auth_sha512() {
+    fn parse_auth_sha512() -> Result<(), String> {
         assert_eq!(
-            parse_auth("username:sha512:hash").unwrap(),
+            parse_auth("username:sha512:hash")?,
             create_required_auth("username", "hash", "sha512")
         );
+
+        Ok(())
     }
 }
--- a/src/args.rs
+++ b/src/args.rs
@@ -38,7 +38,8 @@ struct CLIArgs {
     )]
     interfaces: Vec<IpAddr>,
 
-    /// Set authentication (username:password, username:sha256:hash, or username:sha512:hash)
+    /// Set authentication. Currently supported formats:
+    /// username:password, username:sha256:hash, username:sha512:hash
     #[structopt(short = "a", long = "auth", parse(try_from_str = "parse_auth"))]
     auth: Option<auth::RequiredAuth>,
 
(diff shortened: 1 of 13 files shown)


src/args.rs
```diff
@@ -76,36 +76,42 @@ fn parse_interface(src: &str) -> Result<IpAddr, std::net::AddrParseError> {
 }
 
 /// Checks wether the auth string is valid, i.e. it follows the syntax username:password
-fn parse_auth(src: &str) -> Result<(String, String), String> {
+fn parse_auth(src: &str) -> Result<auth::RequiredAuth, String> {
     let mut split = src.splitn(2, ':');
+    let errmsg = "Invalid credentials string, expected format is username:password".to_owned();
```
KSXGitHub: @svenstaro I see you used `to_owned()` to convert `&'static str` into `String`. Why didn't you just use `&'static str` directly? _(I asked this because I am learning Rust)_

KSXGitHub:
BTW, I find the tests is lacking. It only test whether some basic commands fail, it does not concern with whether or not the commands are working correctly.

KSXGitHub:
Don't merge it yet. I have some testing to do.

svenstaro:
> 
> 
> BTW, I find the tests is lacking. It only test whether some basic commands fail, it does not concern with whether or not the commands are working correctly.

I added testing only recently. However, there are now two real integration tests which check whether stuff is working correctly. I'd appreciate it greatly if you went ahead and implemented some more.

KSXGitHub:
Integration test is kind of a headache. But I will add some unit tests for some of my functions.

svenstaro:
What part exactly are you having trouble with? With your specific functionality that you're adding, it _should_ be straight forward I think.

KSXGitHub:
The functionality I added ([line 88 of auth.rs](https://github.com/svenstaro/miniserve/blob/090958451244c54fa0abe5791a12bedc26674c41/src/auth.rs#L88)) gives false positive: I does not prevent user from logging in with wrong password.

KSXGitHub:
@svenstaro I have fixed both false positive and false negative, I think it is ready to be merged.

KSXGitHub:
BTW, since I'm learning Rust, I would like you to explain [this](https://github.com/svenstaro/miniserve/pull/76/commits/57d923329adaae200c5595e3dcb04e207460d59c#r276970769), please.

src/args.rs
svenstaro: It's because I'm returning this `String` and in order to do that it has to be owned because you can't move &'static str.

src/args.rs
```diff
@@ -76,36 +76,43 @@ fn parse_interface(src: &str) -> Result<IpAddr, std::net::AddrParseError> {
...
-    // After 255 characters, Windows will truncate the value.
-    // As for the username, the spec does not mention a limit in length
-    if password.len() > 255 {
-        return Err("Password length cannot exceed 255 characters".to_owned());
-    }
+    let password = match split.next() {
+        Some(hash) => match second_part {
+            "sha256" => auth::RequiredAuthPassword::Sha256(hash.to_owned()),
+            "sha512" => auth::RequiredAuthPassword::Sha512(hash.to_owned()),
+            _ => return Err("Invalid hash method, valid methods is sha256".to_owned())
```
svenstaro: You also accept sha512 but this error doesn't reflect that.

src/args.rs
```diff
@@ -76,36 +76,43 @@ fn parse_interface(src: &str) -> Result<IpAddr, std::net::AddrParseError> {
...
     };
 
-    // To make it Windows-compatible, the password needs to be shorter than 255 characters.
-    // After 255 characters, Windows will truncate the value.
-    // As for the username, the spec does not mention a limit in length
-    if password.len() > 255 {
-        return Err("Password length cannot exceed 255 characters".to_owned());
-    }
+    let password = match split.next() {
+        Some(hash) => match second_part {
```
svenstaro: You can make this into an `if let Some(hash)` because you don't need the value in the `else` part.

src/args.rs
```diff
@@ -150,3 +152,47 @@ pub fn parse_args() -> crate::MiniserveConfig {
...
+                "sha512" => Sha512(password.to_owned()),
+                _ => panic!("Unknown encryption type")
+            },
+        }
+    }
+
+    #[test]
+    fn parse_auth_plain() {
+        assert_eq!(
+            parse_auth("username:password").unwrap(),
```
svenstaro: You can use `?` in `#[test]` functions now.

KSXGitHub: That would make the function returns an `Err` instead of panicking.

src/args.rs
KSXGitHub: You mean like [this](https://github.com/svenstaro/miniserve/pull/78/commits/8a548fd1eba41ecca91928b855eb4241352ba1d8)?

src/args.rs
KSXGitHub: Would this also mean that I have to specify return type?

error

KSXGitHub: This would certainly make the code more verbose: 

new code

Do you still insist on this change?

svenstaro: Well, it depends. If you look at the integration, tests, you can see that we use quite a few `?`s where we would otherwise have multiple `.unwrap`s which is not quite as tidy in exchange for a single complicated `Result` per function. I think it's a fair trade-off and would rather roll with `?` for the time being.

svenstaro:
Could you also write an integration test for this? I know we have no auth tests but you might just be able to change that.

src/args.rs
svenstaro: Yep, that seems fine. Please don't make a separate PR for stuff like that, though. Just change it here in this one. Also, you should consider running clippy on your code. It should tell you stuff like this.

KSXGitHub: OK. I'll merge the change.

src/args.rs
KSXGitHub: As you wish

KSXGitHub:
I'd like to do it in another time because it is going to be complicated.

src/args.rs
KSXGitHub: My bad.

KSXGitHub:
@svenstaro I think I'm done. Please review!

As for the integration tests, I don't want to do it now, I may do it in the future with a separate pull request.

svenstaro:
I'm currently adding some more integration tests before I'm reviewing this.

KSXGitHub:
@svenstaro I have done as you requested.

KSXGitHub:
I want to change `RequiredAuthPassword` in `auth.rs` from carrying hash method to carrying a box of hash function. Doing this would eliminate the needs of branches (`match &required_auth.password`) inside `match_auth` function. But I'm afraid this would reduce readability of the code so I want your opinion on this.

For illustration, I want to change this:

```rust
enum RequiredAuthPassword {
  Plain(String),
  Sha256(Vec<u8>),
  Sha512(Vec<u8>),
```

into this:

```rust
pub struct RequiredAuthPassword<F>
    where F: FnOnce(String) -> Vec<u8>
{
  // this is for plain password or hash
  pub data: Vec<u8>,

  // use `|x| x` for plain password
  // use `get_hash::<Sha256>` for sha256 encrypted password
  // use `get_hash::<Sha512>` for sha512 encrypted password
  pub hash_func: F,
}
```

svenstaro:
Needs rebasing due to a big tidiness merge by @boastful-squirrel. Also, I now added some integration tests that also work with rstest so now it should be really easy for you to add tests for this new feature. Basically, extend the `auth_works` integration test using only `rstest_parametrize` and add the SHA stuff and I'll be happy with that.

svenstaro:
> I want to change `RequiredAuthPassword` in `auth.rs` from carrying hash method to carrying a box of hash function. Doing this would eliminate the needs of branches (`match &required_auth.password`) inside `match_auth` function. But I'm afraid this would reduce readability of the code so I want your opinion on this.
> 
> For illustration, I want to change this:
> 
> ```rust
> enum RequiredAuthPassword {
>   Plain(String),
>   Sha256(Vec<u8>),
>   Sha512(Vec<u8>),
> ```
> 
> into this:
> 
> ```rust
> pub struct RequiredAuthPassword<F>
>     where F: FnOnce(String) -> Vec<u8>
> {
>   // this is for plain password or hash
>   pub data: Vec<u8>,
> 
>   // use `|x| x` for plain password
>   // use `get_hash::<Sha256>` for sha256 encrypted password
>   // use `get_hash::<Sha512>` for sha512 encrypted password
>   pub hash_func: F,
> }
> ```

I think this might be overdoing it a bit as I don't think we need such a generic way of parsing that. You might be able to find a way to decrease the repetition using some other measure. The stuff you suggested increases complexity a bit too much IMO.

src/args.rs
```diff
@@ -37,9 +37,9 @@ struct CLIArgs {
     )]
     interfaces: Vec<IpAddr>,
 
-    /// Set authentication (username:password)
+    /// Set authentication (username:password, username:sha256:hash, or username:sha512:hash)
```
svenstaro: I suggest:
```   
/// Set authentication. Currently supported formats:
/// username:password, username:sha256:hash, or username:sha512:hash)
```

src/args.rs
```diff
@@ -76,36 +76,47 @@ fn parse_interface(src: &str) -> Result<IpAddr, std::net::AddrParseError> {
...
+        let hash_bin = match hex::decode(hash_hex) {
+            Ok(hash_bin) => hash_bin,
+            _ => return Err("Hash string is not a valid hex code".to_owned()),
+        };
 
-    Ok((username.to_owned(), password.to_owned()))
+        match second_part {
+            "sha256" => auth::RequiredAuthPassword::Sha256(hash_bin.to_owned()),
+            "sha512" => auth::RequiredAuthPassword::Sha512(hash_bin.to_owned()),
+            _ => return Err("Invalid hash method, only accept either sha256 or sha512".to_owned()),
```
svenstaro: _` => return Err("Invalid hash method, must be either sha256 or sha512".to_owned()), `

src/args.rs
```diff
@@ -76,36 +76,47 @@ fn parse_interface(src: &str) -> Result<IpAddr, std::net::AddrParseError> {
...
-        None => {
-            return Err(
-                "Invalid credentials string, expected format is username:password".to_owned(),
-            )
-        }
+        None => return Err(errmsg),
     };
 
-    let password = match split.next() {
+    let second_part = match split.next() {
```
svenstaro: Please document the semantics of `second_part`. I know it's either just plain password or hash stuff but this doesn't become obvious from the new name.

src/args.rs
```diff
@@ -76,36 +76,47 @@ fn parse_interface(src: &str) -> Result<IpAddr, std::net::AddrParseError> {
...
-    // To make it Windows-compatible, the password needs to be shorter than 255 characters.
-    // After 255 characters, Windows will truncate the value.
-    // As for the username, the spec does not mention a limit in length
-    if password.len() > 255 {
-        return Err("Password length cannot exceed 255 characters".to_owned());
-    }
+    let password = if let Some(hash_hex) = split.next() {
+        let hash_bin = match hex::decode(hash_hex) {
+            Ok(hash_bin) => hash_bin,
+            _ => return Err("Hash string is not a valid hex code".to_owned()),
```
svenstaro: You can make this into `if let {...} else {...}` instead of the more complex `match` if you don't need the match `Err` value.

src/auth.rs
```diff
@@ -16,6 +17,20 @@ pub struct BasicAuthParams {
     pub password: String,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum RequiredAuthPassword {
```
svenstaro: Requires doc comment as it's public.

src/args.rs
```diff
@@ -150,3 +156,56 @@ pub fn parse_args() -> crate::MiniserveConfig {
         file_upload: args.file_upload,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest_parametrize;
+
+    fn create_required_auth(username: &str, password: &str, encrypt: &str) -> auth::RequiredAuth {
```
svenstaro: You should describe the semantics of this function. Something like `Helper function that ...`.

src/args.rs
```diff
@@ -150,3 +156,56 @@ pub fn parse_args() -> crate::MiniserveConfig {
...
+            password: match encrypt {
+                "plain" => Plain(password.to_owned()),
+                "sha256" => Sha256(hex::decode(password.to_owned()).unwrap()),
+                "sha512" => Sha512(hex::decode(password.to_owned()).unwrap()),
+                _ => panic!("Unknown encryption type"),
+            },
+        }
+    }
+
+    #[rstest_parametrize(
```
svenstaro: I really like this one. It's minimal and gets the message across.

src/args.rs
```diff
@@ -150,3 +156,56 @@ pub fn parse_args() -> crate::MiniserveConfig {
...
+        case(
+            "foo",
+            "Invalid credentials string, expected format is username:password"
+        ),
+        case(
+            "username:blahblah:abcd",
+            "Invalid hash method, only accept either sha256 or sha512"
+        ),
+        case(
+            "username:sha256:invalid",
```
svenstaro: I also want one for sha512 :)

src/auth.rs
```diff
@@ -34,6 +49,34 @@ pub fn parse_basic_auth(
     })
 }
 
+pub fn match_auth(basic_auth: BasicAuthParams, required_auth: &RequiredAuth) -> bool {
```
svenstaro: Please add doc comment.

src/auth.rs
```diff
@@ -34,6 +49,34 @@ pub fn parse_basic_auth(
...
+        RequiredAuthPassword::Sha256(password_hash) => {
+            compare_hash::<Sha256>(basic_auth.password, password_hash)
+        }
+        RequiredAuthPassword::Sha512(password_hash) => {
+            compare_hash::<Sha512>(basic_auth.password, password_hash)
+        }
+    }
+}
+
+pub fn compare_hash<T: Digest>(password: String, hash: &Vec<u8>) -> bool {
```
svenstaro: Please add doc comment.

src/auth.rs
```diff
@@ -34,6 +49,34 @@ pub fn parse_basic_auth(
...
+    get_hash::<T>(password) == *hash
+}
+
+pub fn get_hash<T: Digest>(text: String) -> Vec<u8> {
```
svenstaro: Please add doc comment.

src/args.rs
KSXGitHub: So the change you are suggesting here is to replace `only accept` with `must be`?

src/args.rs
KSXGitHub: What do you prefer? A comment or a long but descriptive variable name?

svenstaro: I think a comment is a more suitable measure here.

src/args.rs
svenstaro: Yes.

src/args.rs
KSXGitHub: You should comment on line 186 (the last line of the block) instead of this line (the first line of the block) to make it easier for me to know which block you are talking about.

src/auth.rs
```diff
@@ -70,3 +111,72 @@ impl Middleware<crate::MiniserveConfig> for Auth {
         Ok(Response::Done(resp))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rstest::rstest_parametrize;
+
+    fn get_hash_func(name: &str) -> impl FnOnce(String) -> Vec<u8> {
```
svenstaro: Any chance you can put the type into `test_get_hash` directly without converting it from string? Not required but it would make this a bit neater and less complex.

KSXGitHub:
> We might want to document that we're easily attacked by a timing attack as we don't have constant time password comparison. I'm going to make an issue about it.

In README? Then you should do it yourself. I don't want to screw up the layout.

src/args.rs
svenstaro: True, will keep that in mind.

src/auth.rs
KSXGitHub: Once again, you selected the first line of the block to comment.

src/auth.rs
```diff
@@ -70,3 +111,72 @@ impl Middleware<crate::MiniserveConfig> for Auth {
...
+        case("abc", "sha512", "ddaf35a193617abacc417349ae20413112e6fa4e89a97ea20a9eeee64b55d39a2192992a274fc1a836ba3c23a3feebbd454d4423643ce80e2a9ac94fa54ca49f"),
+    )]
+    fn test_get_hash(password: &str, hash_method: &str, hash: &str) {
+        let hash_func = get_hash_func(hash_method);
+        let expected = hex::decode(hash).expect("Provided hash is not a valid hex code");
+        let received = hash_func(password.to_owned());
+        assert_eq!(received, expected);
+    }
+
+    fn create_required_auth(username: &str, password: &str, encrypt: &str) -> RequiredAuth {
```
svenstaro: Would be nice if this helper function were also documented.

src/auth.rs
KSXGitHub: Okay, now regarding your question, I had tried yesterday, but it seems that the macro cannot take variable names, it can only take literals (strings, numbers, booleans).

svenstaro:
Github is also showing me that you're adding a file `stale_outputs_checked` which is probably a mistake.

svenstaro:
> > We might want to document that we're easily attacked by a timing attack as we don't have constant time password comparison. I'm going to make an issue about it.
> 
> In README? Then you should do it yourself. I don't want to screw up the layout.

I made an issue: #83.

src/auth.rs
KSXGitHub: What happened is, I tried to pass the function directly, then `cargo test` skips `test_hash_func` as an unused function.

svenstaro: Ok, that's too bad then.

svenstaro:
All in all pretty solid and well-tested change! Some nitpicks and some more tests required but overall I'm pretty happy with this.

KSXGitHub:
BTW, `PasswordTooLongError` sounds silly, what about `PasswordLengthError`?

svenstaro:
> BTW, `PasswordTooLongError` sounds silly, what about `PasswordLengthError`?

Maybe it's silly but it's super descriptive. :)

KSXGitHub:
The reason I call it silly is because `PasswordTooLong` is a complete sentence and not a norm.

There's another alternative: `ExcessivelyLongPassword`. `ExcessivePasswordLength`.

src/auth.rs
KSXGitHub: Maybe we can open an issue in rstest repository?

svenstaro:
Sure, that's a good idea. Could you do that? You have all the context.

svenstaro:
Technically it's missing the verb so I think it's fine. :)

src/args.rs
KSXGitHub: Welp, this is irrelevant now. But you may want to check out `src/errors.rs`.

KSXGitHub:
@svenstaro It is done. Check it out!

Anyway, I'm curious: Why do you prefer `if let` over `match`?

svenstaro:
@KSXGitHub because it is conceptually simpler than `match` in that the `else` part can't provide a value. It's just the principle of using the least powerful syntax you can use to express what you need. Basically I strive to minimize complexity wherever I can.

svenstaro:
Merged! Super nice work and thanks for the contribution.

KSXGitHub:
You may want to update your README to mention this feature.

PR #79, opened 2019-04-21
WIP: rework errors and avoid panics in main()

boastful-squirrel:
So, this is an attempt to rework the error system and avoid panics in the `main()` function (1). I did quite a lot of changes. I especially tried to have more general errors (with parameters) instead of creating specific errors for each case. I took example in burntsushi's codebase, he does that a lot.

What also changed is that now the error chain is printed on the webpage and does not use `failure`'s chain system (2). I "inherit" from errors, so you always know what's the underlying error. 

I will try to add some tests, in the scope of this PR (hence the WIP), since some errors can be quite easily emulated, so it's not that expensive to add some tests for them.

(1) There's 1 `expect` I am unable to remove, but it's also a case where `panic` can be a reasonable solution (main.rs:254)
(2) I will try to switch back to `failure`'s chain system, though I faced some corner cases I couldn't solve. However, I think the errors are clearer than they were before, but that probably means I was not using `failure` to its full potential :)

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -1,6 +1,5 @@
 use actix_web::http::ContentEncoding;
 use bytes::Bytes;
-use failure::ResultExt;
 use libflate::gzip::Encoder;
 use serde::Deserialize;
 use std::io;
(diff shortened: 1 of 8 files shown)


boastful-squirrel:
Actually, I'll close this PR as it doesn't make much sense while the error chain system is not re-implemented. I'll re-open it once it's done

PR #86, opened 2019-04-27
Add hashed password to integration test

KSXGitHub:
## Missing

* Tests where it rejects (user logs in with wrong username and/or password)
* Tests where value of `--auth` is not valid

### DIFF
--- a/tests/cli.rs
+++ b/tests/cli.rs
@@ -90,10 +90,13 @@ fn serves_requests_with_non_default_port(tmpdir: TempDir, port: u16) -> Result<(
         "testpassword"
     ),
 )]
-fn auth_works(cli_auth_arg: &str, client_username: &str, client_password: &str) -> Result<(), Error> {
-    let tmpdir = self::tmpdir();
-    let port = self::port();
-
+fn auth_works(
+    tmpdir: TempDir,
+    port: u16,
+    cli_auth_arg: &str,
+    client_username: &str,
+    client_password: &str
+) -> Result<(), Error> {
     let mut child = Command::cargo_bin("miniserve")?
         .arg(tmpdir.path())
         .arg("-p")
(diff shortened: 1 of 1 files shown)


tests/cli.rs
```diff
@@ -75,14 +76,30 @@ fn serves_requests_with_non_default_port(tmpdir: TempDir, port: u16) -> Result<(
...
+        "testpassword"
+    ),
+    case(
+        "testuser:sha512:e9e633097ab9ceb3e48ec3f70ee2beba41d05d5420efee5da85f97d97005727587fda33ef4ff2322088f4c79e8133cc9cd9f3512f4d3a303cbdb5bc585415a00",
+        "testuser",
+        "testpassword"
+    ),
+)]
+fn auth_works(cli_auth_arg: &str, client_username: &str, client_password: &str) -> Result<(), Error> {
+    let tmpdir = self::tmpdir();
```
svenstaro: You should use the fixtures rather than calling them explicitly.

KSXGitHub: Hmm, how am I going to combine it with `rstest_parametrize`?

svenstaro: I just tested it and the naïve way works just fine.

KSXGitHub: Okay? You didn't answer [my question](https://github.com/svenstaro/miniserve/pull/86#discussion_r279147355)

svenstaro: I did, I'm saying there is nothing special to do. Naïve way works fine. Example:

```
#[rstest_parametrize(lol, case(3))]
fn serves_requests_with_no_options(tmpdir: TempDir, lol : usize) -> Result<(), Error> {
    let mut child = Command::cargo_bin("miniserve")?
        .arg(tmpdir.path())
        .stdout(Stdio::null())
        .spawn()?;
```

KSXGitHub: To be fair, your first answer is pretty vague.

Anyway, I have done as you requested.

svenstaro:
I like that, good work!

PR #88, opened 2019-04-27
Combine ContextualError and ContextualErrorKind into one

KSXGitHub:
Close https://github.com/svenstaro/miniserve/issues/87

PR #90, opened 2019-04-27
Themed errors

boastful-squirrel:
Error pages should now follow the query parameters the user had before. So, this respects the theme, but also the potential sorting parameters the user had (this is useful when using the _go back to file listing_ link). Also, I implemented `Copy` for some of our enums, as it greatly reduce the boilerplate and makes the code more readable, I think.


boastful-squirrel:
I'll probably merge some errors too

svenstaro:
Needs rebase as #88 is merged.

svenstaro:
Accessing a 404 doesn't yield a nice user friendly error page.

svenstaro:
Do you think it would be a good idea to have all error pages have a header that just says **Error** and then state the actual error?

svenstaro:
Apart from that, I really quite like this, though I kinda wish you'd done the cleanup of the other error stuff in a separate PR.

svenstaro:
Also the auth failure error doesn't have a nice page yet. Is that scope of this PR?

boastful-squirrel:
I will work on them too asap.

boastful-squirrel:
Working on it

boastful-squirrel:
I am quite happy with the result :) I fixed some Clippy's warning which are being solved in #94, sorry, bad habit

src/main.rs
```diff
@@ -266,26 +266,33 @@ fn configure_app(app: App<MiniserveConfig>) -> App<MiniserveConfig> {
...
+                .default_resource(|r| r.method(Method::GET).f(p404))
         }
     } else {
         // Handle single files
         app.resource(&full_route, |r| r.f(listing::file_handler))
+            .default_resource(|r| r.method(Method::GET).f(p404))
     }
 }
 
 fn p404(req: &HttpRequest<crate::MiniserveConfig>) -> Result<HttpResponse, io::Error> {
```
svenstaro: What does the `p` stand for?

src/args.rs
```diff
@@ -185,7 +183,10 @@ mod tests {
     }
 
     #[rstest_parametrize(
-        auth_string, username, password, encrypt,
+        auth_string,
+        username,
+        password,
+        encrypt,
```
svenstaro: Can we keep these like they were? Makes it clearer.

src/main.rs
boastful-squirrel: I guess it means page, I took it from actix examples directly. Do you want me to rename it?

src/args.rs
boastful-squirrel: We can yes. I'll have to tell rustfmt not to touch this part and it should be fine :)

svenstaro:
Bug: When going back using the generated page link while you have a theme active, the chosen theme is forgotten and you get the default.

src/auth.rs
```diff
@@ -99,28 +97,72 @@ impl Middleware<crate::MiniserveConfig> for Auth {
...
+        None => "/".to_string(),
+    };
+
+    renderer::render_error(
+        &error.to_string(),
+        error_code,
+        &return_path,
+        None,
+        None,
+        req.state().default_color_scheme,
```
svenstaro: I think this line is a bug. It should be the current theme, not default.

src/auth.rs
```diff
@@ -163,13 +205,18 @@ mod tests {
     }
 
     #[rstest_parametrize(
-        should_pass, param_username, param_password, required_username, required_password, encrypt,
+        should_pass,
+        param_username,
+        param_password,
+        required_username,
+        required_password,
+        encrypt,
```
svenstaro: Let's keep these as they were as well.

src/main.rs
svenstaro: Yes, as you probably already figured, I like more explicit names generally.

svenstaro: How about `error_404`?

src/file_upload.rs
```diff
@@ -121,37 +116,99 @@ fn handle_multipart(
...
+            let err = ContextualError::InvalidHTTPRequestError(e.to_string());
             return Box::new(create_error_response(
-                "Unspecified parameter path",
+                &err.to_string(),
+                StatusCode::BAD_REQUEST,
                 &return_path,
-            ))
+                None,
+                None,
+                default_color_scheme,
```
svenstaro: This too should return the currently active color scheme, I think.

src/main.rs
boastful-squirrel: Sure ;)

src/renderer.rs
```diff
@@ -761,11 +845,42 @@ fn humanize_systemtime(src_time: Option<SystemTime>) -> Option<String> {
...
+        body {
+            (page_header(&error_code.to_string(), color_scheme, false, true))
+            div.error {
+                p { (error_code.to_string()) }
+                @for error in error_description.lines() {
+                    p { (error) }
+                }
+                div.error-nav {
+                    a.error-back href=(link) {
+                        "Go back to file listing"
```
svenstaro: I'm not sure it makes sense to render this even in the case of a 401 Unauthorized error. After all, it means you cannot view the file listing because you didn't provide auth.

svenstaro:
Is there any chance I can trick you into making a follow up PR which adds some integration tests for all of these errors?

src/auth.rs
boastful-squirrel: Is there a use case where the user is not authenticated to the server, but as a custom theme set in the URL ? The only case I see would be to have an instance of miniserve running, set a custom theme, shutdown miniserve, and restart it, but that's kind of a corner case.

Note you'll observe the same problem on the page 404. I'll try to see how easy it is to parse the URL to retrieve query parameters, because at this stage of the code, that's the only way I can do it :)

src/renderer.rs
boastful-squirrel: Yep, that's my thought too. I'll remove it.

src/file_upload.rs
boastful-squirrel: This one I won't be able to fix. It's the error occurring when trying to extract the query parameters from the URI. I can't retrieve the active color scheme if the extracting fails :P

src/auth.rs
boastful-squirrel: 401 now does not have the back link as you suggested, and I fixed it for 404. Pushing asap.

boastful-squirrel:
I will try

src/auth.rs
boastful-squirrel: Done in next push

src/file_upload.rs
svenstaro: Ah, hah yeah fair enough.

src/auth.rs
svenstaro: > Is there a use case where the user is not authenticated to the server, but as a custom theme set in the URL ? The only case I see would be to have an instance of miniserve running, set a custom theme, shutdown miniserve, and restart it, but that's kind of a corner case.
> 
> Note you'll observe the same problem on the page 404. I'll remove the back link on 401 as you suggested, and fix it for 404.

Fair point. I suppose the custom theme doesn't make much sense there.

boastful-squirrel:
All in all, this made the code simpler. 

I just have a question, right now, I chose that if parsing the query parameters fail, I just set all values to `None` instead of displaying an error page. Do you prefer an error page ?

svenstaro:
I think most other sites just ignore garbage parameters but I think throwing away everything if there's one bit of garbage is also reasonable. After all, how would the garbage get in there to begin with?

svenstaro:
Needs rebase.

boastful-squirrel:
l'll do it asap. l'll also fix some Clippy warnings I introduced

boastful-squirrel:
Note that the Clippy warning in `main.rs:86` is caused by `rustfmt`.

boastful-squirrel:
I can fix those Clippy errors if needed, though I think those fixes should be on another PR

svenstaro:
I'd like to get this merged. Can you make a separate PR for the clippy fixes and maybe some of the formatting fixes as well? It will effectively decrease the size of this PR if we can get some of the formatting fixes into another PR.

boastful-squirrel:
Do you have some formatting fixes in mind? Many fixes are done by rustfmt. I could remove those changes, though it's weird pushing badly formatted code :p

svenstaro:
Well there are some changes made by rustfmt which are strictly speaking not
even part of the code you touched in this PR. You could push those changes
in a separate together with the clippy fixes.

boastful-squirrel:
That exceeds my knowledge of git, I'll look at the proper way to do this :)

boastful-squirrel:
It was a really wierd experience to write badly formatted code :P

boastful-squirrel:
Do you want me to modify anything else?

svenstaro:
Well, it would be nice if it built. :P

boastful-squirrel:
Well, the build fail is due to Clippy warnings. Should I correct them in this PR?

svenstaro:
As I said, I would prefer if you made another PR with the fixes and
cosmetic changes so this one wouldn't have to carry those.

boastful-squirrel:
So it builds :p

svenstaro:
I know, but I just would prefer to merge something that won't make master go red.

svenstaro:
Eh, whatever, I'll just merge. This has been open for long enough. Please make sure to turn master green in follow-ups. :P

PR #94, opened 2019-05-01
Use Clippy

KSXGitHub:
Fix https://github.com/svenstaro/miniserve/issues/93

**TODO:**
- [x] Wait for Travis CI to fail
- [x] Fix the code to satisfy clippy

PR #98, opened 2019-05-02
Move 'rust: [stable, beta]' into 'matrix.includes'

KSXGitHub:
## Reasons

* I think it is less confusing that we do not specify global `rust` versions
* This enable me to do [what I want](https://github.com/svenstaro/miniserve/pull/94#issuecomment-488231000) with Clippy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 language: rust
 dist: xenial
 
-rust:
-  - stable
-  - beta
-  - nightly
+rust: []
+env: []
 
 addons:
   apt:
@@ -20,6 +18,9 @@ matrix:
     - rust: stable
     - rust: beta
   include:
+    - rust: stable
+    - rust: beta
+    - rust: nightly
     - rust: nightly
       env:
         - TARGET=x86_64-unknown-linux-musl
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,16 @@ addons:
 matrix:
   allow_failures:
     - rust: stable
+      env: []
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,10 +1,8 @@
 language: rust
 dist: xenial
 
-rust:
-  - stable
-  - beta
-  - nightly
+rust: []
```
svenstaro: Do we even need this anymore since you now specify this in the matrix explicitly? Same for the `env: []` line. I really just don't know travis that well.

svenstaro:
I generally like this idea. Can you explain why we need the explicit empty `env: []`?

.travis.yml
KSXGitHub: It is required for `allow_failures` to work properly.

(I don't know who at Travis think this madness is a good design, but it is what it is)

KSXGitHub:
> I generally like this idea. Can you explain why we need the explicit empty `env: []`?

Do you really need to repeat yourself? https://github.com/svenstaro/miniserve/pull/98#discussion_r280456289

.travis.yml
KSXGitHub: BTW,

Documentation: https://docs.travis-ci.com/user/customizing-the-build#rows-that-are-allowed-to-fail

See the `env: # important!` line?

image

svenstaro: Ok, thanks for the reference.

svenstaro:
Good suggestion, will do.

Issue #93, opened 2019-04-30
Use clippy to improve code quality


KSXGitHub:
[Clippy](https://github.com/rust-lang/rust-clippy) is a linter for Rust. It prevents us from making common mistakes. I've recently tried running `cargo clippy` for `main.rs` and this is what I got:

<details>
<summary>Clippy warnings</summary>

<pre><font color="#89E234"><b>    Checking</b></font> miniserve v0.4.1 (/home/khai/programming/miniserve)
<font color="#FBE84F"><b>warning</b></font><b>: writing `&amp;Vec&lt;_&gt;` instead of `&amp;[_]` involves one more reference and cannot be used with non-Vec-based slices.</b>
  <font color="#729ECF"><b>--&gt; </b></font>src/auth.rs:78:56
   <font color="#729ECF"><b>|</b></font>
<font color="#729ECF"><b>78</b></font> <font color="#729ECF"><b>| </b></font>pub fn compare_hash&lt;T: Digest&gt;(password: String, hash: &amp;Vec&lt;u8&gt;) -&gt; bool {
   <font color="#729ECF"><b>| </b></font>                                                       <font color="#FBE84F"><b>^^^^^^^^</b></font> <font color="#FBE84F"><b>help: change this to: `&amp;[u8]`</b></font>
   <font color="#729ECF"><b>|</b></font>
   <font color="#729ECF"><b>= </b></font><b>note</b>: #[warn(clippy::ptr_arg)] on by default
   <font color="#729ECF"><b>= </b></font><b>help</b>: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg

<font color="#FBE84F"><b>warning</b></font><b>: identical conversion</b>
   <font color="#729ECF"><b>--&gt; </b></font>src/listing.rs:159:18
    <font color="#729ECF"><b>|</b></font>
<font color="#729ECF"><b>159</b></font> <font color="#729ECF"><b>| </b></font>    for entry in dir.path.read_dir()? {
    <font color="#729ECF"><b>| </b></font>                 <font color="#FBE84F"><b>^^^^^^^^^^^^^^^^^^^^</b></font> <font color="#FBE84F"><b>help: consider removing `dir.path.read_dir()?()`: `dir.path.read_dir()?`</b></font>
    <font color="#729ECF"><b>|</b></font>
    <font color="#729ECF"><b>= </b></font><b>note</b>: #[warn(clippy::identity_conversion)] on by default
[...]

PR #99, opened 2019-05-02
A better Clippy integration

KSXGitHub:
Fix https://github.com/svenstaro/miniserve/issues/93
Close https://github.com/svenstaro/miniserve/pull/94

## Features

* ~~Clippy on Stable Rust that does not allow failures.~~
* ~~Clippy on Nightly Rust that allow failures.~~
* Clippy on pinned nightly.


rls.toml
```diff
@@ -0,0 +1 @@
+clippy_preference = "on"
```
svenstaro: Is this file actually currently being read by anything? Googling for it, I find some bits of old information but that doesn't seem current.

KSXGitHub: Hmm, no wonder it had no effects on my editor. What's the new file?

svenstaro: I think it depends on the editor. In vim, it even depends on the particular LSP plugin. In VScode, you apparently set it in settings.json.

KSXGitHub: I'll just remove this then.

KSXGitHub:
Here's the thing: "#[feature(...)]` is not allowed in stable channel". This makes it impossible to use Clippy on Stable Rust. It seems that the only option left is to pin a Nightly version.

svenstaro:
Alright, well let's just do that then and see how it goes.

svenstaro:
Cool!, let's do it like this. Good work.

PR #102, opened 2019-05-04
Add integration test where it rejects authentication

KSXGitHub:


tests/auth.rs
```diff
@@ -57,3 +57,60 @@ fn auth_works(
...
+    sleep(Duration::from_secs(1));
+
+    let client = reqwest::Client::new();
+    let status = client
+        .get(format!("http://localhost:{}", port).as_str())
+        .basic_auth(client_username, Some(client_password))
+        .send()?
+        .status();
+
+    assert_eq!(status.canonical_reason(), Some("Unauthorized"));
```
svenstaro: I'd prefer if we compared the whole object like this:

```
assert_eq!(status, StatusCode::UNAUTHORIZED);
```

I think that's more readable and slightly less string-ly typed.

svenstaro:
I like the tests apart from a small nitpick.

PR #103, opened 2019-05-04
Add assertion of status code to auth_works

KSXGitHub:
I would even argue that checking for status code is enough, that parsing HTML and accessing every file is irrelevant and unnecessary, and in order to speed up the test, we should delete the irrelevant part.

### DIFF
--- a/tests/auth.rs
+++ b/tests/auth.rs
@@ -9,6 +9,7 @@ use select::predicate::Text;
 use std::process::{Command, Stdio};
 use std::thread::sleep;
 use std::time::Duration;
+use reqwest::StatusCode;
 
 #[rstest_parametrize(
     cli_auth_arg, client_username, client_password,
@@ -49,7 +50,7 @@ fn auth_works(
         .send()?;
 
     let status_code = response.status();
-    assert_eq!(status_code.canonical_reason(), Some("OK"));
+    assert_eq!(status_code, StatusCode::OK);
 
     let body = response.error_for_status()?;
     let parsed = Document::from_read(body)?;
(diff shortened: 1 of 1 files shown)


tests/auth.rs
```diff
@@ -43,11 +43,15 @@ fn auth_works(
...
-    let body = client
+    let response = client
         .get(format!("http://localhost:{}", port).as_str())
         .basic_auth(client_username, Some(client_password))
-        .send()?
-        .error_for_status()?;
+        .send()?;
+
+    let status_code = response.status();
+    assert_eq!(status_code.canonical_reason(), Some("OK"));
```
svenstaro: I like this generally, but I'd prefer if we could do it like [here](https://github.com/svenstaro/miniserve/pull/102#discussion_r281189844).

KSXGitHub:
@svenstaro It is okay if I remove [the part where it parses HTML and accesses every file](https://github.com/KSXGitHub/miniserve/blob/67b2ec8e70d0f781b3f57762c1e8e28c2102c15d/tests/auth.rs#L54-L59)

svenstaro:
I see your logic: If it (accessing files) works without auth there is no use in checking it as well when auth comes into the equation. That's a logical point to make.

However, I usually like to test as many things at the same time as I can so we know that certain program logic paths (for instance auth, but maybe also others) do not lead to the program outputting unexpected stuff. For instance, maybe we have some logic that returns OK 200 on good login but still doesn't display any files? Granted, now we don't have this situation, but I'd just like my tests to reassure me that the end-to-end experience that users see (i.e. seeing a list of files) is always working.

I also see that those lines would get tedious after every test. Would you be ok making a helper function with those lines and then just calling that helper function in the appropriate location in every test?

KSXGitHub:
I'll leave them alone then.

> Would you be ok making a helper function with those lines and then just calling that helper function in the appropriate location in every test?

Another time.

Issue #100, opened 2019-05-03
Multiple auth users


KSXGitHub:
## Example

```sh
miniserve --auth user1:password1 --auth user2:sha256:hash
```

svenstaro:
Yeah, sounds like a good idea.

Issue #101, opened 2019-05-03
Configuration file


KSXGitHub:
## Summary

Right now, the only way to "configure" miniserve is via command-line arguments

```sh
miniserve --interface ::1 --port 8080 --auth user:sha256:hash $@
```

But it would be more convenient to have a global configuration file for recurring command-line options

```yaml
interface: "::1"
port: 8080
auth: "user:sha256:hash"
```

## Discussion

* Which format should we use? [RON](https://github.com/ron-rs/ron), TOML or YAML?
* Where to locate configuration files? `~/.miniserverc.$ext`, `~/.miniserve/config.$ext`, `~/.config/miniserve/config.$ext`?
* What to do when configuration file is invalid? Ignore, Warn, Error?
* Which library should we use? (For now, I recommend `serde`)

boastful-squirrel:
What would be the benefit of a configuration vs using CTRL+R/an alias/...?

KSXGitHub:
@boastful-squirrel I do see your point, especially in a UNIX-like system: A UNIX program only does one thing and does it well. But not all systems are UNIX-like (Windows, obviously), and this tool is hardly an ideal UNIX program.

As for the advantage of configuration as opposed for alias/custom-commands: A configuration file is a data file and can be treated like data. It also works regardless of platform (aliases only work on POSIX shell, UNIX commands only works on UNIX-like system and .bat commands only works on Windows).

BTW, what is the "CTRL-R" that you referred to?

boastful-squirrel:
It allows to search for commands in your bash history.

svenstaro:
I have trouble deciding whether I want this or not. The general idea of miniserve is that it's a simple, basically braindead easy way of serving your files in a somewhat convenient way. It grew to allow auth and uploading and now it's supposedly growing to allow for multiple auth providers at the same time.

It's a logical consequence for it to gain config file support and honestly, at this point, I'm not sure that would exceed the complexity I had originally planned for miniserve. Don't get me wrong. I use miniserve in production myself in a big company and therefore it's certainly not just a toy project to me. However, the whole reason I made miniserve is that I wanted a simple tool without too much bloat.

I'm not saying no. I'm saying I'd appreciate some discussion on this.

KSXGitHub:
@svenstaro  Then according to the UNIX philosophy™, miniserve should not have to handle any of this (including reading configuration file, authentication, themes), rather, there should be other helper programs that handle them for you:
[...]

PR #104, opened 2019-05-05
Multiple user accounts

KSXGitHub:
Fix https://github.com/svenstaro/miniserve/issues/100
Close https://github.com/svenstaro/miniserve/pull/118

**TODO:**
  - [ ] Add test
  - [x] ~~Add `ContextualError` to support multiple errors at once~~

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -202,7 +202,7 @@ mod tests {
         case("username:sha256:abcd", "username", "abcd", "sha256"),
         case("username:sha512:abcd", "username", "abcd", "sha512")
     )]
-    fn parse_auth_valid(auth_string: &str, username: &str, password: &str, encrypt: &str) {
+    fn parse_single_auth_valid(auth_string: &str, username: &str, password: &str, encrypt: &str) {
         assert_eq!(
             parse_auth(auth_string).unwrap(),
             create_required_auth(username, password, encrypt),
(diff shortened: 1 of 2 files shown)


boastful-squirrel:
If @svenstaro validates that this is a useful feature, I think the implementation can be more simple. In particular, I'm pretty sure you could setup `StructOpt` so it takes a `Vec` of `RequiredAuth`, just like we do for the interfaces. Then, you would just have to adapt the middleware so that if at least 1 authentication succeeds, then it's fine.

KSXGitHub:
> I'm pretty sure you could setup `StructOpt` so it takes a `Vec` of `RequiredAuth`

I had considered this too, but the time it takes increase as number of users increases (aka O(n) complexity), so I chose to use `HashMap` which has constant time evaluation.

svenstaro:
I believe that for any number of users miniserve will ever handle, we don't
have to worry about complexity even if it were O(n!) :)

KSXGitHub:
> I believe that for any number of users miniserve will ever handle

The user will decide how many accounts they need. I don't want to prevent them from doing what they want.

I have to admit that a linear list of `RequiredAuth` is easier to implement, but it will also lead to some interesting consequences:
  * It would be possible for a username to have more than one password.
  * It would compare two strings for each user, as opposed to comparing two hashed key (which is, by default, [a pair of two `u64`s](https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#3139-3142)). This would make the use of `Vec<RequiredAuth>` even more expensive.

boastful-squirrel:
Admitting that would be useful (I don't think so, I like the mini part of miniserve), then reading accounts from a file or from a DB would be far more suitable. If we're talking about so much accounts that complexity matters, I wouldn't like to be the guy who has to type the command to start miniserve.

KSXGitHub:
> then reading accounts from a file or from a DB would be far more suitable

Well, I made [this request](https://github.com/svenstaro/miniserve/issues/101)

boastful-squirrel:
Note that according to the spec, an HTTP basic authentication password can contain any characters, which includes spaces and `:`.

For instance, `foo:bar xxx:yyy` should be read as `{user: "foo", password: "bar xxx:yyy"}`

KSXGitHub:
> oh, it seems it doesn't work on `master` anymore. That's probably a regression caused by the handling of hashed passwords.

A necessary evil I think. We can still support escape characters though, but that would make the code more convoluted.

> Handling it through `StructOpt` is probably your only option.

I will try. But I'll still convert it into `HashMap` until I see a good reason not to.

KSXGitHub:
@boastful-squirrel Wait a minute, I don't understand how `StructOpt` is going to solve the problem you described, which in my opinion is more of a parser problem (`parse_auth`).

boastful-squirrel:
If `--auth` takes a `Vec<RequiredAuth>` with `number_of_values=1`, just like the interfaces, you should be able to write something like this: `miniserve -a xxx:yyy -a foo:bar -a bar:baz`. The `parse_auth()` method won't have to be changed since it will be applied to each of those credentials individually. The middleware will require some modifications though.

KSXGitHub:
@boastful-squirrel So you want [this](https://github.com/svenstaro/miniserve/pull/118)?

PR #112, opened 2019-05-10
Fix Clippy warnings + cargo fmt

boastful-squirrel:
So, I just ran `cargo fmt` and fixed `clippy`'s warning the easiest way possible, just so that master turns green. I think that in the case of too many arguments, allowing it is probably cleaner than creating specific struct just for that, especially for the `render_error` method, which must be called a dozen times... But if you decide that a struct would be cleaner, that's not an expensive fix to do, my priority was just to make master turn green.


svenstaro:
Cool stuff.

PR #128, opened 2019-05-29
refactor code

robatipoor:


src/main.rs
```diff
@@ -83,7 +83,9 @@ fn run() -> Result<(), ContextualError> {
         && miniserve_config
             .path
             .symlink_metadata()
-            .map_err(|e| ContextualError::IOError("Failed to retrieve symlink's metadata".to_string(), e))?
+            .map_err(|e| {
+                ContextualError::IOError("Failed to retrieve symlink's metadata".to_string(), e)
+            })?
```
KSXGitHub: You just split this line into 3, is this necessary?

src/themes.rs
```diff
@@ -17,14 +62,13 @@ impl ColorScheme {
     /// Returns the URL-compatible name of a color scheme
     /// This must correspond to the name of the variant, in lowercase
     /// See https://github.com/svenstaro/miniserve/pull/55 for explanations
-    pub fn to_slug(self) -> String {
+    pub fn to_slug(self) -> &'static str {
```
KSXGitHub: I like this change.

src/main.rs
robatipoor: rustfmt auto split line to 3

KSXGitHub: It seems that you need to undo that to please Clippy

https://travis-ci.org/svenstaro/miniserve/jobs/538610810

src/archive.rs
```diff
@@ -97,12 +99,12 @@ fn tar(
...
     // Recursively adds the content of src_dir into the archive stream
     tar_builder
-        .append_dir_all(inner_folder, &src_dir)
+        .append_dir_all(inner_folder, src_dir.as_ref())
         .map_err(|e| {
             ContextualError::IOError(
                 format!(
                     "Failed to append the content of {} to the TAR archive",
-                    &src_dir
+                    src_dir.as_ref().to_str().unwrap()
```
svenstaro: I would prefer not to have this `unwrap()`. Can you choose a sensible default? I know it's very unlikely to fail, but still.

robatipoor: I have no idea except using `unwrap_or("file")` If you think good, I'll change it

svenstaro:
Very cool, I like these changes! Thanks.

PR #138, opened 2019-06-12
Enable streaming tarball download

### LINKED ISSUE #2: Streaming zip/tar of directories

gyscos:
This PR:

* Adds `src/pipe.rs` to provide a `Write` implementation on a `mpsc::Sender<Bytes>`.
* Make `CompressionMethod: Copy`
* Updates `src/archive.rs` to take a `W: Write` as argument rather than returning a byte array.
* Updates `src/listing.rs` to, when a archive is requested:
    * Create a channel of `Bytes` (with a capacity of 10 to give some buffer between compression and upload)
    * Start a new thread, and run `create_archive` in this thread with the `Write` side of the pipe
    * Return an actix response streaming the read side of the pipe
* Adds `CompressionMethod::Tar` for uncompressed archive

The net result is that archive download (either tar or tar.gz) now starts immediately and uses a more reasonable amount of memory on the host. In particular, downloading folders larger than the memory of the host is now possible.

Fixes #2

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -43,6 +43,11 @@ impl CompressionMethod {
         }
     }
 
+    /// Make an archive out of the given directory, and write the output to the given writer.
+    ///
+    /// Recursively includes all files and subdirectories.
+    ///
+    /// If `skip_symlinks` is `true`, symlinks fill not be followed and will just be ignored.
     pub fn create_archive<T, W>(
         self,
         dir: T,
@@ -61,6 +66,7 @@ impl CompressionMethod {
     }
 }
 
+/// Write a gzipped tarball of `dir` in `out`.
 fn tar_gz<W>(dir: &Path, skip_symlinks: bool, out: W) -> Result<(), ContextualError>
 where
     W: std::io::Write,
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -32,7 +32,7 @@ htmlescape = "0.3.1"
 bytesize = "1.0.0"
 nanoid = "0.2.0"
 alphanumeric-sort = "1.0.6"
-structopt = "0.2.17"
+structopt = "0.2.16"
```
svenstaro: Do we really need this particular change here?

gyscos:
0.2.17 was yanked, 0.2.16 is the latest compatible version. We could bump to 0.3.0 in another pr.

gyscos:
Oh wait apparently I forgot to clean up a few things, will do that first thing tomorrow.

gyscos:
Cleaned up the code a bit.

Regarding `structopt`, the version being yanked was breaking `cargo update`, but I guess it doesn't break regular builds, so we can stay on `0.2.17` if you prefer.

svenstaro:
I already fixed the version in master so you can just rebase.

src/main.rs
```diff
@@ -79,17 +80,21 @@ fn run() -> Result<(), ContextualError> {
         TermLogger::init(LevelFilter::Error, Config::default())
     };
 
-    if miniserve_config.no_symlinks
-        && miniserve_config
+    if miniserve_config.no_symlinks {
+        let is_symlink = miniserve_config
```
gyscos: Clippy was complaining about the "complex condition being evaluated in the `if` block", so I split it out and CI now passes.

gyscos:
Rebased

src/listing.rs
```diff
@@ -235,46 +235,49 @@ pub fn directory_listing<S>(
...
+                log::error!("Error during archive creation: {:?}", err);
             }
-        }
+        });
+
+        // `<rx as Stream>::Error == ()` but we want `actix_web::error::Error`
+        let rx = rx.map_err(|_| {
+            ContextualError::CustomError("pipes never fail".to_string())
+                .compat()
+                .into()
```
gyscos: Note that I'm not super familiar with error conversions, so there may be an easier way to create an actix error.

gyscos: Using `unreachable!` instead. [`mpsc::Receiver` never return errors](https://docs.rs/futures/0.1.27/src/futures/sync/mpsc/mod.rs.html#923-958).

gyscos:
I can remove the `is_symlink` change in `main.rs` if you'd prefer to do it separately.

svenstaro:
Sorry for taking so long to review this. I'll get to it!

gyscos:
No worries! I know how easy it is to get busy with something else...

src/pipe.rs
```diff
@@ -0,0 +1,40 @@
+use bytes::{Bytes, BytesMut};
+use futures::sink::{Sink, Wait};
+use futures::sync::mpsc::Sender;
+use std::io::{Error, ErrorKind, Result, Write};
+
+pub struct Pipe {
```
svenstaro: Currently, this whole module has no documentation. Could you add some? Even if the purpose seems kind of clear.

src/main.rs
svenstaro: That's fine. I generally prefer formatting and refactoring stuff like this in a separate PR, though.

src/archive.rs
```diff
@@ -1,79 +1,90 @@
...
             CompressionMethod::TarGz => ContentEncoding::Gzip,
+            CompressionMethod::Tar => ContentEncoding::Identity,
         }
     }
-    /// Creates an archive of a folder, using the algorithm the user chose from the web interface
-    /// This method returns the archive as a stream of bytes
-    pub fn create_archive<T: AsRef<Path>>(
-        &self,
+
+    pub fn create_archive<T, W>(
```
svenstaro: Since this is rather abstract and complex now, I'd definitely like some documentation here.

svenstaro:
So I had a proper look. This looks amazing! It works really and appears to handle all the errors I could produce properly which is great. The only things I found are fairly superficial in nature so it should be a breeze for you to fix those.

src/listing.rs
```diff
@@ -1,7 +1,7 @@
-use actix_web::http::StatusCode;
-use actix_web::{fs, http, Body, FromRequest, HttpRequest, HttpResponse, Query, Result};
+use actix_web::{fs, Body, FromRequest, HttpRequest, HttpResponse, Query, Result};
 use bytesize::ByteSize;
-use futures::stream::once;
+use failure::Fail;
```
svenstaro: This can be removed.

svenstaro:
After this, the next big step is actix-web 1.0 :s

src/main.rs
gyscos: Yeah, I would have skipped it if CI hadn't complained.

src/archive.rs
gyscos: Added documentation for each step - though there's not _that much_ difference with the previous functions.

gyscos:
Thanks for the review!

Added some doc/comments to help describe what's going on.

src/listing.rs
gyscos: Whoops, not sure when I added that.

svenstaro:
Merging as is! Super nice.

Issue #142, opened 2019-06-24
Make whole row clickable in mobile view


svenstaro:
It's a bit tricky to click the right element in mobile view on an actual smartphone. It might be nice if the whole row would react to the click in that case.

bjeanes:
#158 implies this is complete, but it doesn't appear to be the case in my own testing. Is this the case?

PR #158, opened 2019-07-04
accomplish #142

wyhaya:
1.  Accomplish #142
2. URL optimization
```
Good : http://127.0.0.0:8080/src/
Bad  : http://127.0.0.0:8080/src
```

### DIFF
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -138,7 +138,6 @@ pub fn directory_listing<S>(
     let base = Path::new(serve_path);
     let random_route = format!("/{}", random_route.unwrap_or_default());
     let is_root = base.parent().is_none() || req.path() == random_route;
-    let page_parent = base.parent().map(|p| p.display().to_string());
     let current_dir = match base.strip_prefix(random_route) {
         Ok(c_d) => Path::new("/").join(c_d),
         Err(_) => base.to_path_buf(),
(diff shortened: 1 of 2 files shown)


svenstaro:
Thanks! What's the reasoning behind 2.?

wyhaya:
@svenstaro 

There is no standard for this, I think he should have "/" when our target is a `folder`.
This makes the `url` easier to understand.

In `nginx`:

```bash
↪ curl https://example.com/imgs -I
HTTP/2 301 
server: nginx/1.15.9
...
location: https://example.com/imgs/
```

src/renderer.rs
```diff
@@ -218,6 +218,20 @@ fn archive_button(
     }
 }
 
+// Append to the string after the link
+fn get_dir_last(link: &str) -> &str {
+    match link.chars().last() {
+        Some(d) => {
+            if d == '/' {
+                ""
+            }else {
```
svenstaro: Add a space between `}` and `else`.

wyhaya:
1. Fix "Parent Directory"
2. Fix style
3. Add redirect to directory

wyhaya:
Redirect to directory is required. Last time, the problem of clicking the link was solved. If you do not add a redirect, the user directly accessing http:127.0.0.1:8080/src will cause a link error.

src/listing.rs
```diff
@@ -285,14 +284,26 @@ pub fn directory_listing<S>(
...
     } else {
+        // Redirect to directory
+        if !renderer::has_trailing(&serve_path) {
+            let query = match req.query_string() {
+                "" => String::new(),
+                _ => format!("?{}", req.query_string())
+            };
+            return Ok(
+                HttpResponse::MovedPermanenty()
+                    .header("Location", format!("{}/{}", serve_path, query))
```
svenstaro: Let's make it temporary. With Moved Permanently, you give browsers strong caching guarantees but what if the file system changes?

wyhaya: The file pointed to by '/src' '/src/' is consistent, and ‘/src/’ will respond correctly even if the file changes.

svenstaro: But what if `/src` becomes a file instead of a dir?

wyhaya: This may cause the url to increase by ‘/’ but will not affect normal use
I try other web servers like nginx, they are the same

```
Client -> /src -> rewrite -> /src/ (dir) -> Ok 
```

```bash
mv src out && echo 123 > src
```

No cache
```
Client -> /src -> /src (file) -> Ok
```

Cache
```
Client -> /src -> rewrite -> /src/ (file) -> Ok
```
`miniserve` always responds correctly

Issue #156, opened 2019-07-02
Suggestion: Serving index.html, index.xml, index.xhtml, index.htm


KSXGitHub:
## Why?

I want to serve some static website but I don't want to manually find `index.html` in a list of dozen items.

## Suggestion 1

Introduce a `--homepage index.html` option which instructs miniserve to serve `index.html` when the file is found within the folder.

## Suggestion 2

If `index.*` file is found within a folder, display a button next to "Download .tar.gz" that links to said file.

svenstaro:
I quite like suggestion 1. Should be easy to implement, too.

zicklag:
What do you think about adding a `-I` arg that would be a shortcut for `--index index.html`. I almost always want to have the default index, so having a short arg specifically for it would be nice.

PR #189, opened 2019-09-06
add a default index serving option (#156)

efx:
Start addressing #156 

~~Do you know where I should consume this option and how to configure active-web to serve the file? I started modifying [`fn configure_app(`](https://github.com/svenstaro/miniserve/blob/master/src/main.rs#L222) but couldn't figure out how to make it serve an `index.html` from within a configured directory root.~~


efx:
After reading the documentation and codebase for a bit I figured it out. :)
The main things that remain are:
- allow configuring the name of the file `index.html`? Or simply let the user define the value instead of this being a boolean.
- `default-index` an ok name? It is shortened to `-d` which is confusing since it other utilities it means `daemon`.
- tests

svenstaro:
Would probably be nice to be able to define its name with a proper default.

I don't think it should have a shorthand as it's a fairly "dangerous" option. Better not get the options mixed up. How about just `--index`?

efx:
> I don't think it should have a shorthand as it's a fairly "dangerous" option. Better not get the options mixed up. How about just --index?
Good point. I'll update it to that and make it configurable with the default of `index.html`.

svenstaro:
@efx Is this still WIP or shall I take a look?

efx:
@svenstaro feel free to read through and give feedback.
The last thing I think it could use is adding a whitelist for the index file type.  In the current form it will serve up any file you pass. I've not tested with arbitrary files but it works well with `.html`.

Lucretiel:
Status on this? I'd love to switch to miniserve as my default way of browsing `rustdoc` documentation

efx:
@Lucretiel I have tested it locally serving other projects and it worked well. I have held back given it could use a bit more testing, but that is probably needless caution. 
So it just needs a good review. Happy to address anything you see; I'm a Rust beginner :smile: 
/cc @svenstaro

Lucretiel:
I'd propose adding to the docstring:

```
/// The name of a directory ndex file to serve, like "index.html"
///
/// Normally, when miniserve serves a directory, it creates a listing for that directory.
/// However, If a directory contains this file, miniserve will serve that file instead.
```

svenstaro:
Sorry for taking so incredibly long on this. I'll move it way up on my todo, promise!

svenstaro:
Merged, thanks. And again sorry for taking so long. Will make a release with this soon-ish.

efx:
> And again sorry for taking so long. Will make a release with this soon-ish.

I totally understand @svenstaro. Thank you for making this project!

ngirard:
@efx, thanks for this pull request !
@svenstaro, thanks for this nice project ! A new release would indeed be great !

svenstaro:
I want to make a release once the other PRs are in. Shouldn't take too much work to fix them up. :)

ngirard:
Thanks for the release, @svenstaro !

svenstaro:
Yeah, I decided that probably it's a good idea to just make a release now instead of waiting for too long.

PR #208, opened 2019-10-13
Add abillity to set path prefix

mkroman:
This PR should replace the `--random-route` argument with both a `--path-prefix` argument and a conflicting `--random-path-prefix` argument.

I use miniserve behind a load-balancer under a specific path, so I need to control the path in miniserves web server, but while I can show the index page in a subdirectory, the links in the file index aren't relative to the current URL.

[Here's an example](https://web.archive.org/web/20191013061658/https://maero.dk/pub) - the indexed file links don't work because it assumes the relative URL path is `/` and it will link to the root domain accordingly (i.e. not under `/pub/`)

The `--random-path-prefix` argument should be a fully functional replacement to `--random-route`. More testing may be required.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -51,9 +51,9 @@ struct CLIArgs {
     )]
     auth: Vec<auth::RequiredAuth>,
 
-    /// Generate a random 6-hexdigit route
-    #[structopt(long = "random-route")]
-    random_route: bool,
+    /// Use a specific path prefix
+    #[structopt(long = "path-prefix")]
+    path_prefix: Option<String>,
 
     /// Do not follow symbolic links
     #[structopt(short = "P", long = "no-symlinks")]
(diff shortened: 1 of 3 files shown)


svenstaro:
Sorry, I know I've been slacking on this. Could you rebase this for recent master changes, please?

svenstaro:
@mkroman  In addition to rebasing this, is there any chance I can convince you to make a test for this? :)

bjeanes:
It seems like the author of this PR has probably abandoned the effort, but I'd like to weigh in that I also have been trying to run miniserve behind a reverse proxy with a path prefix (which, unfortunately, is required). Any possibility that this is something you'd likely consider doing @svenstaro ?

svenstaro:
Sure, I'd still like for something like this to go in. Do you wanna take a stab at it, @bjeanes?

bjeanes:
Yeah I'd consider taking a stab on the weekend if I don't spend the time away from computer. But in full transparency, I opted to use `rclone serve http` with its base path option since leaving my comment so my immediate need has diminished. _However_ I definitely would like the tgz/zip feature for this use case, so there is still some motivation :)

svenstaro:
Anyone interested in cleaning this up, rebasing it, and getting it in? :)

svenstaro:
Closing in favor of #579.

PR #228, opened 2019-12-02
Implement raw view to allow recursive downloading via wget

Jikstra:
Currently it's not easily possible to use wget to download a folder served with miniserve because the rendered directory listing contains links for the different .tar options, themes and whatnot.
- This implements a raw view which can get accessed via `http://.../somedir/foo/?raw` and is a stripped down html version
- Besides this it also implements a new ui element which contains a generated wget command to download this folder via wget

This is probably a hackish attempt, but works quite well. If there's a chance of this feature getting merged, i would overhaul it, add integration tests and whatnot.

### DIFF
--- a/tests/download_folder.rs
+++ b/tests/download_folder.rs
@@ -1,0 +1,91 @@
+mod fixtures;
+mod utils;
+
+use assert_cmd::prelude::*;
+use assert_fs::fixture::TempDir;
+use fixtures::{port, tmpdir, Error};
+use pretty_assertions::{assert_eq};
+use rstest::rstest;
+use select::document::Document;
+use std::process::{Command, Stdio};
+use std::thread::sleep;
+use std::time::Duration;
+use select::predicate::Class;
+use select::predicate::Name;
+
+#[rstest]
+/// The ui displays the correct wget command to download the folder recursively
+fn ui_displays_wget_element(tmpdir: TempDir, port: u16) -> Result<(), Error> {
+    let mut child = Command::cargo_bin("miniserve")?
+
+        .arg("-p")
+        .arg(port.to_string())
+        .arg(tmpdir.path())
+        .stdout(Stdio::null())
+        .spawn()?;
+
+    sleep(Duration::from_secs(1));
+    // Ensure the links to the archives are not present
+    let body = reqwest::blocking::get(format!("http://localhost:{}", port).as_str())?
+        .error_for_status()?;
+    let parsed = Document::from_read(body)?;
+    let wget_url = parsed.find(Class("downloadWget")).next().unwrap().find(Name("pre")).next().unwrap().text();
+    assert_eq!(wget_url, format!("wget -r -c -nH -np --cut-dirs=0 -R \"index.html*\" http://localhost:{}/?raw=true", port));
+
+    let body = reqwest::blocking::get(format!("http://localhost:{}/very/deeply/nested/", port).as_str())?
+        .error_for_status()?;
+    let parsed = Document::from_read(body)?;
+    let wget_url = parsed.find(Class("downloadWget")).next().unwrap().find(Name("pre")).next().unwrap().text();
+    assert_eq!(wget_url, format!("wget -r -c -nH -np --cut-dirs=2 -R \"index.html*\" http://localhost:{}/very/deeply/nested/?raw=true", port));
+
+    
+    child.kill()?;
+    Ok(())
+}
+
+#[rstest]
+/// All hrefs in raw mode are links to directories or files & directories end with ?raw=true
+fn raw_mode_links_to_directories_end_with_raw_true(tmpdir: TempDir, port: u16) -> Result<(), Error> {
+    let mut child = Command::cargo_bin("miniserve")?
+
+        .arg("-p")
+        .arg(port.to_string())
+        .arg(tmpdir.path())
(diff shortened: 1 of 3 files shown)


svenstaro:
Sorry, I know I've been slacking on this. Could you rebase this for recent master changes, please?

svenstaro:
In addition to rebasing this, could you perhaps also write a quick test for this? Would be amazing. :)

Jikstra:
> Sorry, I know I've been slacking on this. Could you rebase this for recent master changes, please?

Don't worry, I've been slacking on this too! I'm currently doing some digital detoxing but as I expect to have corona lockdown in my region quite soon I will find some time there to fix up this pr :) Thanks for maintaining!

Jikstra:
@svenstaro finally found the time to rebase this pr and add tests. As I did cargo fmt/clippy in between, it probably is best reviewed commit by commit :)

svenstaro:
Could you completely remove your `Cargo.lock` from the changeset? I'd like to generate the lockfile on my end. It'll also cut down on the number of changes.

Jikstra:
@svenstaro resetted it to the version on master, not sure how i would really remove the changes from this pr :thinking:

svenstaro:
Sorry, could you rebase again? I fixed some clippy lints and ran `cargo fmt`.

Jikstra:
Si claro :) It's done :)

svenstaro:
I think you messed up a bit. The `Cargo.lock` is back with a fury!

svenstaro:
Since you're already at it: Could you shrink this down to two pristine commits? One for the tests and one for the rest. Alternatively, just have a single commit. I think the various commits don't really add any context for this particular PR.

Jikstra:
I think i now got it ^^

Jikstra:
@svenstaro anything blocking this to get merged?

tests/download_folder.rs
```diff
@@ -0,0 +1,91 @@
...
+                assert!(true);
+            } else {
+                println!("This node is a link and neither of class directory, root or file: {:?}", node);
+                assert!(false);
+            }
+        }
+    }
+
+    let urls = [
+        format!("http://localhost:{}/?raw=true", port),
```
svenstaro, line 76: Wouldn't it be nicer to use a parametrized test (via rstest) here instead of this loop?

Jikstra:
> Great work and good idea. Thanks for the PR!

Thank you for miniserve :)

> I think it would good if no UI rendered at all if in raw mode. I imagine just raw links to the files.

I like the idea, even if it's maybe some not really needed optimisation? Or what do you think is the benefit of stripping everything besides the bare minimus (besides smaller requests).

> I think code references to wget should be removed. It's quite conceivable that someone might use this with other recursive HTTP downloaders. How about you change references to wget and call them raw_mode (or likewise) instead?

Makes sense!

> The parsing of the parameter is currently a bit weird I think. ?raw=true doesn't work and just ?raw causes these to be printed on the terminal:

Any ideas where this could come from? Not really a rust coder nor experienced with the http lib.

> I like the idea of informing users about a suggested wget command but with the UI gone, how about just printing a suggested at the end of the output or only having one in the README or something like that?

I think it would be nice if we put it in the footer. Having it generated makes sense in my opinion, because the wget command is not really intuitive (cutting the depth of the folders and all that stuff...). And i think supporting wget makes sense as it's more or less everywhere installed or you can install it without a problem (ignoring windows).

tests/download_folder.rs
Jikstra: Oh yes! Will look into it :)

svenstaro:
> > Great work and good idea. Thanks for the PR!
> 
> Thank you for miniserve :)
> 
> > I think it would good if no UI rendered at all if in raw mode. I imagine just raw links to the files.
> 
> I like the idea, even if it's maybe some not really needed optimisation? Or what do you think is the benefit of stripping everything besides the bare minimus (besides smaller requests).
> 

I'm just thinking that if you use this in a polling fashion in some other kind of recursive HTTP explorer that it would be more CPU and bandwidth-efficient if we didn't render and then send stuff which the other side doesn't even want anyway.

> > I think code references to wget should be removed. It's quite conceivable that someone might use this with other recursive HTTP downloaders. How about you change references to wget and call them raw_mode (or likewise) instead?
> 
> Makes sense!
> 
> > The parsing of the parameter is currently a bit weird I think. ?raw=true doesn't work and just ?raw causes these to be printed on the terminal:
> 
> Any ideas where this could come from? Not really a rust coder nor experienced with the http lib.
> 



> > I like the idea of informing users about a suggested wget command but with the UI gone, how about just printing a suggested at the end of the output or only having one in the README or something like that?
> 
> I think it would be nice if we put it in the footer. Having it generated makes sense in my opinion, because the wget command is not really intuitive (cutting the depth of the folders and all that stuff...). And i think supporting wget makes sense as it's more or less everywhere installed or you can install it without a problem (ignoring windows).

Sure, let's put it in the footer. :)

svenstaro:
Hey @Jikstra are you still interested in getting this in?

Jikstra:
@svenstaro I'm currently reimplementing it on top of master with all the requested changes :)

Jikstra:
Closing in favor of https://github.com/svenstaro/miniserve/pull/508

PR #241, opened 2019-12-29
More descriptive instructions for nightly

rofrol:
I have added `-f` so that people who already have it, will update it.


svenstaro:
This is now not actually required any more thanks to upstream updates.

PR #292, opened 2020-04-19
Add packaging status

j2ghz:
Adds a packaging status badge so people can easily see which repositories have an up-to-date version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ Sometimes this is just a more practical and quick way than doing things properly
 
 ## How to install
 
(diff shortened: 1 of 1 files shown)


svenstaro:
Could you switch to markdown-style instead of plain HTML?

j2ghz:
I can, but then it can't be right-aligned. I thought it would be nicer if it doesn't take up more space. But I can do either, it's your choice.

svenstaro:
Sorry for taking so long on this. I tested this and your suggestion actually looks really good. Merging as is, thanks a lot!

PR #297, opened 2020-05-03
Add zip download functionality for windows users

marawan31:
This is quite simple, adding the ability to download folders as zip files instead of the alternatives.
This is mainly to help windows users download directories without the need to also install 7z or a similar app.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -593,6 +593,24 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "bzip2"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bzip2-sys 0.1.8+1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.67 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "bzip2-sys"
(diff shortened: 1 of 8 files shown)


svenstaro:
Thank you for the PR! Looks really neat. One important thing though: This doesn't seem to do streaming and so the browser can't start downloading the zip befor e it is entirely done zipping. This is in contrast to how our `.tar.gz` download behaves. Could you try to convert it to be streaming?

marawan31:
This is hard to do because of the ZIP file format itself. The library used here uses the seek trait (because of the zip limitation i mentioned) to be able to modify the file, so streaming will not work with this.

svenstaro:
Hm I did some research and I'll have to concede that you are correct. Still, the implications aren't great as from my understanding, you can potentially kill the service this way by zipping a large directory. I do see the convenience of zip for some users, though.

How about as a compromise, we document this problem with the zip archive somewhere user-visible and default to archiving-off by default and then add `--enable-zip` and `--enable-tar`? This seems sane and safe as a default.

marawan31:
Yes, too large of a file can break the service and cause issues. How about using a temporary file that gets automatically deleted using the tempfile crate (only for zip)?

svenstaro:
A temporary file can cause issues as well. I think I'd prefer it the way it is but change the archiving settings to be opt-in. Also document the problematic behavior of zip in the parameter help and then we're good.

svenstaro:
Seems pretty good to me, thanks!

PR #327, opened 2020-06-28
Bind to random available port when port 0 specified

### LINKED ISSUE #81: Handle --port=0
The current behaviour of `miniserve` if set to `--port=0` is to accept it (valid `u16`). However, this is not a "valid" port (actually, it is valid, but shouldn't be used as such), an address like `127.0.0.1:0` is not valid .

Chrome's message:
```
This site can’t be reached
The web page at http://127.0.0.1:0/ might be temporarily down or it may have moved permanently to a new web address.
ERR_ADDRESS_INVALID
```

By convention (I don't know if there's a spec for it ?), addressing a system on port 0 means "give me whatever port you want", usually from 49152 to 65535.

So, we have 2 solutions here:
- either we make the port 0 invalid, exception caught at startup, no big deal
- or we implement such a feature if you think that's worth it: generate a random port from 49152 to 65535 and use it. Note that Python's `http.server` implements this feature

In short, either a bug fix, or a new feature :)

parrotmac:
Example of output:

```
miniserve . --port 0
miniserve v0.7.0
Serving path /some/neat/path at http://[::]:45159, http://127.0.0.1:45159

Quit by pressing CTRL-C
```

Should fix #81


Cargo.toml
```diff
@@ -27,6 +27,7 @@ actix-web = "0.7"
 simplelog = "0.8"
 base64 = "0.12"
 percent-encoding = "2.1"
+port_check = "0.1"
```
svenstaro, line 30: We now have this listed twice. Can you just remove it from the dev deps below?

parrotmac: Oh, definitely! Done.

tests/serve_request.rs
```diff
@@ -69,6 +70,31 @@ fn serves_requests_with_non_default_port(tmpdir: TempDir, port: u16) -> Result<(
...
+
+    sleep(Duration::from_secs(1));
+    child.kill()?;
+
+    let output = child.wait_with_output().expect("Failed to read stdout");
+    let all_text = String::from_utf8(output.stdout)?;
+
+    let re = Regex::new(r"http://127.0.0.1:(\d+)").unwrap();
+    let caps = re.captures(all_text.as_str()).unwrap();
+    let port_num = caps.get(1).unwrap().as_str().parse::<u32>().unwrap();
```
svenstaro, line 90: Shouldn't this be parsed as a `u16` in which case the assert down below can be omitted because that's exactly the `u16` value range?

tests/serve_request.rs
```diff
@@ -9,6 +9,7 @@ use select::node::Node;
 use std::process::{Command, Stdio};
 use std::thread::sleep;
 use std::time::Duration;
+use regex::{Regex};
```
svenstaro, line 12: You don't need the `{}` here as it's just a single item.

tests/serve_request.rs
parrotmac: For sure; updated!

PR #329, opened 2020-06-29
Remove 'bytesize' crate

wyhaya:


svenstaro:
Could you provide some reasoning?

wyhaya:
@svenstaro I think this function is very simple, there is no need to introduce other crate, Now `miniserve` is already very `heavy`, Installation and operation is very slow, we should reduce crate as much as possible.

src/renderer.rs
```diff
@@ -365,6 +365,25 @@ fn entry_row(
     }
 }
 
+fn bytes_to_size(bytes: u64) -> String {
+    const UNITS: [&str; 7] = ["B", "KB", "MB", "GB", "TB", "PB", "EB"];
```
svenstaro, line 369: Isn't this technically incorrect as these really appear to be the SI units instead since you use 1024 everywhere? I'm definitely ok with using SI units but we should call them correctly, too.

svenstaro:
I generally agree  with your approach and like minimizing deps. However, in this particular case, I expect the greater benefit actually exists in just leaving bytesize in. I just checked the crate: It has no dependencies on its own and builds in 0.19s. Considering that we might want to add a switch to miniserve later to toggle between showing SI and non-SI units, I think shaving off the mental overhead that this additional code brings into the crate actually is a good benefit.

The thing that we really have to be wary of is those deps that come with tons of recursive deps themselves.

For your information, I'm currently rewriting miniserve a bit to work on modern async Rust with newer tokio, futures and actix-web. It will also compile on stable and not use maud. I'll pay special attention to minimize deps and speed up build times.

wyhaya:
Ok, I see

Issue #45, opened 2019-02-27
How about a QR-code in the terminal?


FrontMage:
I mean it would be handy for phones to access. Maybe a QR-code in the HTML page as well?

svenstaro:
Yeah, I think that'd be cool and helpful.

boastful-squirrel:
That's a cool idea. How would we decide on which IP the QR-code would point ? An argument like `qr-code=my.public.ip.address` maybe ?

svenstaro:
Maybe we can figure out the local domain (from /etc/resolv.conf `domain`) and hostname and construct the full local hostname that way?

whmountains:
I think we could borrow a lot from [Quair](https://crates.io/crates/qair).  Specifically local IP address detection and how to display a QR code for displaying in the terminal.

[qrcodegen](https://crates.io/crates/qrcodegen) looks like a great option for generating the SVG file for use in the web interface.

svenstaro:
Yeah, quair seems really neat. I wish it were a library. The [IP stuff](https://code.willemp.be/willem/qair/src/branch/master/src/myip.rs#L10) seems easy enough.

H-M-H:
There also is https://github.com/kennytm/qrcode-rust which does image, svg and text encoding of QR-codes.

svenstaro:
We now have a QR code in the HTML. We don't yet have one on the terminal but I'd be open for that option, too.

wyhaya:
@svenstaro 

If the user uses the `-i` option, the content of the QR code should be `http://{-i}:{-p}`.

If the user does not use the `-i` option, then search for `interfaces`, then the content of the QR code should be `http://{interface}:{-p}`.

Looks like this:

```
miniserve v0.7.0
Serving path .../miniserve at http://[::]:8080, 
http://127.0.0.1:8080
██████████████      ██  ██████████  ██████████████
██          ██  ██  ██  ██  ██      ██          ██
██  ██████  ██        ██  ██████    ██  ██████  ██
██  ██████  ██  ██        ██  ██    ██  ██████  ██
██  ██████  ██  ██  ██████      ██  ██  ██████  ██
██          ██  ████████      ████  ██          ██
██████████████  ██  ██  ██  ██  ██  ██████████████
                  ██  ██  ████
  ██  ████  ██████████        ██  ██  ██████  ██
██    ██  ██      ████          ██    ████  ██████
██  ████    ██    ██  ██      ██    ██████  ██  ██
[...]

PR #330, opened 2020-06-29
Add generate QR code

wyhaya:
preview

#45


svenstaro:
Also this needs a rebase.

svenstaro:
Do you also want to tackle the task of printing a QR code in the terminal if requested via option flag? That was part of issue #45 but it should likely be done in a separate PR if you want to do it.

wyhaya:
> I think there should be a CLI switch to disable this function (to keep the UI tidy for people who don't like the visual bloat).

It may be considered to add a CLI switch, but is there too many switches at present, it is difficult for users to understand each function, maybe we can make the style smaller?

> Could you add a parametrized test for this that checks multiple values for the qrcode param and see whether you can trigger the error, too?

Ok i will add as soon as possible.

> Do you also want to tackle the task of printing a QR code in the terminal if requested via option flag? That was part of issue #45 but it should likely be done in a separate PR if you want to do it.

Yes, but what characters should we use? The terminal cannot be completely set to a square, because it is impossible to control the user’s ‘line-height’, it is usually difficult to scan successfully

Reference: [qrcode-rust](https://github.com/kennytm/qrcode-rust#string-generation)

svenstaro:
> > I think there should be a CLI switch to disable this function (to keep the UI tidy for people who don't like the visual bloat).
> 
> It may be considered to add a CLI switch, but is there too many switches at present, it is difficult for users to understand each function, maybe we can make the style smaller?

I definitely am worried about making the tool too complex as it touts itself as being simple. However, I think at current we can handle a `--no-qr` (or similar) flag just fine.

> 
> > Do you also want to tackle the task of printing a QR code in the terminal if requested via option flag? That was part of issue #45 but it should likely be done in a separate PR if you want to do it.
> 
> Yes, but what characters should we use? The terminal cannot be completely set to a square, because it is impossible to control the user’s `line-height`, it is usually difficult to scan successfully
> 
> Reference: [qrcode-rust](https://github.com/kennytm/qrcode-rust#string-generation)

True, though its [unicode mode](https://github.com/kennytm/qrcode-rust#unicode-string-generation) works well enough for me so let's just roll with that for the time being.

svenstaro:
FWIW the unicode string mode is gapless for me.

wyhaya:
I added the `--qrcode` option, which is `false` by default. Do you think it is appropriate?

src/renderer.rs
```diff
@@ -133,24 +134,33 @@ fn color_scheme_selector(
...
     html! {
         nav {
-            ul {
-                li {
-                    a.change-theme href="#" title="Change theme" {
-                        "Change theme..."
+            @if show_qrcode {
+                div {
+                    p onmouseover="document.querySelector('#qrcode').src = `/?qrcode=${encodeURIComponent(window.location.href)}`" {
+                        "QrCode"
```
svenstaro, line 144: Minor nitpick but can you call this `QR code` instead? :)

src/renderer.rs
```diff
@@ -133,24 +134,33 @@ fn color_scheme_selector(
...
-                            @if active_color_scheme == color_scheme {
-                                li.active {
-                                    (color_scheme_link(sort_method, sort_order, color_scheme, default_color_scheme, serve_path))
-                                }
-                            } @else {
-                                li {
-                                    (color_scheme_link(sort_method, sort_order, color_scheme, default_color_scheme, serve_path))
-                                }
+                    div.qrcode {
+                        img#qrcode alt="QrCode" title="QR code of this page";
```
svenstaro, line 147: Also in this alt name as above.

PR #339, opened 2020-07-05
Change UTC time to local time

wyhaya:


svenstaro:
Would be cool if you made an option for this. I don't want to confuse users if possible. Should this also affect the times shown in the HTML?

wyhaya:
> Should this also affect the times shown in the HTML?

No, he is just the time in the log

---

I submitted this pr because I saw the wrong time, he did not indicate that it belongs to UTC time.

```
↪ cargo run -- .
...
01:55:31 [ERROR] Route /test could not be found
```
```
↪ date
Mon Jul  6 09:55:32 CST 2020
```

PR #341, opened 2020-07-07
Print the QR code in the terminal

wyhaya:
Complete #45

---

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,11 +3,12 @@
 use actix_web::http::{Method, StatusCode};
 use actix_web::{fs, middleware, server, App, HttpRequest, HttpResponse};
 use std::io::{self, Write};
-use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::thread;
 use std::time::Duration;
 use structopt::clap::crate_version;
 use yansi::{Color, Paint};
+use qrcodegen::{QrCode, QrCodeEcc};
 
 mod archive;
 mod args;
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -355,3 +383,45 @@ fn error_404(req: &HttpRequest<crate::MiniserveConfig>) -> Result<HttpResponse,
         .into_string(),
     ))
 }
+
+// Print the QR code of the specified content in the terminal
+// For the width and height to be equal, two characters are used in each grid
+// Cannot control the line-height of the terminal, so it looks imperfect
+fn print_qrcode(content: String) {
```
svenstaro, line 390: How about using something like this instead? https://github.com/svenstaro/qrcode-rust/blob/master/examples/encode_unicode.rs

Much less code for us and super clean unicode output.

src/main.rs
```diff
@@ -355,3 +383,45 @@ fn error_404(req: &HttpRequest<crate::MiniserveConfig>) -> Result<HttpResponse,
...
+            }
+        }
+        text.push_str("\n");
+    }
+
+    print!("{}", text);
+}
+
+fn get_intranet_ip() -> Option<IpAddr> {
+    pnet_datalink::interfaces()
```
svenstaro, line 420: Why exactly did you pull in this dependency?

src/main.rs
wyhaya: Looks good, should we use `qrcode` create?

src/main.rs
wyhaya: Get local IP address.

src/main.rs
svenstaro: Yeah, then we should also change the current qrcodegen dependency to that.

src/main.rs
svenstaro: I'm concerned about this apparently requring winpcap on Windows which think we can't link statically. I'd very much like this to stay a single static binary on Windows. I just checked this with `cargo-cross` but you could also check this natively.

startewho: Myabe the crate [systemstat](https://github.com/myfreeweb/systemstat) will be an option. it support get the interfaces on diffferent os (contion windwos also).

Issue #147, opened 2019-06-25
actix-web 2.0 upgrade


svenstaro:
We should upgrade to actix-web 1.0 to be on a stable API for some time to come.

efx:
https://github.com/svenstaro/miniserve/pull/185

PR #343, opened 2020-07-17
Update to actix 2 and futures 0.3

equal-l2:
This supersedes #270, #307 and #326, and obsoletes #325.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -1,19 +1,26 @@
-use actix_web::dev::{Body, Service, ServiceRequest, ServiceResponse, Transform};
+use actix_web::dev::ServiceRequest;
 use actix_web::http::{header, StatusCode};
 use actix_web::{HttpRequest, HttpResponse, Result};
-use futures::future;
+use actix_web_httpauth::extractors::basic::BasicAuth;
 use sha2::{Digest, Sha256, Sha512};
 
 use crate::errors::{self, ContextualError};
 use crate::renderer;
-
-pub struct Auth;
 
 #[derive(Clone, Debug)]
 /// HTTP Basic authentication parameters
 pub struct BasicAuthParams {
     pub username: String,
     pub password: String,
+}
+
+impl From<BasicAuth> for BasicAuthParams {
+    fn from(auth: BasicAuth) -> Self {
+        Self {
+            username: auth.user_id().to_string(),
+            password: auth.password().unwrap_or(&"".into()).to_string(),
+        }
+    }
 }
 
 #[derive(Clone, Debug, PartialEq)]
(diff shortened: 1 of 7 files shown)


svenstaro:
Holy smokes, that's great! I've been procrastrinating on this update for so long and you just came along and did it. I'll be reviewing it shortly.

src/auth.rs
```diff
@@ -87,66 +88,120 @@ pub fn get_hash<T: Digest>(text: &str) -> Vec<u8> {
...
 }
 
-impl Middleware<crate::MiniserveConfig> for Auth {
-    fn response(
-        &self,
-        req: &HttpRequest<crate::MiniserveConfig>,
-        resp: HttpResponse,
-    ) -> Result<Response> {
-        let required_auth = &req.state().auth;
+pub struct AuthMiddleware<S> {
```
svenstaro, line 91: While I recognize that this is a direct port of the pre-existing middleware from actix-web 0.x, perhaps by now it might be a better idea to use the official [actix-web-httpauth](https://github.com/actix/actix-extras/tree/master/actix-web-httpauth)? It seems to be flexible enough as you can pass a custom authenticator to it. It might save us from some complicated-looking code. :)

Can you look at it and see whether it's a good fit?

equal-l2:
I confirmed that updating to actix-web 3 doesn't require code change, only requires version bump of dependensies.
Considering actix-web 3 is still in beta, I'd like not to update at this moment.

src/auth.rs
equal-l2: Thanks for the suggestion, I'll take a look at it.

src/pipe.rs
```diff
@@ -10,15 +11,15 @@ use std::io::{Error, ErrorKind, Result, Write};
 pub struct Pipe {
     // Wrapping the sender in `Wait` makes it blocking, so we can implement blocking-style
```
svenstaro, line 12: This comment is now out-of-date.

src/pipe.rs
```diff
@@ -28,7 +29,7 @@ impl Drop for Pipe {
     fn drop(&mut self) {
         // This is the correct thing to do, but is not super important since the `Sink`
         // implementation of `Sender` just returns `Ok` without doing anything else.
-        let _ = self.dest.close();
+        let _ = block_on(self.dest.close());
```
svenstaro, line 32: How about instead of the `block_on` here to make this compatible with this sync trait, you try to use [AsyncWrite](https://docs.rs/tokio/0.2.21/tokio/io/trait.AsyncWrite.html)? It's also re-exported as `actix_codec::AsyncWrite`. It might make this code more nicely integrated if all of this were async and then we wouldn't need the `block_on`. :)

svenstaro: Same applies to the other `block_on`s down below.

svenstaro:
> I confirmed that updating to actix-web 3 requires no code change, it only requires version bump of dependensies.
> Considering actix-web 3 is still in beta, I'd like not to update at this moment.

Ok, that's fair. Especially since I anticipate the update to be trivial once it comes.

src/auth.rs
equal-l2: Done in 8bf16f6.

svenstaro: Just looked at it. Super great change! Really cuts down on some very complicated looking code. Let's face it, actix-web middlewares just aren't pretty. :D

svenstaro:
Also fixes #147.

src/pipe.rs
equal-l2: `Pipe` exists because `create_archive` only accepts `Write`ables.
If it accepted `AsyncWrite`, `Pipe` was not required because `Sender` itself implements `AsyncWrite` (via `Sink` trait).

I'll investigate if I can rewrite `create_archive` and its family so that they can accept `AsyncWrite`.

src/pipe.rs
equal-l2: Done in d88aa89.

svenstaro:
If you rebase you get nice GitHub Actions-based CI! :)

src/pipe.rs
svenstaro: That'd be great! I think it would result in less code and hopefully simpler code, too, given that there'd be no conversion between sync/async code.

equal-l2: I found an async implementation of gzip: [async-compression](https://github.com/Nemo157/async-compression).
The bad news is that I couldn't find the one for tar and zip.
(There is [async-tar](https://github.com/dignifiedquire/async-tar), but it doesn't accept `AsyncWrite`)

If dependencies don't support `AsyncWrite`, we have to write a fair amount of glue code, which will be more complicated than using `Pipe`.

So, I'll pass on that this time.

svenstaro: Alright, that's fair. Thanks for looking into it. I was sure that zip couldn't do it (and we already state in the help that generating a large zip will block the server) but I'm saddened to hear that there's nothing for tar. Alright, let's carry on then and revisit this in some while.

equal-l2:
@svenstaro Could you stop and re-run macOS CI?
It stalls on `cargo test` section.

svenstaro:
Merging as is! Stellar work.

PR #362, opened 2020-08-21
visual and sorting fixes

ahti:
* removes percent-encoding in heading and title
* makes name ordering case insensitive 
* gives name column more space


svenstaro:
Tested and working perfectly, thanks!

Issue #372, opened 2020-09-12
Tiny typo "now ago"


KevCui:
There is a tiny typo "now ago" when creating new file and refresh server page immediately. In this case, it's better to show only "now" without "ago".

![1599904310](https://user-images.githubusercontent.com/14293537/92992967-ba1bcf80-f4ee-11ea-9c5e-76732826c9ad.jpg)

svenstaro:
Oh good catch. Wanna make a PR?

KevCui:
@svenstaro Sure thing, please review PR.

PR #373, opened 2020-09-12
Show "now" when last modification time less than 11 seconds

KevCui:
Fix Issue #372 Tiny typo "now ago"

According to chrono-humanize `rough` period of accuracy, from 0 to 10 seconds duration, the text "now" is used https://gitlab.com/imp/chrono-humanize-rs/-/blob/master/src/humantime.rs#L173 and then followed by "ago" because of `past` tense https://gitlab.com/imp/chrono-humanize-rs/-/blob/master/src/humantime.rs#L141, which cause the tiny typo "now ago".

After the changes in this PR, if last modification time is less than 11 seconds, only "now" will show up.

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -913,7 +913,13 @@ fn humanize_systemtime(src_time: Option<SystemTime>) -> Option<String> {
     src_time
         .and_then(|std_time| SystemTime::now().duration_since(std_time).ok())
         .and_then(|from_now| Duration::from_std(from_now).ok())
-        .map(|duration| HumanTime::from(duration).to_text_en(Accuracy::Rough, Tense::Past))
+        .map(|duration| {
+            if duration < Duration::seconds(11) {
+                format!("now")
+            } else {
+                HumanTime::from(duration).to_text_en(Accuracy::Rough, Tense::Past)
+            }
+        })
 }
 
 /// Renders an error on the webpage
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -913,7 +913,7 @@ fn humanize_systemtime(src_time: Option<SystemTime>) -> Option<String> {
     src_time
         .and_then(|std_time| SystemTime::now().duration_since(std_time).ok())
         .and_then(|from_now| Duration::from_std(from_now).ok())
-        .map(|duration| { format!("{}", HumanTime::from(-duration))})
+        .map(|duration| format!("{}", HumanTime::from(-duration)))
 }
 
 /// Renders an error on the webpage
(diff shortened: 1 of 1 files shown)


svenstaro:
Can you make clippy happy?

KevCui:
clippy should be happy now. :smile:

svenstaro:
Thanks! I was about to merge this and then I thought: Perhaps this is actually a better fix to put into humantime? Could you check into that real quick and see what upstream has to say about this? If they refuse for some reason, we can still merge it hear but I believe this is a general humantime problem.

KevCui:
You're right. I thought the same thing.

I'll try to quickly contact the author of `chrono-humanize-rs` on gitlab. However, judging by the date of its latest commit, I'm afraid that the reply/fix in the upstream would take some time. Meanwhile, I'd suggest to merge the fix here in miniserve. If I got the the fix done in upstream in the future, I'll come back here and revert this change in miniserve.

Anyway, you're the better one to decide whether to merge the PR or not. I'll leave the decision to you. :)

KevCui:
fyi, opened an issue in `chrono-humanize-rs` https://gitlab.com/imp/chrono-humanize-rs/-/issues/7

svenstaro:
Let's let this sit for a few days to give the author some time to reply. Please do nudge me if I still haven't merged this in a week.

Thanks for your efforts!

imp:
As mentioned in the [chrono-humanize issue](https://gitlab.com/imp/chrono-humanize-rs/-/issues/7) - the expected usage is to print `Humantime` value directly instead of going through `to_text_en` - this will do the right thing for all possible values of the `Duration`

KevCui:
Thanks @imp for the hint. I updated the commit to use `HumanTime` value directly instead of `HumanTime:to_text_en`. `HumanTime` handles duration as accuracy `rough` by default and treats negative duration value as `past` tense. As the result, it works well when duration is form 0 to -10 seconds, showing `now`.

imp:
Took me some time to realize that your input is `std::time::SystemTime`. Once I did I went ahead and added

`impl Humanize for SystemTime` (much as that of the other types in `chrono`)

So now you can directly call it on `SystemTime` like this:
```
use chrono_humanize::Humanize;

fn humanize_systemtime(time: Option<SystemTime>) -> Option<String> {
    time.map(|time| time.humanize())
}
```

* requires `chrono-humanize` >= 0.1.1

svenstaro:
Thanks a lot @imp! @KevCui I updated to `chrono-humanize` 0.1.1 so you can now use it in this PR.

KevCui:
Pretty handy, thanks for your support @imp :+1:

PR #380, opened 2020-09-25
client-side color-scheme handling

ahti:


build.rs
```diff
@@ -0,0 +1,18 @@
...
+    fs::write(
+        &dest_path,
+        grass::from_string(
+            include_str!("data/style.scss").to_string(),
+            &grass::Options::default(),
+        )
+        .unwrap(),
+    )
+    .unwrap();
+    println!("cargo:rerun-if-changed=data/style.css");
```
svenstaro, line 17: Shouldn't this be `     println!("cargo:rerun-if-changed=data/style.scss"); `?

After all, it's the scss that changes to begin with.

data/style.scss
```diff
@@ -0,0 +1,723 @@
...
+    --drag_background: #3333338f;
+    --drag_border_color: #f8f8f2;
+    --drag_text_color: #f8f8f2;
+    --size_background_color: #75715e;
+    --size_text_color: #f8f8f2;
+    --error_color: #d02929;
+}
+
+
+@each $theme in $themes {
```
svenstaro, line 697: Please add a comment what this loop is for.

data/style.scss
```diff
@@ -0,0 +1,723 @@
+@use "sass:selector";
+
+/* theme colors can be found at the bottom */
```
svenstaro, line 3: You can use `//` as we don't need this comment in the compiled CSS I reckon.

data/style.scss
```diff
@@ -0,0 +1,723 @@
...
+$themes: squirrel, archlinux, monokai, zenburn;
+
+@function body_not_themed() {
```
svenstaro, line 6: Could you document the idea behind this function and why it's needed?

build.rs
```diff
@@ -0,0 +1,18 @@
+use std::env;
+use std::fs;
+use std::path::Path;
+
+fn main() {
```
svenstaro, line 5: Wouldn't it be neater to use `?` here everywhere instead of unwrap and have `main()` just return an `anyhow::Result`?

src/renderer.rs
```diff
@@ -46,16 +38,37 @@ pub fn page(
     html! {
         (DOCTYPE)
         html {
-            (page_header(&title_path, color_scheme, file_upload, favicon_route))
-            body#drop-container {
+            (page_header(&title_path, file_upload, favicon_route, css_route))
+            body#drop-container.(format!("default_theme_{}", default_color_scheme)).(format!("default_theme_dark_{}", default_color_scheme_dark)) {
+                (PreEscaped(r#"
```
svenstaro, line 43: Maybe add a comment here what this snippet is for.

svenstaro (changes requested):

PR #382, opened 2020-09-30
Remove actix-rt

equal-l2:
`actix_rt::main` is now re-exported as `actix_web::main`.


svenstaro:
Oh cool! I didn't notice. Thanks!

Issue #319, opened 2020-06-07
Add --cors option to include Access-Control-Allow-Origin: *


CrazyRoka:
Can't access from other domain. Need header `Access-Control-Allow-Origin: *`

jeifour:
Maybe it would be nice to be able to set custom headers? Like by doing `--header "Access-Control-Allow-Origin: *"` so then you cannot only do CORS but all kinds of different headers (CSP etc.)

svenstaro:
Yeah that would be a pretty nifty feature. I'd accept a PR for that! :)

svenstaro:
I think we'd need to handle `OPTIONS` properly. I think it would be a good idea for a PR. Would you like to take a stab at it?

rmacklin:
I don't have experience with rust nor this codebase/its dependencies, but are you looking for something as simple as

```diff
diff --git a/src/main.rs b/src/main.rs
index 34e56a0..6d11a3c 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -417,6 +417,13 @@ fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
         app.service(dir_service());
     }
 
+    // Handle OPTIONS requests for CORS and method discovery
+    app.service(
+        web::resource("/{tail:.*}")
+            .guard(guard::Options())
+            .to(options_handler)
+    );
+
     if conf.webdav_enabled {
         let fs = RestrictedFs::new(&conf.path, conf.show_hidden, conf.no_symlinks);
 
@@ -458,6 +465,22 @@ async fn healthcheck() -> impl Responder {
     HttpResponse::Ok().body("OK")
 }
 
+async fn options_handler(config: web::Data<MiniserveConfig>) -> impl Responder {
+    let mut allow_methods = vec!["GET", "HEAD", "OPTIONS"];
+
+    if config.file_upload {
+        allow_methods.push("POST");
+    }
+
+    if config.webdav_enabled {
+        allow_methods.push("PROPFIND");
+    }
+
+    HttpResponse::Ok()
+        .insert_header(("Allow", allow_methods.join(", ")))
+        .finish()
+}
+
 #[derive(Deserialize, Debug)]
 enum ApiCommand {
     /// Request the size of a particular directory

```

where it always accepts `OPTIONS` requests? (That seems to be enough for handling the CORS request that previously failed.) Or did you have something more complex in mind?

PR #452, opened 2021-02-21
Allow set custom headers from CLI

deantvv:
Try to implement feature request in #319 

use [httparse](https://github.com/seanmonstar/httparse) to parse input header and set to actix-web::middleware::DefaultHeader

Please let me know if there is any things I could make this code better. Thanks.


src/args.rs
```diff
@@ -117,6 +117,10 @@ struct CliArgs {
     /// Shown instead of host in page title and heading
     #[structopt(short = "t", long = "title")]
     title: Option<String>,
+
+    /// Custom header from user
+    #[structopt(long = "header")]
+    header: Option<String>,
```
svenstaro, line 123: I think it would be cool if you'd move the header parsing here instead. This way, a user will receive much better error feedback in case of a bad format. In order to accomplish this, you can change the type to `Option<Header>` here and implement the parsing logic in a [parse](https://docs.rs/structopt/0.3.21/structopt/#custom-string-parsers) function.

svenstaro (changes requested):
Cool stuff! I think we can move the parsing logic to a better location but otherwise, this seems fine and succinct.

src/args.rs
deantvv: The logic of parsing header has been moved to args.rs as you suggested.

I reinvent the owned version of Header from [httparse](https://github.com/seanmonstar/httparse) or I need to add lifetime to all config struct. Not sure whether it is a good decision or not.

Thanks for reviewing the code.

Cargo.toml
```diff
@@ -45,6 +45,7 @@ actix-files = "0.5"
 actix-multipart = "0.3.0"
 actix-web-httpauth = "0.5.0"
 mime = "0.3"
+httparse = "1.3.5"
```
svenstaro, line 48: Oh, forgot to mention, but since this is a version 1 dep, please change it depend on `"1"`. Since actix-http depends on hyper which depends on its own version of httparse, giving the versioning more leniency allows us to use the same version in both cases which decreases binary size.

src/args.rs
```diff
@@ -170,6 +174,39 @@ fn parse_auth(src: &str) -> Result<auth::RequiredAuth, ContextualError> {
     })
 }
 
+/// A own header modified from [httparse](https://docs.rs/httparse/1.3.5/src/httparse/lib.rs.html#415-425)
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub struct Header {
```
svenstaro, line 179: Hm, do we really need this custom struct? It's not a problem if we really do but I'd like to understand the reason first.

src/args.rs
```diff
@@ -117,6 +117,10 @@ struct CliArgs {
     /// Shown instead of host in page title and heading
     #[structopt(short = "t", long = "title")]
     title: Option<String>,
+
+    /// Custom header from user
+    #[structopt(long = "header", parse(try_from_str = parse_header))]
+    header: Option<Header>,
```
svenstaro, line 123: Wouldn't it be cool to be able to set multiple headers here? So something like `Vec<Header>` might be nice I think. :)

svenstaro (changes requested):

src/args.rs
svenstaro: Oh, sorry didn't see this comment before asking why we need a custom `Header`. Let me think about this a little. Certainly adding a lifetime to our config struct would make it really annoying to use!

svenstaro:
Oh, unrelated to my other comments: Could you add a little test for this to demonstrate that headers are properly set?

Cargo.toml
deantvv: Ok. I will fix this.

src/args.rs
```diff
@@ -170,6 +174,39 @@ fn parse_auth(src: &str) -> Result<auth::RequiredAuth, ContextualError> {
...
+impl Header {
+    fn new(header: &httparse::Header) -> Self {
+        Header {
+            name: header.name.to_string(),
+            value: header.value.to_vec(),
+        }
+    }
+}
+
+fn parse_header(src: &str) -> Result<Header, httparse::Error> {
```
svenstaro, line 200: Is it perhaps possible to directly output a `Result<DefaultHeaders, httparse::Error>` here? It might make everything a bit more ergonomic.

svenstaro (changes requested):

src/args.rs
deantvv: Because DefaultHeaders is not Send, it won't let us move to another thread. Is it another way that I could fix this?

svenstaro: Can't we clone our way out of this? It's not like performance is very important during arg parsing. In the worst case, we could use the nuclear option: `Arc<RefCell<Header>>`.

deantvv:
> 
> 
> Is `HeaderMap` from https://docs.rs/http/0.2.3/http/header/index.html suitable perhaps? It's already used by actix. It seems to do a lot of the same.

`HeaderMap` internally use `ByteStr` which still is still not owned type and need lifetime for the config struct.

[src](https://docs.rs/http/0.2.3/src/http/header/name.rs.html#32-34)
```rust
#[derive(Debug, Clone, Eq, PartialEq, Hash)]
enum Repr<T> {
    Standard(StandardHeader),
    Custom(T),
}

// Used to hijack the Hash impl
#[derive(Debug, Clone, Eq, PartialEq)]
struct Custom(ByteStr);
```

svenstaro:
Ah, missed that. That's too bad then! You can see I really just want to get out of having that Header struct in here but it apparently can't be helped.

deantvv:
@svenstaro I managed to get rid of custom struct `Header` and `Headers` by using `Bytes::copy_from_slice()`

Could you please take another look at the PR please? Thanks.

tests/header.rs
```diff
@@ -0,0 +1,64 @@
...
+use assert_cmd::prelude::*;
+use assert_fs::fixture::TempDir;
+use fixtures::{port, tmpdir, Error};
+use rstest::rstest;
+use std::process::{Command, Stdio};
+use std::thread::sleep;
+use std::time::Duration;
+
+#[rstest]
+fn custom_header_set(tmpdir: TempDir, port: u16) -> Result<(), Error> {
```
svenstaro, line 12: I think you should be able to use rstest's parametrization feature to write these tests more succinctly by parametrizing over a list of headers.

Cargo.toml
```diff
@@ -45,6 +45,9 @@ actix-files = "0.5"
 actix-multipart = "0.3.0"
 actix-web-httpauth = "0.5.0"
 mime = "0.3"
+httparse = "1"
+http = "0.2.3"
+bytes = "1.0.1"
```
svenstaro, line 50: Please just set "1".

src/args.rs
```diff
@@ -170,6 +176,28 @@ fn parse_auth(src: &str) -> Result<auth::RequiredAuth, ContextualError> {
     })
 }
 
+/// Custom header parser (allow multiple headers input)
+pub fn parse_header(src: &str) -> Result<HeaderMap, httparse::Error> {
+    // Max customized header is limit to 16
```
svenstaro, line 181: Typo: "is limitted to"

svenstaro (changes requested):
We're almost there! :)

deantvv:
> 
> 
> We're almost there! :)

Fix the stuff mentioned above. Thanks for reviewing the code. 👍

svenstaro:
Tests appear to be timing out. Can you check?

deantvv:
`cargo test` pass on my machine (WSL: debian). From the logs of CI, it seems it pass all tests but still fail to exit test.
Could you suggest where I can look more into?

svenstaro (approved):
Excellent work!

Issue #336, opened 2020-07-05
Add footer with version and software information


svenstaro:
It think this information should be there by default with a toggle switch to be able to turn it off.

PR #456, opened 2021-02-28
Add footer with program version

levaitamas:
Implements #336. Looks super basic. Comments, edits, etc. are welcome. 

---

### Directory Listing:
image

---

### An Error Page:
image

---

### DIFF
--- a/data/style.scss
+++ b/data/style.scss
@@ -45,6 +45,7 @@ body {
 .footer {
     text-align: center;
     padding-top: 1.5rem;
+    font-size: 0.7em;
     color: var(--footer_color);
 }
 
--- a/data/style.scss
+++ b/data/style.scss
@@ -42,6 +42,12 @@ body {
     word-break: break-word;
 }
 
+.footer {
+    text-align: center;
+    padding-top: 1.5rem;
+    color: var(--footer_color);
+}
+
 a {
     text-decoration: none;
 }
@@ -564,6 +570,7 @@ th span.active span {
     --size_background_color: #323232;
     --size_text_color: #ffffff;
     --error_color: #d02424;
+    --footer_color: #797979;
 }
 
 @mixin theme_archlinux {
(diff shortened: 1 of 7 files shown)


src/renderer.rs
```diff
@@ -138,12 +139,35 @@ pub fn page(
...
+                    }
                 }
             }
         }
     }
 }
 
+// Get version info
+fn get_version_info() -> String {
+    const PROG_NAME: Option<&'static str> = option_env!("CARGO_PKG_NAME");
```
svenstaro, line 153: Small nit: You can use `crate_version!()` and `crate_name!()` from clap (which is re-exported from structopt which we already use). This should cut down on the code a little.

src/renderer.rs
```diff
@@ -138,12 +139,35 @@ pub fn page(
...
+        "{}/{}",
+        PROG_NAME.unwrap_or("miniserve"),
+        VERSION.unwrap_or("unknown")
+    )
+}
+
+// Partial: version footer
+fn version_footer() -> Markup {
+    html! {
+        p style="text-align:center;padding-top:1em" {
```
svenstaro, line 165: Let's make this a little less visible. I suggest making it smaller and giving it a somewhat succinct color (depending on theme).

The `at_color` in `style.scss` seems like a good candidate to build off of. I suggest make a new variable `footer_color` in every theme, use the respective values of `at_color` and just tone down the contrast a little.

svenstaro (changes requested):
Very nice stuff. Just small nits. :)

svenstaro:
Oh, and you can tell clippy to ignore the complexity for that function that's failing.

levaitamas:
Thanks for your comments, @svenstaro. I managed to address them. Could you please take another look?

data/style.scss
```diff
@@ -42,6 +42,12 @@ body {
     word-break: break-word;
 }
 
+.footer {
+    text-align: center;
+    padding-top: 1.5rem;
+    color: var(--footer_color);
```
svenstaro, line 48: How about `font-size: 0.7em;` here as well?

levaitamas: Sounds good. Done in 474fadd.

svenstaro:
Great stuff, thanks! :)

PR #477, opened 2021-03-24
Disable Content-Disposition header

### LINKED ISSUE #462: [feature request] add an option to serve symlinks with the filename of the link, not the pointed-to file
example of the "issue":

shell 1:
```
_ :: /tmp/miniserve-rename » mkdir in out
_ :: /tmp/miniserve-rename » ls
in  out
_ :: /tmp/miniserve-rename » cd in
_ :: /tmp/miniserve-rename/in » echo foo > foo.txt
_ :: /tmp/miniserve-rename/in » ln -s foo.txt bar.txt
_ :: /tmp/miniserve-rename/in » miniserve .
miniserve v0.11.0
Serving path /tmp/miniserve-rename/in at http://[::]:8080, http://127.0.0.1:8080

Quit by pressing CTRL-C
```

shell 2:

```
_ :: /tmp/miniserve-rename/out » ls
_ :: /tmp/miniserve-rename/out » curl -OJ 127.0.0.1:8080/bar.txt 
  0     0    0     0    0     0      0      0 --:--:-- --:  0     4    0     0    0     0      0      0 --:--:-- --:100     4  100     4    0     0    363      0 --:--:-- --:--:-- --:--:--   333
_ :: /tmp/miniserve-rename/out » ls
foo.txt
_ :: /tmp/miniserve-rename/out » cat foo.txt 
foo
```

note that the requested filepath was `/bar.txt`, but the downloaded file is named `foo.txt`. it seems that the symlink is followed and the name of the link is forgotten. it would be nice to have an option to name the file as the symlink's name.

I'm happy to look into adding this feature, and any pointers on where to start would be appreciated.

aliemjay:
Otherwise, symlink files would be served with the filename of the
pointed-to file.

This should fix #462.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -340,14 +340,14 @@ fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
             )
         } else {
             let u_r = upload_route.clone();
-            let files;
-            if show_hidden {
-                files = actix_files::Files::new(&full_route, path)
-                    .show_files_listing()
-                    .use_hidden_files();
+            let files = actix_files::Files::new(&full_route, path)
+                .show_files_listing()
+                .disable_content_disposition();
+            let files = if show_hidden {
+                files.use_hidden_files()
             } else {
-                files = actix_files::Files::new(&full_route, path).show_files_listing();
-            }
+                files
+            };
 
             let files = files
                 .files_listing_renderer(move |dir, req| {
--- a/tests/serve_request.rs
+++ b/tests/serve_request.rs
@@ -25,6 +25,11 @@ fn serves_requests_with_no_options(tmpdir: TempDir) -> Result<(), Error> {
     let parsed = Document::from_read(body)?;
     for &file in FILES {
         assert!(parsed.find(|x: &Node| x.text() == file).next().is_some());
+        let resp = reqwest::blocking::get(format!("http://localhost:8080/{}", file))?
+            .error_for_status()?;
+
+        // `content-disposition` header is bogus for symlinks. Disable for now!
+        assert!(resp.headers().get("content-disposition").is_none());
     }
 
     child.kill()?;


aliemjay:
The URL for a symlink file is currently treated like a directory: They have a trailing "/" and may contain some query parameters (e.g. sorting method). For example: `http://localhost:1212/sss.pdf/?sort=size&order=asc`. Because if this, when downloading with some download managers (e.g. arias2c or wget), the filename will always be "index.html" and may contain some parameters.

This should be fixed by #479

svenstaro:
Could you rebase this?

svenstaro:
I took a look at this and it does work. I'd just like to discuss: Do you think it would be better to set the proper `filename="symlink-name"` attribute in the `content-disposition` header or to just get rid of it entirely? We do lose the flexibility of being able to tell the browser what to try to display.

aliemjay:
I've considered this, but after looking at actix-files sources, there seems to be no feasible way to do so. actix-files always uses the canonical path of the file to set the content-dispostion header.

svenstaro:
Alright then. Feel like adding a check to an existing test or a new test for this? I'd merge without it if you don't want to add a test condition for such a small change.

svenstaro:
As an aside: You might wanna report a feature request over at actix-files for this particular case. It _would_ be nicer to be able to set the proper name, after all.

tests/serve_request.rs
```diff
@@ -25,6 +25,10 @@ fn serves_requests_with_no_options(tmpdir: TempDir) -> Result<(), Error> {
     let parsed = Document::from_read(body)?;
     for &file in FILES {
         assert!(parsed.find(|x: &Node| x.text() == file).next().is_some());
+
+        let resp = reqwest::blocking::get(format!("http://localhost:8080/{}", file))?
```
svenstaro, line 29: For specific checks like this, I like seeing a comment as to _why_ we're checking for this particular thing.

src/main.rs
```diff
@@ -344,9 +344,12 @@ fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
             if show_hidden {
                 files = actix_files::Files::new(&full_route, path)
                     .show_files_listing()
+                    .disable_content_disposition()
```
svenstaro, line 347: Now what this condition block is growing, how about trying to unify it a little? Maybe something like:
```rust
let files = actix_files::Files::new(&full_route, path).show_files_listing().disable_content_disposition();
if show_hidden {
    let files = files.use_hidden_files();
}
```
Would that work?

svenstaro (changes requested):
Just some tiny nits. :)

svenstaro:
Do you want to have a shot and try to solve this upstream? I think I'm ok integrating a mid-term fix until this is properly fixed upstream, though.

aliemjay:
I don't know if I would solve it upstream, but I've just tried to implement the requested changed.

svenstaro:
So with https://github.com/actix/actix-web/pull/2156 merged and therefore your issue https://github.com/actix/actix-web/issues/2128 fixed in actix itself, I believe it'd be proper to await the actix-web 4 release. What do you think?

aliemjay:
I agree. Closing...

PR #479, opened 2021-03-25
Resolve symlinks on directory listing

aliemjay:
This has the benefit of showing the size and modification date of the pointed-to file. Symlink to directories now respects '--dirs-first' option and broken symlinks don't show up in directory listing.

The only downside AFAIK is more syscalls when listing directories.

@svenstaro Any input on this? should it be configurable? can we abandon the symlink symbol?

### DIFF
--- a/data/style.scss
+++ b/data/style.scss
@@ -94,7 +94,8 @@ a.symlink:hover {
     color: var(--symlink_link_color);
 }
 
-.symlink-symbol {
+.symlink-symbol::after {
+    content: "⇢";
     display: inline-block;
     border: 1px solid var(--symlink_link_color);
     margin-left: 0.5rem;
--- a/data/style.scss
+++ b/data/style.scss
@@ -88,7 +88,7 @@ a.file:hover {
 .symlink-symbol::after {
     content: "⇢";
     display: inline-block;
-    border: 1px solid var(--symlink_link_color);
+    border: 1px solid;
     margin-left: 0.5rem;
     border-radius: 0.2rem;
     padding: 0 0.1rem;
@@ -521,7 +521,6 @@ th span.active span {
     --text_color: #323232;
     --directory_link_color: #d02474;
     --file_link_color: #0086b3;
-    --symlink_link_color: #ed6a43;
     --table_background: #ffffff;
     --table_text_color: #323232;
     --table_header_background: #323232;
@@ -566,7 +565,6 @@ th span.active span {
     --text_color: #fefefe;
     --directory_link_color: #03a9f4;
     --file_link_color: #ea95ff;
-    --symlink_link_color: #ff9800;
     --table_background: #353946;
     --table_text_color: #eeeeee;
     --table_header_background: #5294e2;
@@ -611,7 +609,6 @@ th span.active span {
     --text_color: #efefef;
     --directory_link_color: #f0dfaf;
     --file_link_color: #87d6d5;
-    --symlink_link_color: #ffccee;
     --table_background: #4a4949;
     --table_text_color: #efefef;
     --table_header_background: #7f9f7f;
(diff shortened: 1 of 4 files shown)


svenstaro:
Could you rebase this?

aliemjay:
@svenstaro done.

svenstaro:
I quite like this as it is. I like keeping the symlink symbol around. Of course, I'd appreciate a test for this if you want to make one but I'd say for this change it's optional. Once you're happy, remove the `[RFC]` prefix and I'll merge it. :)

aliemjay:
RFC tag removed!

CSS styles may still need some cleanup though: `symlink` class is no longer used, `symlink-symbol` border color seems off to me.

svenstaro:
Alright, feel free to remove the disused class, no reason to keep it around unless we maybe want to make use of it still? Perhaps we can underline the symlinks in the symlink color?

tests/fixtures/mod.rs
```diff
@@ -42,23 +48,42 @@ pub fn tmpdir() -> TempDir {
     let tmpdir = assert_fs::TempDir::new().expect("Couldn't create a temp dir for tests");
     let mut files = FILES.to_vec();
     files.extend_from_slice(HIDDEN_FILES);
-    for file in &files {
+    let (files_hard, files_symlink) = files.split_at(1);
+    println!("path = {:?}", tmpdir.path());
```
svenstaro, line 52: Is this a debug leftover?

tests/fixtures/mod.rs
```diff
@@ -42,23 +48,42 @@ pub fn tmpdir() -> TempDir {
...
+    for file in files_hard {
         tmpdir
             .child(file)
             .write_str("Test Hello Yes")
             .expect("Couldn't write to file");
     }
+    for file in files_symlink {
```
svenstaro, line 59: Could you sprinkle a few comments in here about what we're doing in all the loops and such now that this fixture has become a bit more complex?

svenstaro:
Cool stuff with the fixtures! Now perhaps just make one test check whether the symlinks are actually rendered as they should be (while other files and dirs are still rendered as _they_ should be.

tests/fixtures/mod.rs
svenstaro: By the way, you can use `dbg!()` instead as an easier way to achieve exactly the same print!

aliemjay:
> Cool stuff with the fixtures! Now perhaps just make one test check whether the symlinks are actually rendered as they should be (while other files and dirs are still rendered as _they_ should be.

I believe that the current tests would still fail with the old behaviour because symlink filenames do always have a trailing `/`.
But I think that a single standalone test is better and that symlink files and directories should be made local to such test. Fixing that soon..

aliemjay:
I don't really have a strong opinion on symlink symbol style, but it looks ok to me without a special style.

svenstaro:
@aliemjay could you resolve the conflict? I think this was in pretty good shape already the last time I looked at it.

svenstaro:
Looking at this again, I'm thinking: Wouldn't it be kind of cool to render the pointed-to relative location of the symlink behind the arrow?

aliemjay:
I have no strong opinion on this, but a symlink may be absolute or may point outside the serve_path  and that would show potentially unwanted path names.

svenstaro:
Perhaps that's something we could show actually in case it's pointing to somewhere that would be off-limits?

aliemjay:
Well, as I said,I have no strong opinion but I see that symlinks should be totally abstracted away from the user and should be used by the admin only to organize the file tree.

svenstaro:
That's a fair point! How about a follow-up PR with a new flag to show symlink destination? :)

aliemjay:
That would be interesting. I shall give it a try!

svenstaro:
Merging this one as is! Follow ups: https://github.com/svenstaro/miniserve/issues/498 and https://github.com/svenstaro/miniserve/issues/499.

PR #480, opened 2021-03-26
Fix QR code display when using --random-route

rouge8:
Previously the QR code would not display because `/` is a 404 when using `--random-route`.


svenstaro:
Oh noice, never noticed this bug as I don't ever use this combination of options.

PR #481, opened 2021-03-26
Fix CI on Rust nightly by bumping standback

rouge8:
Previously building would fail with:

```
error[E0432]: unresolved import `core::task::Wake`
Error:    --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/standback-0.2.16/src/lib.rs:520:13
    |
520 |     pub use core::task::Wake;
    |             ^^^^^^^^^^^^----
    |             |           |
    |             |           help: a similar name exists in the module (notice the capitalization): `wake`
    |             no `Wake` in `task`

error: aborting due to previous error
Error: aborting due to previous error
For more information about this error, try `rustc --explain E0432`.
error: could not compile `standback`
Error: could not compile `standback`
```

https://github.com/jhpratt/standback/releases/tag/v0.2.17

PR #482, opened 2021-03-26
Generate completions with `miniserve --print-completions <shell>`

### LINKED ISSUE #377: Request: Generate shell completion files
This application uses [structopt](https://github.com/TeXitoi/structopt), which uses [clap](https://clap.rs/), which has APIs to generate completion code for bash, fish, zsh, powershell, and elvish. You can use [`structopt::clap::App::gen_completions`](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.gen_completions) or [`structopt::clap::App::gen_completions_to`](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.gen_completions_to) to generate completion code from your `StructOpt::clap()`.

## How to provide completion code

There are two common ways of providing completion code:

### Embedded functionality within the main application

Create a subcommand (`miniserve completions <shell>`) or option (`miniserve --completions=<shell>`) that prints completion code to stdout and exits.

**Pros:**
  * Simple to implement.
  * Simple for repo owner to distribute: It is already within the main application.
  * Completions will always be up-to-date.

**Cons:**
  * Completion is almost always generated once, after that, the sub-functionality is considered bloat.
  * More complicated for Linux distros to distribute: It must execute miniserve to generate completions.

### Generate completions ahead of time

Use [`build.rs`](https://doc.rust-lang.org/cargo/reference/build-scripts.html) to generate completions whenever the main application is built.

**Pros:**
  * Completion is generated ahead of time, separated from the main application, user just has to download what is needed, no unnecessary bloat.
  * Simple for Linux distros to distribute: Just install the pre-generated completions.

**Cons:**
  * More complicated to implement.
  * Repo owner must distribute completion files either via GitHub Release or via the git repo itself.
  * If repo owner chooses to distribute completion via git repo, they must make sure completions always up-to-date.

rouge8:
This patch adds a ~`completions` subcommand~ `--print-completions` option to generate shell completion files at runtime. This ensures the completions are always up to date.

Fixes #377.

CI should pass once #481 is merged. This is some of my first Rust code so lmk if I'm doing anything weird...

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -21,7 +21,7 @@ const ROUTE_ALPHABET: [char; 16] = [
     about,
     global_settings = &[structopt::clap::AppSettings::ColoredHelp],
 )]
-struct CliArgs {
+pub struct CliArgs {
     /// Be verbose, includes emitting access logs
     #[structopt(short = "v", long = "verbose")]
     verbose: bool,
@@ -131,6 +131,10 @@ struct CliArgs {
     /// Hide version footer
     #[structopt(short = "F", long = "hide-version-footer")]
     hide_version_footer: bool,
+
+    /// Generate completion file for a shell
+    #[structopt(long = "print-completions", value_name = "shell")]
+    pub print_completions: Option<structopt::clap::Shell>,
 }
 
 /// Checks wether an interface is valid, i.e. it can be parsed into an IP address
@@ -205,9 +209,7 @@ pub fn parse_header(src: &str) -> Result<HeaderMap, httparse::Error> {
 }
 
 /// Parses the command line arguments
-pub fn parse_args() -> crate::MiniserveConfig {
-    let args = CliArgs::from_args();
-
+pub fn parse_args(args: CliArgs) -> crate::MiniserveConfig {
     let interfaces = if !args.interfaces.is_empty() {
         args.interfaces
     } else {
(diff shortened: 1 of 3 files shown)


svenstaro (changes requested):
Hey, thanks a ton! I've been meaning to do this but you just went ahead and did it and that's great! My thoughts are: Shall we make this an option rather than a subcommand? We don't use any subcommands in miniserve so far and I don't really want to change the user flow for this particular thing. I imagine something like `--print-completions <shell>` and then also restrict it to the available shells for better user feedback.

rouge8:
> Shall we make this an option rather than a subcommand? We don't use any subcommands in miniserve so far and I don't really want to change the user flow for this particular thing. I imagine something like --print-completions <shell> 

Agreed, done! I got the subcommand idea from #377, but I agree another option makes more sense here.

> also restrict it to the available shells for better user feedback.

Yup, that part is done by using the `structopt::clap::Shell` enum as the option type: https://github.com/svenstaro/miniserve/pull/482/files#diff-c3fd08257cd326dfa1c2a3c0a69e6b842f194950ac32e7b2c6e2750f52d1765fR137. I added a test with an invalid shell option to ensure it fails.

svenstaro (approved):
Excellent, thanks!

PR #484, opened 2021-03-27
List directory if the index file doesn't exist

### LINKED ISSUE #275: 'miniserve --index index.html' gives 404 when index.html is not present in a directory.
## Steps to reproduce

```sh
mkdir -p public/sub
echo 'Hello There' > public/sub/index.html
cd public # public does not contain an index.html
miniserve -i 8080 --index index.html .
```

Open the browser, go to http://localhost:8080/ and http://localhost:8080/sub/.

## Expected behavior

* http://localhost:8080/ should display a directory list.
* http://localhost:8080/sub/ should display content of `public/sub/index.html`.

## Actual behavior

* http://localhosy:8080/ is not found (404).

aliemjay:
actix-files doesn't support such behaviour, which makes it necessary to
implement in the "listing renderer".

Should fix #275

### DIFF
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -169,6 +169,7 @@ pub fn directory_listing(
     dirs_first: bool,
     hide_version_footer: bool,
     title: Option<String>,
+    index: Option<PathBuf>,
 ) -> Result<ServiceResponse, io::Error> {
     use actix_web::dev::BodyEncoding;
     let serve_path = req.path();
@@ -251,6 +252,26 @@ pub fn directory_listing(
             }
         };
         return Ok(ServiceResponse::new(req.clone(), res));
+    }
+
+    // Serve index file. If it doesn't exist, ignore and list content.
+    if let Some(index_path) = index {
+        if let Ok(index_file) = actix_files::NamedFile::open(dir.path.join(index_path.clone())) {
+            let res = match index_file.into_response(req) {
+                Ok(res) => res,
+                Err(err) => {
+                    log::error!("Error while serving index file: {:?}", err);
+                    HttpResponse::InternalServerError().body(Body::Empty)
+                }
+            };
+            return Ok(ServiceResponse::new(req.clone(), res));
+        } else {
+            log::warn!(
+                "The file '{}' provided for option --index could not be found in {}.",
+                index_path.to_string_lossy(),
+                serve_path,
+            );
+        }
     }
 
     let mut entries: Vec<Entry> = Vec::new();
--- a/src/main.rs
+++ b/src/main.rs
@@ -420,6 +420,7 @@ fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
         let dirs_first = conf.dirs_first;
         let hide_version_footer = conf.hide_version_footer;
         let title = conf.title.clone();
+        let index = conf.index.clone();
         upload_route = if let Some(random_route) = conf.random_route.clone() {
             format!("/{}/upload", random_route)
         } else {
@@ -427,10 +428,6 @@ fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
         };
         if path.is_file() {
             None
-        } else if let Some(index_file) = &conf.index {
-            Some(
-                actix_files::Files::new(&full_route, path).index_file(index_file.to_string_lossy()),
-            )
         } else {
             let u_r = upload_route.clone();
             let files;
@@ -463,6 +460,7 @@ fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
                         dirs_first,
                         hide_version_footer,
                         title.clone(),
+                        index.clone(),
                     )
                 })
                 .prefer_utf8(true)
--- a/src/main.rs
+++ b/src/main.rs
@@ -114,7 +114,7 @@ pub struct MiniserveConfig {
 
 impl MiniserveConfig {
     /// Parses the command line arguments
-    fn from_args(args: args::CliArgs) -> Self {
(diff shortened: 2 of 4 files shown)


svenstaro:
Good stuff! Please do look into that test failure though and see whether perhaps you actually need to change or make new tests to make sure the new behavior is also being covered.

src/listing.rs
```diff
@@ -246,6 +247,20 @@ pub fn directory_listing(
         return Ok(ServiceResponse::new(req.clone(), res));
     }
 
+    // Serve index file. If it doesn't exist, ignore and list content.
+    if let Some(index_path) = index {
```
svenstaro, line 251: I think we might want to log an error/warning in this case though. If an index was explicitly set but didn't exist, the user might be confused as to why miniserve is still just providing a file list. A log entry will make the user aware that maybe a path is wrong.

svenstaro (changes requested):

src/listing.rs
aliemjay: I had the impression that this is the default and the expected behaviour, may be because of python's `http.server`. But I think a waning is useful!

src/main.rs
```diff
@@ -250,15 +250,6 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
         ContextualError::IoError("Failed to resolve path to be served".to_string(), e)
     })?;
 
-    if let Some(index_path) = &miniserve_config.index {
-        let has_index: std::path::PathBuf = [&canon_path, index_path].iter().collect();
-        if !has_index.exists() {
-            error!(
-                "The file '{}' provided for option --index could not be found.",
```
svenstaro, line 257: This will remove the warning upon startup and will only show it upon access to the directory. I think it's still useful to show the missing index directly upon startup.

src/listing.rs
```diff
@@ -246,6 +247,22 @@ pub fn directory_listing(
...
+            let res = match index_file.into_response(req) {
+                Ok(res) => res,
+                Err(err) => {
+                    log::error!("Error while serving index file: {:?}", err);
+                    HttpResponse::InternalServerError().body(Body::Empty)
+                }
+            };
+            return Ok(ServiceResponse::new(req.clone(), res));
+        } else {
+            log::warn!("Index file could not be found for {}", serve_path);
```
svenstaro, line 262: Can you use the same phrase I use in my startup warning?

svenstaro (changes requested):

svenstaro:
This makes me wonder: Should we warn in case the file provided by `--index` lays outside of the current serving root?

For instance, it's possible to serve /a while providing `--index /b/index.html` which might be unexpected and potentially a security problem if users assume that `--index /b/index.html` also sets the serving root to `/b/`.

In fact, now that I think about it, perhaps we should default to serving the directory the index file is in if no explicit serving path is provided. What do you think?

aliemjay:
I don't see the point of allowing the user to set a full path for `--index` option!
I think that the `--index` option should only accept a single file name and the user should always be required to set the root directory explicitly. My reasoning is that:

* the command `miniserve --index /b/index.html` may be interpreted as serving a single file, yet it would expose the whole tree under `/b/`
* It is not always true that the root directory has an index file. In my setup, only some subdirectories have an index file and I expect to see a directory listing for root directory. This was my motivation for this PR.

I believe index option may need a check of something like:
`if let Some((0, Component::Normal(_))) = index.components().enumerate().last() {}`

svenstaro:
@aliemjay I'm very much interested in merging this. See also [this effort in actix-web to offer something similar upstream](https://github.com/actix/actix-web/pull/2135). However, I'd like to get the path questions that I asked cleaned up and tested. As you said, it probably doesn't make much sense to allow the user to set a path outside the index file path or at worst it's a bad surprise for the user!

How about we change the behavior of the program like this:

1) Don't let `--index` take an argument.
2) If `--index` is provided, the path for miniserve needs to point to something that looks like a file.
3) If index doesn't exist, serve base dir of serve path.

Does that make sense to you? This way, we'd establish more intuitive behavior for `--index`.

aliemjay:
[The PR mentioned](https://github.com/actix/actix-web/pull/2135) conflicts with the behaviour implemented here!
I think then we need to offer an option of "What to to do when the index file doesn't exist" and this makes it neccessary to not rely on the upstream implementation of default page in order to allow the user choose.

svenstaro:
I agree, we don't necessarily follow upstream. Let's roll with your adapted implementation instead. How do you want to resolve the paths problems laid about above?

aliemjay:
> Does that make sense to you?

I've just gave it a thought and had some problems:
* First, This behaviour conflicts with serving a single file, which is already supported.

* Second, It doesn't allow such use case:

> It is not always true that the root directory has an index file. In my setup, only some subdirectories have an index file and I expect to see a directory listing for root directory. This was my motivation for this PR.

* Second, I don't know if  `miniserve --index ./xxx.html`, for example,  is expected to find and serve `xxx.html` at root ony (`/`) or at any route?

svenstaro:
Hmm I think in that case it would indeed be the best next thing to just serve each directory's respective index file if that exists and fall back to listing the files otherwise.

aliemjay:
> The compromise would be that `--index` can never list a full path but only ever a single file without preceding directories. How does that sound?

If you mean the current behaviour, then I think it's totally fine and intuitive :)

In this case, --index option needs a check to accept a single file name.

aliemjay:
Rebased and tried to fix the path issue with the following behaviour:

> In fact, now that I think about it, perhaps we should default to serving the directory the index file is in if no explicit serving path is provided. What do you think?

Plus, if the serving path is explicitely provided, then `--index` should only accept a single file name

svenstaro:
Superceded by #583.

PR #485, opened 2021-04-06
Fix URL percent encoding for special characters

### LINKED ISSUE #229: Some files cant get downloaded due to filename
I currently have some files which I can't download (miniserve responds with a 404)
Filenames are:
```
OBSOLESZNZ - TestSession #01.cue
OBSOLESZNZ - TestSession #01.mp3
OBSOLESZNZ - TestSession #01.mp3
```

Probably because of the whitespaces and `#`

aliemjay:
Currently some special chars (e.g. '%#') are passed as is in the file URLs. These chars are not allowed in the path portion of the URL and should be percent encoded.

Fixes #229

### DIFF
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -15,7 +15,7 @@ use strum_macros::{Display, EnumString};
 use crate::archive::CompressionMethod;
 use crate::errors::{self, ContextualError};
 use crate::renderer;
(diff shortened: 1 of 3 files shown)


svenstaro:
That's great, thanks! Nice and proper standards-compliant implementation, too! Merging as is. Please continue to contribute your good work. :)

PR #489, opened 2021-04-09
Fix breadcrumbs

aliemjay:
# 1. For RTL languages
* Before
image
* After
image

# 2. Fix mess mobile phones
* ~~This was introduced in #158.~~ See commit message!
*

### DIFF
--- a/data/style.scss
+++ b/data/style.scss
@@ -42,6 +42,11 @@ body {
     word-break: break-word;
 }
 
+.title a {
+    font-weight: bold;
+    color: var(--directory_link_color);
+}
+
 .footer {
     text-align: center;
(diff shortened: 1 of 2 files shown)


aliemjay:
I also suggest to allow wrapping at any character. This would make an efficient use of space in mobile view. This, for example, seems to take too much space:
image

svenstaro:
> I also suggest to allow wrapping at any character. This would make an efficient use of space in mobile view. This, for example, seems to take too much space:
> image

I like that suggestion. Could you make a separate PR with a screenshot comparison for ti?

svenstaro:
Thanks!

Issue #451, opened 2021-02-21
.tar.gz and .tar


JqrSo:
![`(LJE7 8%VR6UWZAF4%IVML](https://user-images.githubusercontent.com/52708641/108618750-8fe61a80-745b-11eb-9590-7ab0dd54ed81.png)

Can you turn on only one of these two buttons
![tar](https://user-images.githubusercontent.com/52708641/108618965-fe77a800-745c-11eb-8212-a35f87789220.jpg)
![tar gz](https://user-images.githubusercontent.com/52708641/108618969-02a3c580-745d-11eb-88af-65cde0a46c2f.jpg)

FLAGS:
           -r, --enable-tar
                      Enable tar archive generation

svenstaro:
An additional flag would make sense, I guess.

JqrSo:
> An additional flag would make sense, I guess.
我写了个bat
![QQ录屏20210228134854](https://user-images.githubusercontent.com/52708641/109409378-23c06500-79cd-11eb-9ca9-37f925ea737a.gif)
`@echo off
setlocal enabledelayedexpansion

for /f "tokens=2 delims=:" %%j in ('ipconfig ^| find /i "IPv4"') do (
set url=%%j
set ip=-i !url: =! !ip!
start http://!url: =!:8080
)

D:\app\miniserve-v0.11.0-x86_64-pc-windows-msvc.exe -D -r -q -u !ip! %1`

JqrSo:
> > An additional flag would make sense, I guess.
> 
> I wrote a bat.
> To achieve this effect
> 
> ![QQ录屏20210228134854](https://user-images.githubusercontent.com/52708641/109409378-23c06500-79cd-11eb-9ca9-37f925ea737a.gif)
> 
> code
> 
> @echo off
> setlocal enabledelayedexpansion
> for /f "tokens=2 delims=:" %%j in ('ipconfig ^| find /i "IPv4"') do (
> set url=%%j
> set ip=-i !url: =! !ip!
> start http://!url: =!:8080
> )
> D:\app\miniserve-v0.11.0-x86_64-pc-windows-msvc.exe -D -r -q -u !ip! %1

svenstaro:
This has been implemented by @deantvv!

PR #492, opened 2021-04-11
Separate tar archive and tar flags

deantvv:
It used to have one flag (-r) to enable both tar archive and tar.
Now it has two flags [ -r: for tar, -g: for tar archive].

Related to #451

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -53,15 +53,10 @@ impl CompressionMethod {
         }
     }
 
-    pub fn is_enabled(
-        self,
-        tar_enabled: bool,
-        tar_archive_enabled: bool,
-        zip_enabled: bool,
-    ) -> bool {
+    pub fn is_enabled(self, tar_enabled: bool, tar_gz_enabled: bool, zip_enabled: bool) -> bool {
         match self {
+            CompressionMethod::TarGz => tar_gz_enabled,
             CompressionMethod::Tar => tar_enabled,
-            CompressionMethod::TarGz => tar_archive_enabled,
             CompressionMethod::Zip => zip_enabled,
         }
     }
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -53,10 +53,15 @@ impl CompressionMethod {
         }
     }
 
-    pub fn is_enabled(self, tar_enabled: bool, zip_enabled: bool) -> bool {
+    pub fn is_enabled(
+        self,
+        tar_enabled: bool,
+        tar_archive_enabled: bool,
+        zip_enabled: bool,
+    ) -> bool {
         match self {
-            CompressionMethod::TarGz => tar_enabled,
             CompressionMethod::Tar => tar_enabled,
+            CompressionMethod::TarGz => tar_archive_enabled,
             CompressionMethod::Zip => zip_enabled,
         }
     }
--- a/src/args.rs
+++ b/src/args.rs
@@ -99,9 +99,13 @@ pub struct CliArgs {
     #[structopt(short = "o", long = "overwrite-files")]
     pub overwrite_files: bool,
 
-    /// Enable tar archive generation
+    /// Enable tar generation
     #[structopt(short = "r", long = "enable-tar")]
     pub enable_tar: bool,
+
+    /// Enable tar archive generation
+    #[structopt(short = "g", long = "enable-tar-archive")]
+    pub enable_tar_archive: bool,
 
     /// Enable zip archive generation
     ///
--- a/src/args.rs
+++ b/src/args.rs
@@ -99,13 +99,13 @@ pub struct CliArgs {
     #[structopt(short = "o", long = "overwrite-files")]
     pub overwrite_files: bool,
 
-    /// Enable tar generation
+    /// Enable uncompressed tar archive generation
     #[structopt(short = "r", long = "enable-tar")]
     pub enable_tar: bool,
 
-    /// Enable tar archive generation
-    #[structopt(short = "g", long = "enable-tar-archive")]
-    pub enable_tar_archive: bool,
+    /// Enable gz-compressed tar archive generation
+    #[structopt(short = "g", long = "enable-tar-gz")]
+    pub enable_tar_gz: bool,
 
     /// Enable zip archive generation
     ///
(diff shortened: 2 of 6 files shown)


deantvv:
@svenstaro Could you take a look at this? Thanks in advance.

svenstaro:
Will do, sorry, pretty busy these days.

svenstaro (changes requested):
I really like this and would basically accept it as is code-wise but I'd really like to distinguish the terminology here: Let's call uncompressed tarballs `uncompressed tar archives` and compressed ones `gz-compressed tar archives` for the sake of being able to add on more compression methods later on without it being to confusing sounding.

Keep in mind: tar are _always_ archives, whether they are compressed or not.

src/archive.rs
```diff
@@ -53,10 +53,15 @@ impl CompressionMethod {
         }
     }
 
-    pub fn is_enabled(self, tar_enabled: bool, zip_enabled: bool) -> bool {
+    pub fn is_enabled(
+        self,
+        tar_enabled: bool,
```
svenstaro, line 58: Let's go with `tar_enabled` and `tar_compressed_enabled`.

src/args.rs
```diff
@@ -99,10 +99,14 @@ pub struct CliArgs {
     #[structopt(short = "o", long = "overwrite-files")]
     pub overwrite_files: bool,
 
-    /// Enable tar archive generation
+    /// Enable tar generation
```
svenstaro, line 102: `/// Enable uncompressed tar archive generation`

src/args.rs
```diff
@@ -99,10 +99,14 @@ pub struct CliArgs {
...
     #[structopt(short = "r", long = "enable-tar")]
     pub enable_tar: bool,
 
+    /// Enable tar archive generation
```
svenstaro, line 106: `/// Enable gz-compressed tar archive generation`

svenstaro (changes requested):
As per my other longer comment, I gave you some in-code hints for what I mean. Could you change that code and the rest of your changes to conform to those suggestions?

tests/archive.rs
```diff
@@ -51,3 +51,44 @@ fn archives_are_disabled(tmpdir: TempDir, port: u16) -> Result<(), Error> {
 
     Ok(())
 }
+
+#[rstest]
```
svenstaro, line 55: Good stuff, I like tests! :D

src/archive.rs
deantvv: Yes, it is now changed.

tests/archive.rs
deantvv: Me too!

src/archive.rs
```diff
@@ -53,15 +53,10 @@ impl CompressionMethod {
...
 
-    pub fn is_enabled(
-        self,
-        tar_enabled: bool,
-        tar_archive_enabled: bool,
-        zip_enabled: bool,
-    ) -> bool {
+    pub fn is_enabled(self, tar_enabled: bool, tar_gz_enabled: bool, zip_enabled: bool) -> bool {
         match self {
+            CompressionMethod::TarGz => tar_gz_enabled,
```
svenstaro, line 58: Come to think of it, it probably wasn't proper of me to go with `CompressionMethod` here as it should really have been `ArchiveMethod` from the beginning. Oh well, might change this after merging your PR. :)

deantvv:
> I really like this and would basically accept it as is code-wise but I'd really like to distinguish the terminology here: Let's call uncompressed tarballs `uncompressed tar archives` and compressed ones `gz-compressed tar archives` for the sake of being able to add on more compression methods later on without it being to confusing sounding.
> 
> Keep in mind: tar are _always_ archives, whether they are compressed or not.

Thanks for the suggestions and fixing the wrong terminology. It is now fixed in c9506c7.

svenstaro:
Cool stuff! Will merge right once CI passes.

tests/archive.rs
```diff
@@ -51,3 +51,44 @@ fn archives_are_disabled(tmpdir: TempDir, port: u16) -> Result<(), Error> {
...
+        .arg(tmpdir.path())
+        .arg("-p")
+        .arg(port.to_string())
+        .arg("-g")
+        .stdout(Stdio::null())
+        .spawn()?;
+
+    sleep(Duration::from_secs(1));
+
+    // Ensure the links to the tar archive exists and tar not exists
```
svenstaro, line 67: One small thing I just found:
`// Ensure the links to the compressed tar archive exists and link to plain tar doesn't exist`

tests/archive.rs
```diff
@@ -51,3 +51,44 @@ fn archives_are_disabled(tmpdir: TempDir, port: u16) -> Result<(), Error> {
...
+    let body = reqwest::blocking::get(format!("http://localhost:{}", port).as_str())?
+        .error_for_status()?;
+    let parsed = Document::from_read(body)?;
+    assert!(parsed.find(Text).any(|x| x.text() == "Download .tar.gz"));
+    assert!(parsed.find(Text).all(|x| x.text() != "Download .tar"));
+
+    // Try to download, only tar_gz should works
```
svenstaro, line 74: `    // Try to download, only tar.gz should work`

tests/archive.rs
svenstaro, line 55: If you wanna do something cool here, you could use rstest for test parameterization to test `-p` and `-g` with very little test code overhead. Want to give that a try?

svenstaro: Actually, on second thought, I want to more properly clean up these tests anyhow. I'll just do that after merging.

tests/archive.rs
deantvv: I am happy to fix this. But since you already merge this, should I still change this?

tests/archive.rs
deantvv: I can try in maybe another PR?

tests/archive.rs
deantvv: Same as above.

tests/archive.rs
svenstaro: Sure, if you like, that'd be much appreciated. So what I'm thinking here: Have a single parametrized test in `tests/archive.rs` that would test for all possible combinations of compression flags whether the path is reachable whether the links are being rendered. Thanks to rstest, we'd end up with a bunch of separately generated distinct tests. I think that'd be rather clean and make quite a bit of sense. Wanna give it a shot?

Take a look at: https://docs.rs/rstest/0.7.0/rstest/attr.rstest.html#test-parametrized-cases

PR #500, opened 2021-04-18
Interfaces

### LINKED ISSUE #405: Can not visist when use default interface option
I use a win10 pc, when I  used the default option

> miniserve

it shows
> Serving path \\?\D:\Tool\bin at http://[::]:8080, http://127.0.0.1:8080

my inner ip is 192.168.3.103 when I visit the web it can not show anything. When I use the set the interface 

> miniserve -i 192.168.3.103

it works.it seems the [::] can not work in my windows

### LINKED ISSUE #422: [feature suggestion] show the IP address from all interfaces
miniserve could print the IP address from all interfaces it's listening on:

*http-server*
```
lvmc@Luizs-MacBook-Pro ~ % http-server
Starting up http-server, serving ./public
Available on:
  http://127.0.0.1:8081
  http://192.168.0.101:8081
  http://192.168.15.7:8081
Hit CTRL-C to stop the server
```

*miniserve*
```
lvmc@Luizs-MacBook-Pro Desktop % miniserve -q -v .
miniserve v0.10.3
00:32:55 [INFO] Starting 16 workers
00:32:55 [INFO] Starting "actix-web-service-[::]:8080" service on [::]:8080
00:32:55 [INFO] Starting "actix-web-service-127.0.0.1:8080" service on 127.0.0.1:8080
Serving path /Users/lvmc/Desktop at http://[::]:8080, http://127.0.0.1:8080

Quit by pressing CTRL-C
```

It helps a lot to share the links on local networks.

### LINKED ISSUE #470: [feature request]add 0.0.0.0 to bind to all interfaces
Add support for using 0.0.0.0 to bind to all interfaces.

If my pc have 2 ip address, 192.168.1.111 and 192.168.2.222, i appreciate the program find all the ip address automatically and bind to them when i using the following command.

miniserve -i 0.0.0.0 /tmp/myshare

aliemjay:
## 1. Report if any address fails to bind
`::actix_web::HttpServer::bind()` fails if and only if ALL the requested addresses fail to bind! an the docs require multiple calls for multiple addresses.

## 2. Fix: faulty bahvior when setting `-i 0.0.0.0`
Currrently, `0.0.0.0` address is converted into `127.0.0.1`!

Closes #470

## 3. Fix: On Windows, miniserve binds only to localhost(127.0.0.1) by default
This is in part because of the issue mentioned above, and also because binding to `::` binds only to IPv6 by default on Windows!

Closes #405 
Also reported by @KaneGreen in #470.

## 4. cross-platform default binding addresses
Currently, miniserve by default binds to both `0.0.0.0` and `::` ant the same time. This behaviour has a problem:
> On some platforms (e.g. majority of linux distros and OSX?) ,  the `::` address also accepts ipv4 connections. This makes these two addresses incompatible!
See https://stackoverflow.com/questions/1618240/how-to-support-both-ipv4-and-ipv6-connections

~I currently have no idea how this should be fixed!~

## 5. implement #422
Closes #422

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -52,10 +52,11 @@ fn main() -> Result<()> {
 
     let miniserve_config = MiniserveConfig::try_from_args(args)?;
 
-    match run(miniserve_config) {
-        Ok(()) => (),
-        Err(e) => errors::log_error_chain(e.to_string()),
-    }
+    run(miniserve_config).map_err(|e| {
+        errors::log_error_chain(e.to_string());
+        e
+    })?;
+
     Ok(())
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,6 @@
 use std::io;
 use std::io::Write;
-use std::net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener};
+use std::net::{IpAddr, SocketAddr, TcpListener};
 use std::thread;
 use std::time::Duration;
 
@@ -102,23 +102,6 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
 
     let inside_config = miniserve_config.clone();
 
-    let interfaces = miniserve_config
-        .interfaces
-        .iter()
-        .map(|&interface| {
-            if interface == IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)) {
-                // If the interface is 0.0.0.0, we'll change it to 127.0.0.1 so that clicking the link will
-                // also work on Windows. Why can't Windows interpret 0.0.0.0?
-                "127.0.0.1".to_string()
-            } else if interface.is_ipv6() {
-                // If the interface is IPv6 then we'll print it with brackets so that it is clickable.
-                format!("[{}]", interface)
-            } else {
-                format!("{}", interface)
-            }
-        })
-        .collect::<Vec<String>>();
-
     let canon_path = miniserve_config.path.canonicalize().map_err(|e| {
         ContextualError::IoError("Failed to resolve path to be served".to_string(), e)
     })?;
@@ -164,37 +147,49 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
             thread::sleep(Duration::from_millis(500));
         }
     }
-    let mut addresses = String::new();
-    for interface in &interfaces {
-        if !addresses.is_empty() {
-            addresses.push_str(", ");
-        }
-        let protocol = if miniserve_config.tls_rustls_config.is_some() {
-            "https"
-        } else {
-            "http"
-        };
-        addresses.push_str(&format!(
-            "{}",
-            Color::Green
-                .paint(format!(
-                    "{protocol}://{interface}:{port}",
-                    protocol = protocol,
-                    interface = &interface,
-                    port = miniserve_config.port
-                ))
-                .bold()
-        ));
 
-        if let Some(random_route) = miniserve_config.clone().random_route {
-            addresses.push_str(&format!(
-                "{}",
-                Color::Green
-                    .paint(format!("/{random_route}", random_route = random_route,))
-                    .bold()
-            ));
+    let addresses = {
+        let (mut ifaces, wildcard): (Vec<_>, Vec<_>) = miniserve_config
+            .interfaces
+            .clone()
+            .into_iter()
+            .partition(|addr| !addr.is_unspecified());
+
+        // Replace wildcard addresses with local interface addresses
+        if !wildcard.is_empty() {
+            let all_ipv4 = wildcard.iter().any(|addr| addr.is_ipv4());
+            let all_ipv6 = wildcard.iter().any(|addr| addr.is_ipv6());
+            ifaces = get_if_addrs::get_if_addrs()
+                .unwrap_or_else(|e| {
+                    error!("Failed to get local interface addresses: {}", e);
+                    Default::default()
+                })
+                .into_iter()
+                .map(|iface| iface.ip())
+                .filter(|ip| (all_ipv4 && ip.is_ipv4()) || (all_ipv6 && ip.is_ipv6()))
+                .collect();
+            ifaces.sort();
         }
-    }
+
+        let urls = ifaces
+            .into_iter()
+            .map(|addr| match addr {
+                IpAddr::V4(_) => format!("{}:{}", addr, miniserve_config.port),
+                IpAddr::V6(_) => format!("[{}]:{}", addr, miniserve_config.port),
+            })
+            .map(|addr| match miniserve_config.tls_rustls_config {
+                Some(_) => format!("https://{}", addr),
+                None => format!("http://{}", addr),
+            })
+            .map(|url| match miniserve_config.random_route {
+                Some(ref random_route) => format!("{}/{}", url, random_route),
+                None => url,
+            })
+            .map(|url| Color::Green.paint(url).bold().to_string())
+            .collect::<Vec<_>>();
+
+        urls.join(", ")
+    };
 
     let socket_addresses = miniserve_config
         .interfaces
--- a/src/main.rs
+++ b/src/main.rs
@@ -251,6 +251,10 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
 }
 
 /// Allows us to set low-level socket options
+///
+/// This mainly used to set `set_only_v6` socket option
+/// to get a consistent behavior across platforms.
+/// see: https://github.com/svenstaro/miniserve/pull/500
 fn create_tcp_listener(addr: SocketAddr) -> io::Result<TcpListener> {
     use socket2::{Domain, Protocol, Socket, Type};
     let socket = Socket::new(Domain::for_address(addr), Type::STREAM, Some(Protocol::TCP))?;
--- a/src/main.rs
+++ b/src/main.rs
@@ -196,30 +196,11 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
         }
     }
 
-    let socket_addresses = interfaces
+    let socket_addresses = miniserve_config
+        .interfaces
         .iter()
-        .map(|interface| {
-            format!(
-                "{interface}:{port}",
-                interface = &interface,
-                port = miniserve_config.port,
-            )
-            .parse::<SocketAddr>()
-        })
-        .collect::<Result<Vec<SocketAddr>, _>>();
-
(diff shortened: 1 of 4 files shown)


svenstaro:
Hey, so just to be clear: this is currently wip or shall I have a look?

aliemjay:
This is in progress. I made it a draft.
You can still review the first 3 points. I think they are finished.

src/main.rs
```diff
@@ -378,6 +363,18 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
         .map_err(|e| ContextualError::IoError("".to_owned(), e))
 }
 
+// Allows us to set low-level socket options:
```
svenstaro, line 366: Let's use proper docstrings (`///`) for documenting functions.

aliemjay: Okay.

What do you think about this overall?
I want to settle this to push #422 fix

svenstaro: I really like it. I'm just skeptical about cross-platform support which is very important to me. This needs to work the same on all operating systems (which arguably currently it doesn't really, especially on Windows).

aliemjay:
> I really like it. I'm just skeptical about cross-platform support which is very important to me. This needs to work the same on all operating systems (which arguably currently it doesn't really, especially on Windows).

Well, it should be cross platform: the `IPv6_ONLY` sockopt is a standard and the whole function is adapted from actix-web:
https://github.com/actix/actix-web/blob/f462aaa7b6129b633fa53935d0ee54dbc01b70ea/src/server.rs#L609

aliemjay:
Can we avoid binding to "::" altogether an bind only to "0.0.0.0" by default?
In this case it would be simpler and we should then drop https://github.com/svenstaro/miniserve/pull/500/commits/2a28e6b31e7773debdfd8983e62496cd4af3b9f4 as it would be no longer neccessary.

svenstaro:
I believe on Windows that would result in it not binding to IPv6 at all (but that would have to be tested). Perhaps it might be better to bind to every interface directly individually instead of using `0.0.0.0/::`. The current situation isn't great so I'm happy for alternative implementations.

aliemjay:
Binding to every interface individually is the worst option here:
* First, there is no portable cross-platform way to enumerate all local interfaces at all.
* Second, it has a different behaviour to binding to `0.0.0.0/::` - changing the network configuration, for example, or adding a new interface may render the server inaccessible because the bindings do not update accordingly. 

This leaves us with, I believe, only two options:

1. The solution I tried to implement here - Directly controlling low-level socket options: It is completely portable, it uses standard sockopts, and it is what `actix-web` actually uses under the hood. The only counterargument I can think of is it adds an extra complexity to such a small project.
2. The second option is to not bind to IPv6 by default: Bind only to `0.0.0.0` and the user can always choose to bind to IPv6 by using `-i ::`.
In this way, the default behaviour is reliable across platforms while IPv6 is still a supported feature that the user can use, but the command line argument he should then use depends on his OS and network configuration. A windows user, foe example, shall realize that he needs to bind to both `-i 0.0.0.0 -i ::` in order to support IPv6, and a linux user will quickly realize, with the help of the error messages implemented here, that binding to both addresses is not allowed and binding to `-i ::` is sufficient.

aliemjay:
To implement #422, I pulled `get_if_addrs` crate. It is very simple and cross-platform. The only problem with it is it is no longer maintained!
But I don't see this as a problem because it is only used here for a convenience feature and it is popular enough to trust and has a relatively small code base and it is used here in such a way that any error while getting interface addresses only prints a warning and resumes normally. But I realize that this is controversial.

Other options I have explored are `interfaces`, `nix`, `pnet`, but those are either too heavy or support only unix.
Other less popular options: https://github.com/mensi/netifs-rs , https://github.com/misuzu/rust-interfaces

aliemjay:
> To implement #422, I pulled `get_if_addrs` crate. 
 I tried to use `pnet_datalink`, but unfortunately it turned out it requires libpcap for windows! Reverting this soon...

aliemjay:
To summarize the decision tree regarding the binding behavior issue:

* Bind by default to IPv4 only and let the user deal with the platform-specific behavior of binding to IPv6.
* Or, make consistent cross-platform behavior. In this case the options are:
    * Bind to each interface individually. I mentioned some downsides to this approach previously.
    * Set `ipv6only` sockopt manually (more code and dependencies)
        * Do we set it to `true` or `false` ?

aliemjay:
`README` needs an update!

aliemjay:
@svenstaro I'd appreciate your input here while rebasing this.

tests/bind.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+#[rstest]
+#[case(&[] as &[&str])]
+#[case(&["-i", "::"])]
+#[case(&["-i", "127.0.0.1"])]
+#[case(&["-i", "0.0.0.0"])]
+#[case(&["-i", "::", "-i", "0.0.0.0"])]
+#[case(&["--random-route"])]
+fn validate_prtinted_urls(tmpdir: TempDir, port: u16, #[case] args: &[&str]) -> Result<(), Error> {
```
svenstaro, line 59: Typo: `validate_printed_urls`

tests/bind.rs
```diff
@@ -0,0 +1,92 @@
...
+#[case(&["-i", "012.123.234.012"])]
+#[case(&["-i", "::", "-i", "012.123.234.012"])]
+fn bind_fails(tmpdir: TempDir, port: u16, #[case] args: &[&str]) -> Result<(), Error> {
+    Command::cargo_bin("miniserve")?
+        .arg(tmpdir.path())
+        .arg("-p")
+        .arg(port.to_string())
+        .args(args)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
```
svenstaro, line 23: Check for the actual error returned here. Not all of it should be a OS error. For instance, it should be possible to check `stderr` for at least `"Failed to bind server"`.

src/main.rs
```diff
@@ -253,6 +230,19 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
         .map_err(|e| ContextualError::IoError("".to_owned(), e))
 }
 
+/// Allows us to set low-level socket options
```
svenstaro, line 233: Please add some more comments here with some context as to why we do this, perhaps even link the PR for the discussion in addition as this is a fairly unobvious low-level thing that we have going on here.

svenstaro (changes requested):
A few thoughts in no order:
- Let's order the interface output listing vertically instead of in a long line.
- In your proposed solution 1., do we have to use some OS-specific logic to make the behavior identical between platforms or does it just work?
- I'm generally happy with your solution 1. as I think it's conceptually cleaner with not too much low-level code to maintain in our code base as a result but I'm a little bit worried about compatibility still and we'd definitely have to test it everywhere.
- All in all, I'm pretty happy with the implementation so far and I think we should go ahead with it but it will require some manual testing on the three major OS at least.

aliemjay:
>     * In your proposed solution 1., do we have to use some OS-specific logic to make the behavior identical between platforms or does it just work?

As long as the supported OS's support the IPv6Only option (which all does AFAIK), we don't need to include OS-specific code.

svenstaro:
Please rebase again so that we can see whether these changes build on all target platforms.

svenstaro:
Huh somehow CI doesn't run at all now?

svenstaro:
Ah, probably I need to add back to run it on PR as well.

svenstaro:
I pushed a fix that should do it, please try rebasing again.

svenstaro:
Also, you probably need to update the lock file.

svenstaro:
Looks like it builds fine on all platforms. I'll have access to my Windows machine in a bit and then I should be able to properly test drive this. In the mean time, please also go ahead and test this on as many platforms as you can and see how you like it yourself.

svenstaro:
The tests appear to be timing out with this PR. Can you take a look?

svenstaro:
If you want to rebase again, the changes on master will enable to use easily retrieve all the compiled artifacts from the PR which will allow for much easier cross OS testing. :)

svenstaro:
Evaluating this a bit further, I think we should change the behavior a little bit:
1. Always display to the user what we're binding.
2. Don't advertise the shown IP addresses as being what we're binding but rather provide them as a non-exhaustive list of addresses where miniserve is then available from. Something like
```
Bound to 0.0.0.0, ::1, ::1337
Serving path /tmp
Available at (non-exhaustive list):
- http://10.0.0.1:8080
- http://192.168.1.1:8080
- http://[::1]:8080
- http://127.0.0.1:8080
- http://[::1337]:8080
```
This way, we don't have to lie about users in terms of what we're binding but we can still provide a nice clickable list of output URLs where miniserve is most probably available on. What do you think? If you want to go ahead with this, feel free to change the list format and output to something nicer. This is just what I came up with. :)

aliemjay:
Oh, I've just seen you comment. This is quite reasonable! fixing it now..

src/main.rs
```diff
@@ -240,40 +220,59 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
...
+
+    println!("Bound to {}", display_sockets.join(", "));
+
+    println!("Serving path {}", Color::Yellow.paint(path_string).bold());
 
     println!(
-        "Serving path {path} at {addresses}",
-        path = Color::Yellow.paint(path_string).bold(),
-        addresses = addresses,
+        "Availabe at (non-exhaustive list):\n\t{}\n",
```
svenstaro, line 247: I think it'd look better if this were four spaces instead of the 8 spaces tab.

src/main.rs
```diff
@@ -236,10 +239,13 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
...
+    println!("Bound to {}", display_sockets.join(", "));
+
+    println!("Serving path {}", Color::Yellow.paint(path_string).bold());
+
     println!(
-        "Serving path {path} at:\n\t{addresses}\n",
-        path = Color::Yellow.paint(path_string).bold(),
-        addresses = addresses,
+        "Availabe at (non-exhaustive list):\n    {}\n",
+        display_urls.join("\n    "),
```
aliemjay, line 248: I wounder if there is a better way to encode spaces?

svenstaro: Eh, don't sweat it. We don't need to golf _everything_. :P

svenstaro (approved):

svenstaro:
Amazing work all in all. It's a fairly hard problem to solve in a cross-platform fashion but I think this is the best possible compromise!

PR #506, opened 2021-04-24
tests: Refactor!

aliemjay:
Remove duplicate code responsible for the initial setup and teardown of
the test binary.

This introduces `TestServer` as a resource manager for a running
miniserve binary, which can be created with the fixtures `server()` and
`server_no_stderr()`

It also provides convenience function for handling server url.

### DIFF
--- a/tests/auth.rs
+++ b/tests/auth.rs
@@ -1,17 +1,12 @@
 mod fixtures;
 
-use assert_cmd::prelude::*;
-use assert_fs::fixture::TempDir;
-use fixtures::{port, tmpdir, Error, FILES};
+use fixtures::{server, server_no_stderr, Error, FILES};
 use pretty_assertions::assert_eq;
 use reqwest::blocking::Client;
 use reqwest::StatusCode;
 use rstest::rstest;
 use select::document::Document;
 use select::predicate::Text;
-use std::process::{Command, Stdio};
-use std::thread::sleep;
-use std::time::Duration;
 
 #[rstest(
     cli_auth_arg, client_username, client_password,
@@ -28,26 +23,14 @@ use std::time::Duration;
     ),
 )]
 fn auth_accepts(
-    tmpdir: TempDir,
-    port: u16,
     cli_auth_arg: &str,
     client_username: &str,
     client_password: &str,
 ) -> Result<(), Error> {
-    let mut child = Command::cargo_bin("miniserve")?
-        .arg(tmpdir.path())
-        .arg("-p")
-        .arg(port.to_string())
-        .arg("-a")
-        .arg(cli_auth_arg)
-        .stdout(Stdio::null())
-        .spawn()?;
-
-    sleep(Duration::from_secs(1));
-
+    let server = server(&["-a", cli_auth_arg]);
     let client = Client::new();
     let response = client
-        .get(format!("http://localhost:{}", port).as_str())
+        .get(server.url())
         .basic_auth(client_username, Some(client_password))
         .send()?;
 
@@ -59,8 +42,6 @@ fn auth_accepts(
     for &file in FILES {
         assert!(parsed.find(Text).any(|x| x.text() == file));
     }
-
-    child.kill()?;
 
     Ok(())
 }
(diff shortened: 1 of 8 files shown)


tests/fixtures/mod.rs
```diff
@@ -76,3 +81,85 @@ pub fn tmpdir() -> TempDir {
...
+    }
+    pub fn port(&self) -> u16 {
+        self.port
+    }
+}
+
+impl Drop for TestServer {
+    fn drop(&mut self) {
+        self.child.kill().expect("Couldn't kill test server");
+        /* TODO may need .wait() ? */
```
Jikstra, line 163: Adding 
```
let ecode = self.child.wait()
                 .expect("failed to wait on child");

assert!(ecode.success());
```

should not harm. Rust docs say `On some systems, calling wait or similar is necessary for the OS to release resources. A process that terminated but has not been waited on is still around as a "zombie". Leaving too many zombies around may exhaust global resources (for example process IDs).` https://doc.rust-lang.org/std/process/struct.Child.html

tests/serve_request.rs
```diff
@@ -130,79 +101,47 @@ fn serves_requests_hidden_files(tmpdir: TempDir, port: u16) -> Result<(), Error>
...
-        .stdout(Stdio::null());
-    if no_symlinks {
-        comm.arg("--no-symlinks");
-    }
-
-    let mut child = comm.spawn()?;
-    sleep(Duration::from_secs(1));
-
+fn serves_requests_symlinks(no_symlinks: bool) -> Result<(), Error> {
+    let server = server(["--no-symlinks"].iter().filter(|_| no_symlinks));
```
Jikstra, line 126: I think this is easier to read:
```
let server = server(if no_symlinks {
    &["--no-symlinks"]
} else {
    &[] as &[&str]
});
```

Jikstra (commented):
Looks very good to me, but i think it makes sense to add the wait code and/or test it on windows (I guess that's the os that makes this trouble most likely)

tests/fixtures/mod.rs
aliemjay: The "or similar" in the documentation confused me especially `.kill()` seems to leave no zombies on system! But I agree. A `.wait()` is reasonable.

svenstaro:
I really like this and had something similar in mind! Will get to it soon-ish.

tests/qrcode.rs
```diff
@@ -1,74 +1,35 @@
...
-        .arg("-p")
-        .arg(port.to_string())
-        .stdout(Stdio::null())
-        .stderr(Stdio::null())
-        .spawn()?;
-
-    sleep(Duration::from_secs(1));
-
+fn get_svg_qrcode(server_no_stderr: TestServer) -> Result<(), Error> {
+    let server = server_no_stderr;
```
svenstaro, line 30: In this and other occurrences where you do this, can you instead use the [rename feature](https://docs.rs/rstest/0.11.0/rstest/attr.fixture.html#rename) of rstest?

tests/serve_request.rs
svenstaro: I'd prefer it as provided by @Jikstra in this example.

tests/serve_request.rs
```diff
@@ -130,79 +101,47 @@ fn serves_requests_hidden_files(tmpdir: TempDir, port: u16) -> Result<(), Error>
...
         symlink_dir(orig, link).expect("Couldn't create symlink");
     }
     for &file in files {
-        let orig = Path::new(FILES[0]);
-        let link = tmpdir.path().join(Path::new(file));
-        symlink_file(orig, link).expect("Couldn't create symlink");
+        symlink_file(FILES[0], server.path().join(file)).expect("Couldn't create symlink");
     }
     for &file in broken {
-        let orig = Path::new("should-not-exist.xxx");
```
svenstaro, line 199: Can you keep it `"should-not-exist.xxx"`? I want the `not` to stand out.

tests/upload_files.rs
```diff
@@ -1,33 +1,17 @@
...
 use rstest::rstest;
 use select::document::Document;
 use select::predicate::{Attr, Text};
-use std::process::{Command, Stdio};
-use std::thread::sleep;
-use std::time::Duration;
 
-#[rstest]
-fn uploading_files_works(tmpdir: TempDir, port: u16) -> Result<(), Error> {
+#[rstest(server(&["-u"]))]
```
svenstaro, line 9: Isn't this the [old compact syntax](https://docs.rs/rstest/0.11.0/rstest/attr.rstest.html#old-compact-syntax)? If so, could you adopt this to the new way? I suppose what you're doing here is partial injection and the new syntax for that is [here](https://docs.rs/rstest/0.11.0/rstest/attr.fixture.html#partial-injection).

If you agree, please change this here and in other places where you use this idiom.

tests/fixtures/mod.rs
svenstaro: Let's resolve this. Can you switch it to `wait()`? If this fails too often, we'll kill instead.

tests/fixtures/mod.rs
```diff
@@ -76,3 +81,85 @@ pub fn tmpdir() -> TempDir {
 pub fn port() -> u16 {
     free_local_port().expect("Couldn't find a free local port")
 }
+
+/// Run miniserve as a server; Start with a temporary directory, a free port and some
+/// optional arguments then wait for a while for the server setup to complete.
+#[fixture(args=&[] as &[&str])]
```
svenstaro, line 87: Could you switch this to the [modern way](https://docs.rs/rstest/0.11.0/rstest/attr.fixture.html#default-values)?

svenstaro (changes requested):
These are all great changes! Sorry for procrastinating on this for so very long. I'll take care to get this and your other improvements merged over the next couple of days. A few nitpicks and then we're good to merge!

tests/auth.rs
```diff
@@ -183,37 +134,22 @@ fn auth_multiple_accounts_pass(
...
-            .arg(port.to_string())
-            .stdout(Stdio::null())
-            .stderr(Stdio::null()),
-    )
-    .spawn()?;
-
-    sleep(Duration::from_secs(1));
-
+fn auth_multiple_accounts_wrong_username() -> Result<(), Error> {
+    let server = server_no_stderr(ACCOUNTS);
```
svenstaro, line 142: You ought to be able to do this via partial injection as linked before. There are more occurrences of this.

tests/fixtures/mod.rs
```diff
@@ -76,3 +81,85 @@ pub fn tmpdir() -> TempDir {
...
+        .expect("Couldn't find test binary")
+        .arg(tmpdir.path())
+        .arg("-p")
+        .arg(port.to_string())
+        .args(args)
+        .stdout(Stdio::null())
+        .spawn()
+        .expect("Couldn't run test binary");
+
+    sleep(Duration::from_secs(1));
```
svenstaro, line 105: How about we switch this fairly annoying manual wait to something that's faster and less error prone in this PR as well?

Something like:
```
        // Wait a max of 1s for the port to become available.
        let start_wait = Instant::now();
        while start_wait.elapsed().as_secs() < 1
            && !port_check::is_port_reachable(format!("localhost:{}", port))
        {
            sleep(Duration::from_millis(100));
        }
```

tests/fixtures/mod.rs
```diff
@@ -76,3 +81,85 @@ pub fn tmpdir() -> TempDir {
...
+    let port = port();
+    let tmpdir = tmpdir();
+    let child = Command::cargo_bin("miniserve")
+        .expect("Couldn't find test binary")
+        .arg(tmpdir.path())
+        .arg("-p")
+        .arg(port.to_string())
+        .args(args)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
```
svenstaro, line 125: Reading into these tests again, was there ever a reason to not always ignore stderr AND stdout? Perhaps you could test whether we can get away with using just a single fixture.

svenstaro (changes requested):

svenstaro:
After this is merged, please start rebasing your other PRs with tests onto this one as I want to merge everything swiftly. :) Thank you for your great contributions!

PR #507, opened 2021-04-24
Add openssl support for miniserve.

zhageww:
Add openssl support.

To use security http, users can launch miniserve with two more args: 
-C and -K
They are cacert file and key file.
NB: these two args are required simultaneously.

Signed-off-by: Ming Yang <my18345@alumni.bristol.ac.uk>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,6 +33,8 @@ dependencies = [
  "futures-util",
  "http",
  "log",
+ "openssl",
+ "tokio-openssl",
  "trust-dns-proto",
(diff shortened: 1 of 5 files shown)


svenstaro:
Hello, thanks for your contribution! I'd certainly like to add TLS support but I'd prefer to use rustls so that I can keep shipping an unproblematic single executable. Any chance you could change this?

svenstaro:
Closing this as miniserve now has offiical TLS support via rustls. Please open an issue if you'd still like actual OpenSSL support.

Issue #502, opened 2021-04-23
[feature request] add the ability to create directory


david0u0:
Hi,

I've recently started use miniserve and think it would be great to have a button/api to create a directory, but after searching for a while it seems this is currently not available.

With this feature, users can easily write a script to "uplaod a directory recursively". This can be toggle by command line flag, such as `--enable-mkdir`

If this feature is desirable, I'd like to help submit a PR. Thanks again for the great crate

svenstaro:
Yeah sounds good. I'd accept a PR for that.

Issue #214, opened 2019-11-06
suggestion:enable uploading folders


piaoger:
Uploading folders is quite useful..

I'd like to switch [thecoshman/http](https://github.com/thecoshman/http) to miniserve but gave up for lacking of uploading folders. oops ：）

svenstaro:
I'll take a look.

hashFactory:
I second this. Would be very useful! :)

piaoger:
any plan on this issue？

svenstaro:
Does the newly merged #781 resolve this for you?

PR #512, opened 2021-04-28
add the ability to create directory

david0u0:
For #502 

Screenshot
Screenshot from 2021-04-28 22-36-43

* forbid creating directory with colliding name
* forbid creating directory with `parent dir` or `rootdir` in name

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,13 @@ Sometimes this is just a more practical and quick way than doing things properly
 
 (where `$FILE` is the path to the file. This uses miniserve's default port of 8080)
 
+### Creating a directory using `curl`:
+
+    # in one terminal
+    miniserve -m .
+    # in another terminal
+    curl -d "mkdir_name=test_directory" http://localhost:8080/mkdir\?path\=/
+
 ## Features
 
 - Easy to use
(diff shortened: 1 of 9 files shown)


aliemjay:
Great job!

Regarding the api, I think it would be great to use the existing api (ie `/upload` handler) in such a way that to create directory:
`$ curl -F "mkdir=./b" http://localhost:123/upload?path=/`
This can be easily implemented, I imagine, as as a new handler in `file_upload::handle_multipart()` for the new field name `mkdir`.

My argument is that it would greatly reduce the patch size and avoid duplicate code and this would also allow us to easily  implement  recursive upload (#214) with minimal change. One idea I have for uploading folders in a single request is:
`$ curl -F mkdir=folder1 -F file_upload=@$FILE1 -F mkdir=folder2 -F file_upload=@$FILE2 ...`

What do you think? If you like the idea you can rebase your changes on #513. I don't know if @svenstaro approves such workflow though!

svenstaro:
@david0u0 are you interested in following @aliemjay's advice and rebasing this?

svenstaro:
I'm going to close this for now as the original author appears to have lost interest and this is very out of date at this point.

Issue #474, opened 2021-03-22
[feature request] add an ability to serve SPA (single page application) app 


azzamsa:
Hi.

IMHO. `miniserve` doesn't able to serve SPA app.
Below is the demo. `quiz-seed` is quiz app built using `seed.rs`.

``` bash
quiz-seed/dist on  master [!?] via ⬢ v14.15.4
❯ miniserve . --index index.html

03:37:11 [ERROR] Route /quiz could not be found
03:37:14 [ERROR] Route /point could not be found
03:37:15 [ERROR] Route /browser-sync/socket.io/ could not be found
03:37:21 [ERROR] Route /browser-sync/socket.io/ could not be found
03:37:27 [ERROR] Route /browser-sync/socket.io/ could not be found
03:37:33 [ERROR] Route /browser-sync/socket.io/ could not be found
03:37:39 [ERROR] Route /browser-sync/socket.io/ could not be found
^C⏎
```

In the browser:

![azra-quiz-miniserve](https://user-images.githubusercontent.com/17734314/111940372-2c561800-8b01-11eb-95a2-31511de4652d.gif)

I tried the same step with `python3 -m http.server 8080` and had the same result as `miniserve`.
It can't find `quiz` and `point` page.

---

Here I try using `local-web-server.js`
It has `--spa` parameter.

```bash
quiz-seed on  master [!?] is 📦 v0.1.0 via ⬢ v14.15.4 via 🦀 v1.50.0
❯ yarn global add local-web-server

quiz-seed/dist on  master [!?] via ⬢ v14.15.4 took 3s
❯ ws --spa index.html
Listening on http://debianm:8000, http://127.0.0.1:8000, http://192.168.8.101:8000
^C⏎   
```

In the browser:

![azra-quiz-ws](https://user-images.githubusercontent.com/17734314/111940495-73dca400-8b01-11eb-9f05-f097504e43f6.gif)

Thanks for miniserve!

svenstaro:
Hm, yeah for SPAs it's not quite clear that one shouldn't really currently use `--index` as it is right now. I suppose for an SPA you'd simply provide the path to the files and then just navigate there. I'm not sure `--index` should allow you to access other files as well. Perhaps we could have two flags.

azzamsa:
I also tried using: 
- `❯ miniserve index.html`
- `❯ miniserve .`

Both had similar result:

```bash
quiz-seed/dist on  master [!?] via ⬢ v14.15.4
[...]

PR #515, opened 2021-05-01
Add --spa-index option

sinking-point:
Closes https://github.com/svenstaro/miniserve/issues/474

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -31,6 +31,14 @@ pub struct CliArgs {
     #[structopt(long, parse(from_os_str), name = "index_file")]
     pub index: Option<PathBuf>,
 
+    /// The index file of a single page application
+    ///
+    /// If this option is set, miniserve will serve the specified file instead of a 404 page when
+    /// a non-existent path is requested. This is intended for single-page applications where
+    /// routing takes place on the client side.
+    #[structopt(long, parse(from_os_str), name = "spa_index_file")]
+    pub spa_index: Option<PathBuf>,
+
     /// Port to use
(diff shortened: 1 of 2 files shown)


aliemjay:
Good job!
This would be even easier to implement with https://github.com/actix/actix-web/pull/2135

Now I wonder if `--spa-index` and `--index` options should be compatible or if there is any use case for both of them together?
If not, then I think it is better to have a `--spa`  flag instead of `--spa-index`. In this way, we avoid conflicting command line options. So then the usage would be something like:
`$ miniserve --spa --index ./index.html`

sinking-point:
Thanks. I'm aware of the PR, but those changes aren't yet released and I needed this feature sooner rather than later.

Hmm, your suggestion is certainly more intuitive. But if you want to have an SPA that is not at the root of your site then having the flexibility of my implementation would be useful.

sinking-point:
That is something of an edge case though. I can change it to be how you suggested if you like.

aliemjay:
Well, I don't have a strong opinion on this because I don't use SPAs!

I think we need a second opinion @svenstaro .

svenstaro:
Is the behavior of the current `--index` as you would expect for your needs, @sinking-point? There were a ton of changes to miniserve behavior recently and this might just be fixed for your use case now.

sinking-point:
@svenstaro I have just tried with the latest version and unless I'm missing something the problem is the same. Is there a specific change that you expected to resolve this?

svenstaro:
Clippy doesn't seem happy yet. :)

svenstaro:
@sinking-point Can you fix this real quick?

PR #521, opened 2021-05-06
add special colors for visited links

raffomania:
As it is, there's no way to see which files I've already downloaded and which directories I've visited.

This PR adds extra colors for each theme, for both file and directory links.

It might be nice to add a visited color for the download links at the top as well - I think adding them for the `back` links would not be good UX, though.


svenstaro:
This _is_ a tasty pull request! :hamburger:

Issue #497, opened 2021-04-18
Consider turning on transfer compression


svenstaro:
actix-web has a ready-to-go [compress middleware](https://docs.rs/actix-web/3.3.2/actix_web/middleware/struct.Compress.html). I currently don't really see any downsides with turning that on. Will have to do a little benchmarking to see whether it's ok to generally turn this on.

deantvv:
After reading NGINX document of compression [link](https://docs.nginx.com/nginx/admin-guide/web-server/compression/), here is some of my thoughts from it.

> By default, NGINX compresses responses only with MIME type text/html.

This is clear now why the test benchmark in #523 looks so bad, and NGINX by default do not enable this except text/html.

NGINX tried to work around this by introducing a new option called `gzip_static`.
> Note that the gzip_static directive does not enable on-the-fly compression. It merely uses a file compressed beforehand by any compression tool.

In this option, you need to provide a gzip version of the file before the request comes in. This will save network bandwidth and also not increase CPU load by a lot (only a simple directory looking for the compressed file). 

It seems to me that both options may be overkilled to miniserve. Especially the first line of the README for miniserve is "For when you really just want to serve some files over HTTP right now!".

PR #523, opened 2021-05-08
Compress response using `actix_web::middleware::compress`

deantvv:
Using `actix-web` compress middleware to compress the response. Simple benchmark is done using [gtmetrix](https://gtmetrix.com) which shows that the size of content (html, css, svg) shrink from 32.1kB -> 8.6 kB and the page load time is also reduced from 678ms -> 568ms. Though this isn't a thorough benchmark, it shows that it can gain some performance and reduce internet transmission size. Even if the user isn't interested in this feature, the browser could set `accept-encoding` to empty string to opt-out compression.

Related to #497

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -355,6 +355,7 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
                 HttpAuthentication::basic(auth::handle_auth),
             ))
             .wrap(middleware::Logger::default())
+            .wrap(middleware::Compress::default())
             .route(
                 &format!("/{}", inside_config.favicon_route),
                 web::get().to(favicon),


svenstaro:
Could you do a benchmark for how this influences actual file transfer performance? Does it also influence for instance large incompressible files and then how's the performance I wonder?

deantvv:
@svenstaro You are right. There is **regression** when downloading big file. I do the following test with curl downloading 10GB file. When using `compressed` flag, the time spent increase from 16s to 56s. Although the internet transmission size will be considerably smaller, but the transmit speed penalty is too large and the result is way slower than before.

I think we should add `compress` as an option and let users decide whether to enable this or not. IMO, there would be some cases when the internet bandwidth is scarce or the users would like to control the compression behavior through `accept-encoding`.

Test result is attached below.

```
# Test 1 (Master)
100 10.0G  100 10.0G    0     0   612M      0  0:00:16  0:00:16 --:--:--  617M

# Test 2 (Compress-response w/o --compressed)
100 10.0G  100 10.0G    0     0   618M      0  0:00:16  0:00:16 --:--:--  624M

# Test 3 (Compressed-response w/ --compressed)
100 46.5M    0 46.5M    0     0   847k      0 --:--:--  0:00:56 --:--:-- 2334k

```

svenstaro:
I'd be fine with adding a flag (and test) that conditionally enables compression. The help text should also explain that for fast connections it'll likely make transfers slower depending on CPU. Definitely off by default.

deantvv:
This will be closed and will open a new PR if the option version is ready. Thanks for the suggestion.

PR #524, opened 2021-05-08
Print QR code to terminal

### LINKED ISSUE #45: How about a QR-code in the terminal?

aliemjay:
Followup for #341 

Based on #500 

Closes #45

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -261,7 +261,7 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
         {
             match QrCode::encode_text(url, QrCodeEcc::Low) {
                 Ok(qr) => {
-                    println!("QR code for {} :", Color::Green.paint(url).bold());
+                    println!("QR code for {}:", Color::Green.paint(url).bold());
                     print_qr(&qr);
                 }
                 Err(e) => {
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,6 +12,7 @@ use clap::{crate_version, Clap, IntoApp};
 use clap_generate::generators::{Bash, Elvish, Fish, PowerShell, Zsh};
 use clap_generate::{generate, Shell};
 use log::{error, warn};
+use qrcodegen::{QrCode, QrCodeEcc};
 use yansi::{Color, Paint};
 
 mod archive;
@@ -183,7 +184,6 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
                 Some(ref random_route) => format!("{}/{}", url, random_route),
                 None => url,
             })
-            .map(|url| Color::Green.paint(url).bold().to_string())
             .collect::<Vec<_>>()
     };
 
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -46,7 +46,7 @@ strum_macros = "0.21"
 sha2 = "0.9"
 hex = "0.4"
 zip = "0.5.11"
-qrcodegen = "1"
```
svenstaro, line 49: I'm not sure I like this. While I get why you're doing this, `qrcodegen` seems to be well-maintained while `qrcode` appears to be pretty dead. Probably QR code stuff doesn't need a lot of maintenance but I'm still a bit unhappy about this. Any chance to make `qrcodegen` do the work?

svenstaro: See for instance https://github.com/nayuki/QR-Code-generator/blob/master/rust/examples/qrcodegen-demo.rs.

svenstaro (changes requested):

Cargo.toml
aliemjay: that would be low

> Any chance to make `qrcodegen` do the work?

I'll give it a try!

src/main.rs
```diff
@@ -246,9 +246,31 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
...
 
+    // print QR code to terminal
+    if miniserve_config.show_qrcode && atty::is(atty::Stream::Stdout) {
+        for url in display_urls
+            .iter()
+            .filter(|url| !url.contains("//127.0.0.1:") && !url.contains("//[::1]:"))
+        {
+            match QrCode::encode_text(&url, QrCodeEcc::Low) {
+                Ok(qr) => {
+                    println!("QR code for {} :", Color::Green.paint(url).bold());
```
svenstaro, line 264:
```suggestion
                    println!("QR code for {}:", Color::Green.paint(url).bold());
```

src/main.rs
```diff
@@ -338,3 +360,21 @@ async fn css() -> impl Responder {
         .insert_header(ContentType(mime::TEXT_CSS))
         .message_body(css.into())
 }
+
+// Prints the given QrCode object to the console.
+fn print_qr(qr: &QrCode) {
+    let border: i32 = 4;
+    for y in (-border..qr.size() + border).step_by(2) {
```
svenstaro, line 367: Can you print a two character-wide white border around these?

svenstaro (changes requested):
Pretty clean but somehow it doesn't scan for me.

src/main.rs
```diff
@@ -246,9 +246,31 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
...
-        display_urls.join("\n    "),
+        display_urls
+            .iter()
+            .map(|url| Color::Green.paint(url).bold().to_string())
+            .collect::<Vec<_>>()
+            .join("\n    "),
     );
 
+    // print QR code to terminal
+    if miniserve_config.show_qrcode && atty::is(atty::Stream::Stdout) {
```
svenstaro, line 257: Are we perhaps going to use a new CLI flag for this? As it stands, this one is used for presenting the QR code in the web UI. Not sure about this.

src/main.rs
aliemjay: I'm not sure I understand this properly~ I mean the content is already white. If doesn't scan then the problem more probably is with the rendering code `print_qr` because I modified it from https://github.com/nayuki/QR-Code-generator/blob/master/rust/examples/qrcodegen-demo.rs to show smaller code. But I have tested it locally and it scans well.

Can you try another scanner to pinpoint the cause here?

src/main.rs
aliemjay: I think it is too much to have two different options for QR code. though I have no strong opinion on this. I'd add another option if you want.

svenstaro: Eh let's leave it like it is until someone complains.

src/main.rs
svenstaro: I tested it and it works on my CLI if the colors are reversed (and with a border). If you want to try it:
`echo "lolomg" | qrencode -t UTF8` works but ``echo "lolomg" | qrencode -t UTF8i` doesn't.

Suggestion, since we have lots of horizontal space anyway: Just put the inversed and bordered version right next to the other one. This way, we'll make sure it works for everybody.

svenstaro (changes requested):

PR #513, opened 2021-04-29
[Refactor] file_upload.rs: Use async/await instead of explicit futures

aliemjay:
... and rewrite `file_upload::{save_file, handle_multipart}` since they are greatly modified.
... and fix clippy warning about `create_error_response()` being always `Ok()`

PR #529, opened 2021-05-12
[Refactor] Fix clippy::too_many_arguments and rework error page rendering

### LINKED ISSUE #588: Error page isn't rendered with the correct theme
It seems the error page is always rendered with the default theme and not the user's selected theme.

aliemjay:
Too many arguments are moved around and many of them are already stored in `MiniserveConfig`. Many of these are used to render error pages.

To fix this issue, it was necessary to rework error page rendering:

1. Implement `ResponseError` for `ContextualError` so that it can be
   returned from service handlers as is and will then be automatically
   logged to the console and converted into an error response.

2. At service handler level, all error responses are now rendered as
   plain text.

3. `error_page_middleware` is now responsible for the rendering of the
   final error page from plain text responses.

Based on #513 

Closes #588

### DIFF
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -1,6 +1,5 @@
 use actix_web::body::Body;
 use actix_web::dev::ServiceResponse;
-use actix_web::http::StatusCode;
 use actix_web::web::Query;
 use actix_web::{HttpRequest, HttpResponse};
 use bytesize::ByteSize;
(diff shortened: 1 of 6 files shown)


aliemjay:
Hi @svenstaro,

I currently base my new PRs on some previous PRs to avoid huge merge conflicts when merging. What do you think of such workflow?  I found this necessary because PRs are piling up.

aliemjay:
Is such code refactoring welcome? I hope it is not invasive!

aliemjay:
@svenstaro I hope this gets reviewed next

svenstaro:
Sure but let's get #500 in first. We're almost there! \o/

svenstaro (changes requested):
I get a plain (unthemed) error page for invalid or missing auth here. Granted, this also was the previous behavior but how about we fix this in the same PR? The 404 error, for instance, is already themed. It would make sense for 401/403 to also be themed and it would be more consistent, too.

svenstaro:
While testing this, I noticed #588 which you could address in this PR as well (or it could just be a follow-up).

aliemjay:
> I get a plain (unthemed) error page for invalid or missing auth here. Granted, this also was the previous behavior but how about we fix this in the same PR? The 404 error, for instance, is already themed. It would make sense for 401/403 to also be themed and it would be more consistent, too.

Is it even possible to get themed in invalid auth? I mean unless we embed css.

aliemjay:
> While testing this, I noticed #588 which you could address in this PR as well (or it could just be a follow-up).

This should be easy. I'll implement it here.

svenstaro:
> > I get a plain (unthemed) error page for invalid or missing auth here. Granted, this also was the previous behavior but how about we fix this in the same PR? The 404 error, for instance, is already themed. It would make sense for 401/403 to also be themed and it would be more consistent, too.
> 
> Is it even possible to get themed in invalid auth? I mean unless we embed css.

I suppose the internal route to the media resources doesn't really need to be protected, no secrets will ever be kept there. The theme information is stored inside the browser itself so that should also be always available.

svenstaro:
The rest of it looks like a vast improvement to me! Probably there are some parts which could be golfed a little (like the middleware bit) but I wouldn't immediately know how to best golf it.

src/renderer.rs
```diff
@@ -469,48 +462,34 @@ fn humanize_systemtime(time: Option<SystemTime>) -> Option<String> {
...
+                .(format!("default_theme_dark_{}", conf.default_color_scheme_dark)) {
 
                 div.error {
                     p { (error_code.to_string()) }
                     @for error in error_description.lines() {
                         p { (error) }
                     }
-                    @if display_back_link {
+                    // WARN don't expose random route!
+                    @if !conf.random_route.is_some() {
```
svenstaro, line 485: Good catch!

src/listing.rs
```diff
@@ -185,13 +168,18 @@ pub fn directory_listing(
...
+            .clone()
+            .unwrap_or_else(|| req.connection_info().host().into());
 
         let decoded = percent_decode_str(&encoded_dir).decode_utf8_lossy();
 
         let mut res: Vec<Breadcrumb> = Vec::new();
-        let mut link_accumulator =
-            format!("/{}", random_route.map(|r| r + "/").unwrap_or_default());
+        let mut link_accumulator = match &conf.random_route {
+            Some(random_route) => format!("/{}/", random_route),
```
svenstaro, line 180: Yeah, while this is more verbose, I think it's much nicer to read.

aliemjay: This was a necessary evil, because otherwise it would be 7 lines long!

svenstaro:
I snuck in #542. Could you rebase on top of that? Sorry!

aliemjay:
#588 and themed 401 page should be fixed

svenstaro:
Very cool!

PR #563, opened 2021-08-05
Fix README.md instruction for bash completions and add Termux install instruction

pvonmoradi:
According to https://github.com/scop/bash-completion#faq , the completion file should reside in the `completions` subdir.


aliemjay (approved):

svenstaro:
Thanks!

PR #578, opened 2021-08-27
Add backslash to percent_encode_sets.PATH_SEGMENT

### LINKED ISSUE #545: Cannot retrieve file with \ in name
I was trying to reproduce a bug where files with `#` in the name wouldn't be served by an older version (that one seems fixed), and found that current master turns `\` into `/`.

I looked into the source a bit to find out why, remembered the whole set of problems with Paths and strings and `to_string_lossy` and promptly bailed. (Of course, it won't serve a file like `touch "$(echo 日本語 | iconv -f utf-8 -t shift-jis)"` either, and that's probably not easy to fix. I'm not sure that's common enough a problem to tackle.)

Jikstra:
Fixes #545

### DIFF
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -24,7 +24,7 @@ mod percent_encode_sets {
     const BASE: &AsciiSet = &CONTROLS.add(b'%');
     pub const QUERY: &AsciiSet = &BASE.add(b' ').add(b'"').add(b'#').add(b'<').add(b'>');
     pub const PATH: &AsciiSet = &QUERY.add(b'?').add(b'`').add(b'{').add(b'}');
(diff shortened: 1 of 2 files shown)


svenstaro:
Any chance for a test? I think it should be trivial as you can probably just add an entry to the file fixtures with this name.

Jikstra:
Oh yes! I somehow thought there was already a file with a backslash, but the backslash was just escaping another character ^^

Jikstra:
Ay the test now fails on windows, i guess creating a file with a backslash on windows panics?

aliemjay:
>  Is there some cool macro that we could add to that file so that it doesnt get created on windows?

`#[cfg(not(windows))]`

Issue #533, opened 2021-05-19
Can we specify the route digit?


wizyoung:
Currently we can use `--random-route` to set the generated random 6-hexdigit route. Is there a method to manually specify the route digit?

aliemjay:
Actually there is a PR for this feature #208

4920441:
Hmmm that issue bothers me as well, since behind a reverse-proxy miniserve makes lots of problems because the first subdir changes everytime, so the wrong url is called (the wrong subdir)

svenstaro:
This is now fixed by #728! :D

PR #534, opened 2021-05-20
Allow custom route prefix

aliemjay:
Followup for #208 
Based on #537

Closes #533
Closes #208

PR #508, opened 2021-04-25
Implement a raw rendering mode for recursive folder download

Jikstra:
- Raw mode only displays file/folders and focused on computer processing. So no js, no hints, no colors, no unneeded css classes...
- Show a banner in the footer displaying the command to recursively download the current folder with wget

# Screenshots
Screenshot_20210425_175359

Screenshot_20210425_175437

This PR is a rewrite of https://github.com/svenstaro/miniserve/pull/228

PR #579, opened 2021-08-27
 Add abillity to set path prefix

Jikstra:
This PR should replace the --random-route argument with both a --path-prefix argument and a conflicting --random-path-prefix argument.

The --random-path-prefix argument should be a fully functional replacement to --random-route.

This PR reimplements the code of @mkroman from this https://github.com/svenstaro/miniserve/pull/208 and adds a simple test

Fixes https://github.com/svenstaro/miniserve/issues/533


tests/serve_request.rs
```diff
@@ -220,3 +221,23 @@ fn serves_requests_custom_index_notice(tmpdir: TempDir, port: u16) -> Result<(),
 
     Ok(())
 }
+
+#[rstest]
+fn serves_requests_with_path_prefix() -> Result<(), Error> {
+    let route = "foobar";
+    let server = server(&["--path-prefix", route]);
+    let client = Client::new();
```
svenstaro, line 229: You can probably get away without this line by using `reqwest::blocking::get` directly below where you need it. This is what we do in other tests and would keep this consistent.

tests/serve_request.rs
```diff
@@ -220,3 +221,23 @@ fn serves_requests_custom_index_notice(tmpdir: TempDir, port: u16) -> Result<(),
...
+
+    let status = client.get(server.url()).send()?.status();
+
+    assert_eq!(status, StatusCode::NOT_FOUND);
```
svenstaro, line 233: I'd appreciate a comment here about your expectation at this point in the test.

svenstaro (approved):
Excellent stuff, just some minor nits. :)

tests/serve_request.rs
Jikstra: Sorry, i don't get it. What do you mean?

Jikstra:
I will rebase this after https://github.com/svenstaro/miniserve/pull/508 got merged :)

svenstaro:
Wanna rebase this, I'll review and then I'll cut a release?

src/config.rs
```diff
@@ -43,8 +43,11 @@ pub struct MiniserveConfig {
     /// Show hidden files
     pub show_hidden: bool,
 
-    /// Enable random route generation
-    pub random_route: Option<String>,
+    /// Use a specific path prefix
+    pub path_prefix: Option<String>,
```
aliemjay, line 47: I prefer this to be just `String`, where absent route prefix is represented as empty string. This would make it easy to use when building urls.

You may want to take a look at my previous PR https://github.com/svenstaro/miniserve/pull/534/commits/7ea3beb8de4ec134fb33b6808632d6a69d4e8161

tests/serve_request.rs
```diff
@@ -220,3 +221,23 @@ fn serves_requests_custom_index_notice(tmpdir: TempDir, port: u16) -> Result<(),
 
     Ok(())
 }
+
+#[rstest]
+fn serves_requests_with_path_prefix() -> Result<(), Error> {
```
aliemjay, line 226: another test case should be added to `tests/bind.rs:validate_printed_urls`.

tests/serve_request.rs
aliemjay, line 229: can you add more user input cases? I'm thinking of something like:

```rust
#[case("foo", "/foo")]
#[case("/foo", "/foo")]
#[case("/foo/", "/foo")]
fn test(#[case] prefix_arg, #[case] accessible_url) { }
```

aliemjay (changes requested):
Some pedantic review :)

tests/serve_request.rs
svenstaro: Basically, _why_ do you expect it to be `NOT_FOUND` there but `OK` later in the test? Give it a bit of a narrative. :)

Jikstra:
Closed in favor of https://github.com/svenstaro/miniserve/pull/682

PR #580, opened 2021-08-27
Conditionally enable TLS only on supported platforms

svenstaro:
Use rustls only on architecturs supported by ring. See also https://github.com/briansmith/ring/issues/1182 and https://github.com/briansmith/ring/issues/562 and https://github.com/briansmith/ring/issues/1367


svenstaro:
Good idea, I switched it.

PR #582, opened 2021-08-28
migrate to actix-web v4.0-beta

### LINKED ISSUE #462: [feature request] add an option to serve symlinks with the filename of the link, not the pointed-to file

### LINKED ISSUE #581: Bump actix-web to beta versions
The beta channel is pretty stable and used by many i production, it supports tokio v1.0 and would fix many open issues here.

And  as I see v4 is nowhere in the near future, I think it is reasonable to switch to beta.

I'm happy to submit a PR if it is approved @svenstaro .

aliemjay:
Closes #581 
Closes #462

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -1,8 +1,9 @@
-use actix_web::dev::ServiceRequest;
+use actix_web::dev::{Service, ServiceRequest, ServiceResponse};
 use actix_web::http::{header, StatusCode};
-use actix_web::{HttpRequest, HttpResponse, Result};
-use actix_web_httpauth::extractors::basic::BasicAuth;
+use actix_web::{HttpRequest, HttpResponse};
+use futures::future::Either;
 use sha2::{Digest, Sha256, Sha512};
+use std::future::{ready, Future};
 
 use crate::errors::{self, ContextualError};
 use crate::renderer;
@@ -14,12 +15,16 @@ pub struct BasicAuthParams {
     pub password: String,
 }
 
-impl From<BasicAuth> for BasicAuthParams {
-    fn from(auth: BasicAuth) -> Self {
-        Self {
+impl BasicAuthParams {
+    fn try_from_request(req: &HttpRequest) -> actix_web::Result<Self> {
+        use actix_web::http::header::Header;
+        use actix_web_httpauth::headers::authorization::{Authorization, Basic};
+
+        let auth = Authorization::<Basic>::parse(req)?.into_scheme();
+        Ok(Self {
             username: auth.user_id().to_string(),
             password: auth.password().unwrap_or(&"".into()).to_string(),
-        }
+        })
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/auth.rs
```diff
@@ -72,25 +77,48 @@ pub fn get_hash<T: Digest>(text: &str) -> Vec<u8> {
...
+    req: ServiceRequest,
+    srv: &S,
+) -> impl Future<Output = actix_web::Result<ServiceResponse>> + 'static
+where
+    S: Service<ServiceRequest, Response = ServiceResponse, Error = actix_web::Error>,
+    S::Future: 'static,
+{
+    match handle_auth(req) {
+        Ok(req) => Either::Left(srv.call(req)),
+        Err(resp) => Either::Right(ready(Ok(resp))),
```
aliemjay, line 121: Withe the new error model in actix, we could no longer use `actix_web_httpauth::middleware::HttpAuthentication` to send send customized error reponse in case of failed authorization. Hence, it was neccessary to implement our own middleware to handle authentication and only use utility functions from `actix_web_httpauth` crate.

src/listing.rs
```diff
@@ -169,25 +169,10 @@ pub fn directory_listing(
     dirs_first: bool,
     hide_version_footer: bool,
     title: Option<String>,
-) -> Result<ServiceResponse, io::Error> {
+) -> io::Result<ServiceResponse> {
     use actix_web::dev::BodyEncoding;
     let serve_path = req.path();
 
-    // In case the current path is a directory, we want to make sure that the current URL ends
-    // on a slash ("/").
```
svenstaro, line 177: Is this now handled directly by actix-web?

src/auth.rs
svenstaro: I don't quite understand. actix_web_httpauth 0.6.0-beta.2 should be compatible with actix-web 4.0.0-beta.8 in my understanding. While your fix is acceptable, I wonder whether the compatibility couldn't be established here somehow.

src/listing.rs
aliemjay: yes, see the last commit. It now uses `Files::redirect_to_slash_directory()`

src/auth.rs
aliemjay: I think it is broken. In case of failed authorization we can  no longer send customized error response; only a basic one would be sent automatically.

svenstaro:
Could you rebase and force push instead of that last merge? Then I'll merge into master.

aliemjay:
last minute thought: what do you think of abandoning sending custom body with the `Unauthrized` response. I don't think it is rendered at all to the user!

aliemjay:
This way we can keep the original code unmodified.

svenstaro:
> last minute thought: what do you think of abandoning sending custom body with the `Unauthrized` response. I don't think it is rendered at all to the user!

Yeah I think that's a fair tradeoff.

aliemjay:
Unfortuantely, even this is not possible :( I thought the the upgrade would be trivial but now I am quite sure that the middleware in `actix_web_httpauth` is broken.

If you want we can keep this hack to overcome this, or we can wait till this is fixed upstream but actix-web developement seems pretty slow.

svenstaro:
I'm happy to merge for the time being if you open an upstream issue and link it in the code somewhere.

aliemjay:
there already was an open issue for a while https://github.com/actix/actix-extras/issues/127

svenstaro:
Thanks, good stuff!

PR #614, opened 2021-10-08
Bump actix-web to v4.0-beta.9

aliemjay:
... and partially reverts #582 to use `HttpAuthentication` middleware because it turned out the it can be perfectly used with web-v4 and the linked issue is not the same as I thought we had. The docs are to blame in this case ;)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,9 +20,9 @@ dependencies = [
 
 [[package]]
 name = "actix-files"
-version = "0.6.0-beta.6"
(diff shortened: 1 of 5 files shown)


aliemjay:
CI fails due to regression  actix/actix-web#2398

I'll split this PR to update later

svenstaro:
@aliemjay Just to confirm: Can this be closed?

aliemjay:
sure, I forgot it.

PR #618, opened 2021-10-13
Patially revert #582

aliemjay:
Partially reverts #582 to use HttpAuthentication middleware because it turned out the it can be perfectly used with web-v4 and the linked issue is not the same as I thought we had. The docs are to blame in this case ;)


svenstaro:
Cool!

Issue #546, opened 2021-06-12
[feature request] Add ability to set accept attribute on file input


mayjs:
Hi,

I've been using miniserve to scan important documents using my smartphone. For this, I serve a directory on my PC using miniserve (allowing file uploads). On my phone I take some pictures and upload them to this directory. Afterwards I use some shell scripts to post process the images to PDF and apply OCR to it.

I found that I could get my phone to offer the option to take a picture directly from the browser if the `accept` attribute of the file input was set to `image/*`.
Could this be something that other people would use as well? I'd be happy to create a PR with a new command line option to enable this attribute.

PR #750, opened 2022-03-06
Add a flag to enable image uploads

mayjs:
This implements the feature I asked about in #546.
An additional flag allows to add the `accept="image/*"` attribute to the file input.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -104,6 +104,10 @@ pub struct CliArgs {
     #[clap(short = 'u', long = "upload-files")]
     pub file_upload: bool,
 
+    /// Enable image uploading
+    #[clap(short = 'I', long = "upload-images")]
+    pub image_upload: bool,
+
     /// Enable overriding existing files during file upload
     #[clap(short = 'o', long = "overwrite-files")]
     pub overwrite_files: bool,
--- a/src/config.rs
+++ b/src/config.rs
@@ -81,6 +81,9 @@ pub struct MiniserveConfig {
     /// Enable file upload
     pub file_upload: bool,
 
+    /// Set the input type for file uploads to image, allowing to take pictures from some browsers on smartphones
+    pub image_upload: bool,
+
     /// Enable upload to override existing files
     pub overwrite_files: bool,
 
@@ -188,6 +191,8 @@ impl MiniserveConfig {
         #[cfg(not(feature = "tls"))]
         let tls_rustls_server_config = None;
 
+        let file_upload = args.file_upload || args.image_upload;
+
         Ok(MiniserveConfig {
             verbose: args.verbose,
             path: args.path.unwrap_or_else(|| PathBuf::from(".")),
@@ -206,7 +211,8 @@ impl MiniserveConfig {
             spa: args.spa,
             overwrite_files: args.overwrite_files,
             show_qrcode: args.qrcode,
-            file_upload: args.file_upload,
+            file_upload,
+            image_upload: args.image_upload,
             tar_enabled: args.enable_tar,
             tar_gz_enabled: args.enable_tar_gz,
             zip_enabled: args.enable_zip,
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -107,7 +107,11 @@ pub fn page(
                                 form id="file_submit" action=(upload_action) method="POST" enctype="multipart/form-data" {
                                     p { "Select a file to upload or drag it anywhere into the window" }
                                     div {
-                                        input #file-input type="file" name="file_to_upload" required="" multiple {}
+                                        @if conf.image_upload {
+                                            input #file-input type="file" accept="image/*" name="file_to_upload" required="" multiple {}
+                                        } @else {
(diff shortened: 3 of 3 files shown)


src/args.rs
```diff
@@ -104,6 +104,10 @@ pub struct CliArgs {
     #[clap(short = 'u', long = "upload-files")]
     pub file_upload: bool,
 
+    /// Enable image uploading
+    #[clap(short = 'I', long = "upload-images")]
```
svenstaro, line 108: Should this flag also imply `file_upload` or should it be exclusive to that flag?

svenstaro (changes requested):
Interesting approach and neat use case for miniserve! Do you think there's value in actually verifying that a valid image has been uploaded? You can still upload anything you want even if it's not an image and nothing will actually check that.

src/args.rs
mayjs: Oh, can you actually specify something like that in clap? I've never used it before, so I solved this implication in `config.rs`. Come to think of it, this approach would theoretically allow to create inconsistent `MiniserveConfig` values, where `image_upload` is true, but `file_upload` is false. Maybe we should combine both of these values into an enum or just always specify the `accept` attribute and default it to `*`. Then we could have a CLI argument to configure any `accept` value you want, thus making this feature more flexible.

svenstaro: Yes, check for instance [this `conflicts_with`](https://github.com/svenstaro/miniserve/blob/eb3d565e1e67ff139a42ee2e8b0ea285df481e2a/src/args.rs#L68) or [this `requires`](https://github.com/svenstaro/miniserve/blob/eb3d565e1e67ff139a42ee2e8b0ea285df481e2a/src/args.rs#L174).

mayjs:
> Interesting approach and neat use case for miniserve! Do you think there's value in actually verifying that a valid image has been uploaded? You can still upload anything you want even if it's not an image and nothing will actually check that.

Yeah, miniserve allows to create small utilities like this really easily, thanks for creating it! I almost started to create my own web app for this before I found out about miniserve, this saved me a lot of time.

For my use case there would not be any real benefit of validating the contents - I'm the only user here and don't intend to make this publicly accessible.
I also think that validating the file types would be unnecessary overhead in general, since miniserve does not actually do anything with the images and subsequent scripts may actually require a more narrow set of image types.

svenstaro:
> > Interesting approach and neat use case for miniserve! Do you think there's value in actually verifying that a valid image has been uploaded? You can still upload anything you want even if it's not an image and nothing will actually check that.
> 
> Yeah, miniserve allows to create small utilities like this really easily, thanks for creating it! I almost started to create my own web app for this before I found out about miniserve, this saved me a lot of time.

Glad to hear!

> For my use case there would not be any real benefit of validating the contents - I'm the only user here and don't intend to make this publicly accessible. I also think that validating the file types would be unnecessary overhead in general, since miniserve does not actually do anything with the images and subsequent scripts may actually require a more narrow set of image types.

I'd agree, just wanted to hear your thoughts.

How do you think upload and image upload should interact?

src/args.rs
mayjs: That's neat, thanks for letting me know. I always used argh so far because of the lower compile time impact, maybe I should check out clap.
What do you think about the idea to pass the value for the `accept` attribute instead of making this a flag? Any preference on your side?

svenstaro: I think that'd be kinda neat and a bit more elegant but potentially also a bit more annoying for users. If we do go that way, we need to make sure to add proper documentation to make users aware of the possibilities and implications.

mayjs: Yeah that's true, we'd almost have to provide a table of potentially useful values and their effects if we choose this approach. But I think it may be worth it, maybe you can get a smartphone to open a video recorder or audio recorder with the right values.

svenstaro: Can you try your theory? In that case we should perhaps enable it.

mayjs: Just gave it a try and ran the examples given on w3schools (https://www.w3schools.com/tags/att_input_accept.asp):
* `video/*` does indeed open the camera in video mode instead of stills mode
* `audio/*` opens the audio recorder app

So I guess there would be a benefit in allowing users to set these values as well.

svenstaro: Yeah I think that's pretty cool. Though I wonder how to best expose this and show it to the user especially considering this is most useful on mobile. I suppose perhaps we'd want to actually show multiple buttons on mobile potentially each with their own media specialization? Not quite sure to be honest, the UI might get too crowded. What are your thoughts on this?

mayjs: I think we don't need multiple buttons, you can actually just combine the individiual values. When browsing the site on Firefox on Android, a value like `image/*,audio/*` lets me choose between the audio recorder and camera.

I think that leaves us with two useful options:
1. Allow users to specify the value of the accept attribute directly in its raw form
2. Provide an argument that collects values of a predefined `enum` into a list

Of course, we could also combine both approaches and have option 2 for ease of use and still allow option 1 for special use cases.

PR #755, opened 2022-03-12
Allow to set the accept input attribute to arbitrary values

mayjs:
This supersedes #750.
It still solves #546 but may also prove to be useful in other scenarios by allowing to set more mediatypes.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -8,6 +8,13 @@ use crate::auth;
 use crate::errors::ContextualError;
 use crate::renderer;
 
+#[derive(clap::ArgEnum, Clone)]
+pub enum MediaType {
+    Image,
+    Audio,
+    Video,
+}
+
 #[derive(Parser)]
 #[clap(name = "miniserve", author, about, version)]
 pub struct CliArgs {
@@ -103,6 +110,19 @@ pub struct CliArgs {
     /// Enable file uploading
     #[clap(short = 'u', long = "upload-files")]
     pub file_upload: bool,
+
+    /// Specify uploadable media types
+    #[clap(arg_enum, short = 'm', long = "media-type", requires = "file-upload")]
+    pub media_type: Option<Vec<MediaType>>,
+
+    /// Directly specify the uploadable media type expression
+    #[clap(
+        short = 'M',
+        long = "raw-media-type",
+        requires = "file-upload",
+        conflicts_with = "media-type"
+    )]
+    pub media_type_raw: Option<String>,
 
     /// Enable overriding existing files during file upload
     #[clap(short = 'o', long = "overwrite-files")]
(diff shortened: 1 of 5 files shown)


svenstaro:
Pretty neat and clean! I'd almost be happy to accept it as is but I think this feature is complex enough to require another example line of documentation in the README and also I'd like to see a test for this. The test should be rather simple and you can copy from the existing body of tests.

mayjs:
Thanks for the feedback!

I added some tests and a line for the README.
How / when do you update the `USAGE` section in the README?
Is that something you do when a new release is created or should I adjust that part as well?
I was worried about the version number.

Are the tests fine for you or did you have something else in mind?
Thanks for making me write them, that way I learned about `rstest` and how easily you can do fixture-based testing and parametrized tests using it 🙂.

src/args.rs
```diff
@@ -104,6 +111,19 @@ pub struct CliArgs {
...
+    pub media_type: Option<Vec<MediaType>>,
+
+    /// Directly specify the uploadable media type expression
+    #[clap(
+        short = 'M',
+        long = "raw-media-type",
+        requires = "file-upload",
+        conflicts_with = "media-type"
+    )]
+    pub media_type_raw: Option<String>,
```
svenstaro, line 125: Is this really useful? At the point where you need this level of control, wouldn't it be better to just generally set a header using `--header`?

tests/upload_files.rs
```diff
@@ -167,3 +167,23 @@ fn upload_to_symlink_directory(
...
+#[case(server(&["-u", "-M", "test_value"]), Some("test_value"))]
+fn set_media_type(
+    #[case] server: TestServer,
+    #[case] expected_accept_value: Option<&str>,
+) -> Result<(), Error> {
+    let body = reqwest::blocking::get(server.url())?.error_for_status()?;
+    let parsed = Document::from_read(body)?;
+
+    let input = parsed.find(Attr("id", "file-input")).next().unwrap();
+    assert_eq!(input.attr("accept"), expected_accept_value);
```
svenstaro, line 186: Very nice test!

svenstaro:
Don't worry about updating the usage, I will do it with the next release. :)

svenstaro (changes requested):
Almost there! :)

src/args.rs
mayjs: How would you achieve the same thing using a header?

I think the `raw-media-type` parameter can be useful if you want to specify a filetype or a media type that is not included in the `media-type` parameter.
There are a lot of possibilities for browsers to use this value and offer the right file choosing option to the user, and I think by adding this option we gain a lot of flexibility without adding a lot of complexity.

To give an example, if you wanted to upload a PDF document, you could use `miniserve -u -M application/pdf` and your file chooser would automatically filter for PDF files.
On stock Android, this results in a nice overview of your most recent PDF files all in one place, allowing to quickly find your file.

svenstaro: Nevermind, of course this can't be accomplished using header. I really like your explanation. Merging as is!

PR #760, opened 2022-03-22
docs: add installation via Cask section

axetroy:
install miniserve via [Cask](https://github.com/axetroy/cask.rs)

This is not just recommendations, but I want to use miniserve and simply install it.

NOTE:

Cask does not need to do any additional work, such as create a PR when releasing a new version https://github.com/Homebrew/homebrew-core/pull/97241

Just need the `Cask.toml` file in the root of the repo. that's all.

```bash
# try it out
cask install github.com/axetroy/miniserve
```

截屏2022-03-23 00 59 33

### DIFF
--- a/Cask.toml
+++ b/Cask.toml
@@ -7,4 +7,4 @@ description = """
 """
 
 [darwin]
-x86_64 = "https://github.com/svenstaro/miniserve/releases/download/v{version}/miniserve-v{version}-x86_64-apple-darwin"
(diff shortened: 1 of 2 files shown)


svenstaro:
I'm not convinced Cask is prevalent enough. It's the first time I've heard of it at any rate.

axetroy:
Cask is a young package manager with a small price, in a fixed platform to install binary

Since I added [gitHub.com/svenstaro/miniserve](
https://github.com/cask-pkg/cask-core/blob/main/github.com/svenstaro/miniserve/Cask.toml) to the remote, this PR can be closed now.

Users still can use the following command to install

```bash
cask install gitHub.com/svenstaro/miniserve
```

Issue #173, opened 2019-08-04
"--random-route" leaks the 6-hexdigit token in 404 pages


dsferruzza:
Hi!

I believe that the `--random-route` option is meant to be used as a "soft authentication method", as shown in the [README example](https://github.com/svenstaro/miniserve#generate-random-6-hexdigit-url).
Unfortunately, the generated token seems to appear in 404 pages, which makes the whole feature awkwardly useless...
I am using miniserve `0.5.0` (downloaded from the releases page) on Windows 10.

Here is what I get when doing `miniserve --random-route` and then `curl http://127.0.0.1:8080/`:

```html
<body>
    <!DOCTYPE html>
    <html>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>404 Not Found</title>
    <else>
        <title>Index of 404 Not Found</title>
    </else>
    <style>
        /* CSS code */
    </style>

    </html>
    <div class="error">
        <p>404 Not Found</p>
        <p>Route / could not be found</p>
        <div class="error-nav"><a class="error-back" href="/b0e615">Go back to file listing</a></div>
    </div>
</body>
```
_(HTML code was reformatted to improve readability.)_

1. You can see the link to `/[token]` at the end
2. What's up the HTML markup BTW? It makes little sense...

boastful-squirrel:
Oops, that's a fail. I sadly don't have time to make the fix, but in case someone wants to take it, here are the changes that should be made:

For the random route leak:
1. in `main.rs`, line `309`, the parameter should be set to `false`, that should fix this behaviour for 404 errors.
2. for the errors occurring during upload, the `upload_file` method should take a `uses_random_route` boolean parameter, which should then be given to the `create_error_response` method (add a parameter `display_back_link` and propagate it to the `render_error` method). 

For the wierd html:
1. in `render.rs`, `DOCTYPE` and `<html>` should be extracted from the `page_header` method and be set in the `page` method, around everything
2. in the `page_header` method, it should be `@else` instead of `else` for the title thing

boastful-squirrel:
Good catch by the way

PR #764, opened 2022-03-30
Do not use the route for the favicon and stylesheet

steffhip:
In order to no leak the random generated route, we must not use it as prefix in the 404 error page.
Indeed we can just use a top-level path for these files.


steffhip:
This issue is very similar to https://github.com/svenstaro/miniserve/issues/173.
However the leak occurs at a different part of the 404 page

svenstaro:
Very good find, thanks!

Issue #768, opened 2022-04-09
Create directory feature


sheepy0125:
Hello there!

I've been using Miniserve as a sort of "non-deletable" backup system, and I love it!
However, I'd love for there to be a way to create a directory, and it seems a couple other people would too (#502 and #510).
In #502, [david0u0](https://github.com/david0u0) also addressed this and submitted a pull request (#512). However, they disappeared after a change was requested.

I was wondering if I could try doing this and if such a change would be accepted.

My idea is to have another box similar to the "upload files" one, where one could enter a name for the directory to create. I would also like to enter something like `hello/there` to create a directory `there` inside of another directory `hello`. Lastly, I would like to have an `--enable-mkdir` flag. 

Thanks!

svenstaro:
Concept sounds fine to me, feel free to go ahead as planned. :)

Issue #498, opened 2021-04-18
Add option to restrict symlinks to serve path


svenstaro:
Currently, we either get symlinks or we don't (`-P`). It might be cool to get symlinks but restrict resolving them to only inside the serve path. This would be a safe default.

PR #781, opened 2022-04-23
Create directory

sheepy0125:
Addresses #768.
Directories can be created with the `/upload` endpoint, with a `mkdir` multipart form key.
Attached is a video of it in use.

https://user-images.githubusercontent.com/73724119/164916100-4c537b0f-d305-4686-887d-5c2805f96e63.mp4

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -111,6 +111,10 @@ pub struct CliArgs {
     #[clap(short = 'u', long = "upload-files")]
     pub file_upload: bool,
 
+    /// Enable creating directories
+    #[clap(short = 'U', long = "mkdir", requires = "file-upload")]
+    pub mkdir_enabled: bool,
+
     /// Specify uploadable media types
     #[clap(arg_enum, short = 'm', long = "media-type", requires = "file-upload")]
     pub media_type: Option<Vec<MediaType>>,
--- a/tests/create_directories.rs
+++ b/tests/create_directories.rs
@@ -1,0 +1,122 @@
+mod fixtures;
+
+use fixtures::{server, Error, TestServer};
+use reqwest::blocking::{multipart, Client};
+use rstest::rstest;
+use select::document::Document;
+use select::predicate::{Attr, Text};
+
+/// This should work because the flags for uploading files and creating directories
+/// are set, and the directory name and path are valid.
+#[rstest]
+fn creating_directories_works(
+    #[with(&["--upload-files", "--mkdir"])] server: TestServer,
+) -> Result<(), Error> {
+    let test_directory_name = "hello";
+
+    // Before creating, check whether the directory does not yet exist.
+    let body = reqwest::blocking::get(server.url())?.error_for_status()?;
+    let parsed = Document::from_read(body)?;
+    assert!(parsed.find(Text).all(|x| x.text() != test_directory_name));
+
+    // Perform the actual creation.
+    let create_action = parsed
+        .find(Attr("id", "mkdir"))
+        .next()
+        .expect("Couldn't find element with id=mkdir")
+        .attr("action")
+        .expect("Directory form doesn't have action attribute");
+    let form = multipart::Form::new();
+    let part = multipart::Part::text(test_directory_name);
+    let form = form.part("mkdir", part);
+
+    let client = Client::new();
+    client
+        .post(server.url().join(create_action)?)
+        .multipart(form)
+        .send()?
+        .error_for_status()?;
+
+    // After creating, check whether the directory is now getting listed.
+    let body = reqwest::blocking::get(server.url())?;
+    let parsed = Document::from_read(body)?;
+    assert!(parsed
+        .find(Text)
+        .any(|x| x.text() == test_directory_name.to_owned() + "/"));
+
+    Ok(())
+}
+
+/// This should fail because the server does not allow for creating directories
+/// as the flags for uploading files and creating directories are not set.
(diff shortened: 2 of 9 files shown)


svenstaro:
From a quick look, this look like great work! I'll take a more thorough look later.

sheepy0125:
Sorry about the failed checks, hopefully it should work now...

tests/create_directories.rs
```diff
@@ -0,0 +1,115 @@
+mod fixtures;
+
+use fixtures::{server, Error, TestServer};
+use reqwest::blocking::{multipart, Client};
+use rstest::rstest;
+use select::document::Document;
+use select::predicate::{Attr, Text};
+
+#[rstest]
+fn creating_directories_works(#[with(&["-u", "-U"])] server: TestServer) -> Result<(), Error> {
```
svenstaro, line 10: Use the long flags here.

tests/create_directories.rs
```diff
@@ -0,0 +1,115 @@
...
+    let parsed = Document::from_read(body)?;
+    assert!(parsed
+        .find(Text)
+        .any(|x| x.text() == test_directory_name.to_owned() + "/"));
+
+    Ok(())
+}
+
+#[rstest]
+fn creating_directories_is_prevented(server: TestServer) -> Result<(), Error> {
```
svenstaro, line 47: Add a doc comment here describing _when_ or _why_ it is prevented.

tests/create_directories.rs
```diff
@@ -0,0 +1,115 @@
...
+/// see: https://github.com/svenstaro/miniserve/issues/518
+#[rstest]
+#[case("foo", "bar", "foo/bar")] // Not CWE-22, but `foo` isn't a directory
+#[case("/../foo", "bar", "foo/bar")]
+#[case("/foo", "/../bar", "foo/bar")]
+#[case("C:/foo", "C:/bar", if cfg!(windows) { "foo/bar" } else { "C:/foo/C:/bar" })]
+#[case(r"C:\foo", r"C:\bar", if cfg!(windows) { "foo/bar" } else { r"C:\foo/C:\bar" })]
+#[case(r"\foo", r"\..\bar", if cfg!(windows) { "foo/bar" } else { r"\foo/\..\bar" })]
+fn prevent_path_transversal_attacks(
+    #[with(&["-u", "-U"])] server: TestServer,
```
svenstaro, line 94: Also use the long flags here.

svenstaro (changes requested):
Great work! Just some small things and we're there. :)

svenstaro:
Also make sure to resolve the conflicts/rebase. :)

sheepy0125:
I have no clue how to rebase... I don't think I did it right

svenstaro:
I'm afraid this doesn't look right :) You have two weird merges in there right now. Make sure your local `master` is up to date with upstream. Then, let's try an interactive rebase onto it while at the same time getting rid of the merges in there. I'd try to start something with `git rebase -i master~3`. There are plenty of great interactive tutorials on this.

svenstaro:
While trying this out I noticed something: It's possible to specify `-U` without `-u` which will actually render a button but that button can't be used due to error 405. I think perhaps we should specify `-U` to require `-u`?

sheepy0125:
Hmm... since it is using the `/upload` endpoint, it'd be easier to require `-U` to also have `-u`. I think that's a good idea. What should happen if `-U` is set without `-u`, though?

svenstaro:
> What should happen if `-U` is set without `-u`, though?

Just let [clap deal with it](https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires) and show the user an error.

src/args.rs
```diff
@@ -110,9 +110,13 @@ pub struct CliArgs {
     /// Enable file uploading
     #[clap(short = 'u', long = "upload-files")]
     pub file_upload: bool,
+    
+    /// Enable creating directories
+    #[clap(short = 'U', long = "mkdir", requires("file-upload"))]
+    pub mkdir_enabled: bool,
 
     /// Specify uploadable media types
-    #[clap(arg_enum, short = 'm', long = "media-type", requires = "file-upload")]
```
svenstaro, line 115: Hey, what's the reason for touching this one?

src/args.rs
```diff
@@ -110,9 +110,13 @@ pub struct CliArgs {
     /// Enable file uploading
     #[clap(short = 'u', long = "upload-files")]
     pub file_upload: bool,
+    
+    /// Enable creating directories
+    #[clap(short = 'U', long = "mkdir", requires("file-upload"))]
```
svenstaro, line 115: Can you use the `=` syntax?

svenstaro (changes requested):

src/args.rs
sheepy0125: Sorry, I have no idea why I touched that.

svenstaro:
Ok, one last issue I found with this is that you're currently apparently able to create directories in symlinks which point to directories even when symlinks are disabled. Coincidentally, I also found that it's currently possible to upload files to directories pointed to by symlinks when symlinks are disabled.

Would you like to take a stab at fixing that?

sheepy0125:
That's not good... sure, I'll try!

svenstaro:
Amazing job! Merging as is. Thanks for this great contribution. Since you're now very familiar with this code, especially the symlinking stuff, any chance you could take a look at #498?

PR #805, opened 2022-05-23
Add configuration `change-theme`

flamingoodev:
Add configuration `change-theme` for `Change theme` button.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -84,7 +84,7 @@ pub struct CliArgs {
     pub hidden: bool,
 
     /// Enable change theme display
-    #[clap(short = 'c', long = "change-theme")]
+    #[clap(long = "change-theme")]
     pub change_theme: bool,
 
     /// Default color scheme
--- a/src/args.rs
+++ b/src/args.rs
@@ -83,7 +83,7 @@ pub struct CliArgs {
     #[clap(short = 'H', long = "hidden")]
     pub hidden: bool,
 
-    /// Enable change theme display
+    /// Show theme selector in frontend
     #[clap(long = "change-theme")]
     pub change_theme: bool,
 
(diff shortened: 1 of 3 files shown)


src/args.rs
```diff
@@ -83,6 +83,10 @@ pub struct CliArgs {
     #[clap(short = 'H', long = "hidden")]
     pub hidden: bool,
 
+    /// Enable change theme display
```
svenstaro, line 86: How about
```suggestion
    /// Show theme selector in frontend
```
?

src/config.rs
```diff
@@ -81,6 +81,9 @@ pub struct MiniserveConfig {
     /// Enable QR code display
     pub show_qrcode: bool,
 
+    /// Enable change theme display
```
svenstaro, line 84: How about
```suggestion
    /// Show theme selector in frontend
```
?

svenstaro (changes requested):
Seems fair enough. How about changing the text slightly?

src/args.rs
```diff
@@ -83,6 +83,10 @@ pub struct CliArgs {
...
+    #[clap(short = 'c', long = "change-theme")]
```
svenstaro, line 87: Also this can't work since now you use `-c` twice. Does this need a short option?

svenstaro (changes requested):

src/config.rs
flamingoodev: Sounds good, thanks for the suggestion.

src/args.rs
flamingoodev: Oh,I didn't notice it was used,I'll remove it.

src/args.rs
svenstaro: This still isn't fixed.

svenstaro (changes requested):
You only applied one of the suggestions.

svenstaro:
Thanks!

PR #846, opened 2022-07-20
Some refactoring

cyqsimon:
I saw some rather ugly code so I did some refactoring.

### Things of note:
- Add an empty `rustfmt.toml`
- Switch from `qrcodegen` to `qrcode` crate for QR code support
  - This is so that we don't have to maintain a bespoke (and rather crude TBH) output implementation
- I noticed a clear mistake, but I'm not sure how you want to fix it. See code review

### DIFF
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -8,14 +8,14 @@ use actix_web::web::Query;
 use actix_web::{HttpMessage, HttpRequest, HttpResponse};
 use bytesize::ByteSize;
 use percent_encoding::{percent_decode_str, utf8_percent_encode};
-use qrcodegen::{QrCode, QrCodeEcc};
+use qrcode::QrCode;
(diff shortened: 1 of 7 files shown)


src/listing.rs
```diff
@@ -219,10 +219,10 @@ pub fn directory_listing(
 
     // If the `qrcode` parameter is included in the url, then should respond to the QR code
     if let Some(url) = query_params.qrcode {
-        let res = match QrCode::encode_text(&url, QrCodeEcc::Medium) {
```
cyqsimon, line 222: Looks like someone was in a hurry and overlooked this. `url` is not actually the URL, so we're basically encoding an arbitrary query value.

svenstaro:
Hey, thanks! Is there any chance to split this up into individual PRs? I especially like the QR Code work.

PR #848, opened 2022-07-20
Switch to `fast_qr` lib

cyqsimon:
As mentioned in #846 , there is a clear mistake but I'm unsure how to fix.


cyqsimon:
Seems like VSCode auto-sorted `Cargo.toml`. Looks okay though.

src/listing.rs
```diff
@@ -219,10 +219,10 @@ pub fn directory_listing(
 
     // If the `qrcode` parameter is included in the url, then should respond to the QR code
     if let Some(url) = query_params.qrcode {
-        let res = match QrCode::encode_text(&url, QrCodeEcc::Medium) {
+        let res = match QrCode::with_error_correction_level(url, consts::QR_EC_LEVEL) {
```
cyqsimon, line 222: 👆 `url` is not actually URL.

svenstaro (changes requested):

Cargo.toml
```diff
@@ -1,60 +1,60 @@
 [package]
-name = "miniserve"
```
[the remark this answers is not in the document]
cyqsimon: Sure, done.

svenstaro:
Please rebase, sorry, I updated some deps. Won't update deps anymore until this is merged.

Cargo.toml
```diff
@@ -11,50 +11,50 @@ categories = ["command-line-utilities", "network-programming", "web-programming:
...
 mime = "0.3"
-httparse = "1"
-http = "0.2"
-atty = "0.2"
-rustls = { version = "0.20", optional = true }
-rustls-pemfile = { version = "1.0", optional = true }
+nanoid = "0.4"
+percent-encoding = "2"
+port_check = "0.1"
+qrcode = "0.12.0"
```
svenstaro, line 45: This qrcode library is technically not as well maintained but then again I think QR code libs probably don't need that much maintenance so it should be fine.

svenstaro (changes requested):
Great stuff! But the test stuff needs looking at I think.

src/renderer.rs
```diff
@@ -216,6 +217,45 @@ pub fn raw(entries: Vec<Entry>, is_root: bool) -> Markup {
...
+                            margin: 0;\
+                            display: grid;\
+                            align-items: center;\
+                            justify-items: center;\
+                        }\
+                        svg {\
+                            width: 80%;\
+                            height: 80%;\
+                        }\
+                    "))
```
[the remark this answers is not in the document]
cyqsimon: Done. Please take a look at `style.scss` to see if you approve of the code style and organisation.

src/consts.rs
```diff
@@ -0,0 +1,4 @@
+use qrcode::EcLevel;
+
+/// The error correction level to use for all QR code generation.
+pub const QR_EC_LEVEL: EcLevel = EcLevel::M;
```
aliemjay, line 4: IIRC low error correction level was more suitable for the terminal when I tested it. The generated QR takes less space and reliable enough if we expect it to be scanned by high res camera. It might be true for the web page as well, though I haven't tested.

data/style.scss
```diff
@@ -131,7 +156,7 @@ nav {
     justify-content: flex-end;
 }
 
-nav > div {
+nav>div {
```
cyqsimon, line 159: All this formatting was done automatically by VSCode, all of which I believe is correct from a best-practice POV. Let me know if you are okay with keeping those changes here.

src/consts.rs
cyqsimon: Yeah sure we can use `EcLevel::L` if you believe it's more appropriate; I don't really have an opinion on this. If it comes with the immediate and tangible benefit of reducing screen footprint, why not. If it proves to be too low in the future it's always trivial to change it.

Cargo.toml
cyqsimon: > Can we give it a different writer?

I'll investigate

cyqsimon: Turns out, simply setting `stdin(Stdio::null())` fixes the problem.

cyqsimon:
Also I've noticed an issue in the QR code dropdown menu when run with `-q`. Give me some time to fix it. Meanwhile I'll mark this PR as draft.

data/style.scss
```diff
@@ -9,13 +9,37 @@ $themes: squirrel, archlinux, monokai, zenburn;
...
+    align-items: center;
+    justify-items: center;
+}
+
+html.qr_code_page svg {
+    width: 80%;
+    height: 80%;
+}
+
+// --------------------------------------------------
```
svenstaro, line 42: I believe these indicators are unnecessary.

data/style.scss
```diff
@@ -9,13 +9,37 @@ $themes: squirrel, archlinux, monokai, zenburn;
...
+
     @each $t in $themes {
         $s: selector.append($s, unquote(":not(.theme_#{$t})"));
     }
+
     @return $s;
 }
 
+// make QR code expand and fill page
+// --------------------------------------------------
```
svenstaro, line 21: I believe these indicators are unnecessary. We use them in no other location.

data/style.scss
svenstaro: I suppose it's fine but how about you make another PR which I'll merge before this one where you _only_ reformat the scss? It's going to be cleaner for the history then and will minimize the differences here.

tests/qrcode.rs
```diff
@@ -1,43 +1,101 @@
...
+    assert!(output.contains("QR code for "));
     Ok(())
 }
 
 #[rstest]
-fn show_qrcode_element(#[with(&["-q"])] server: TestServer) -> Result<(), Error> {
-    let body = reqwest::blocking::get(server.url())?.error_for_status()?;
-    let parsed = Document::from_read(body)?;
-    assert!(parsed.find(Attr("id", "qrcode")).next().is_some());
+fn qrcode_hidden_in_non_tty_when_enabled(tmpdir: TempDir, port: u16) -> Result<(), Error> {
```
svenstaro, line 69: This still prints a bunch of QR codes to my actual terminal while running tests. Is this intended? Can we capture the output here and process it without it showing up?

svenstaro (changes requested):
The QR code doesn't actually show up for me in the browser currently:
image

Cargo.toml
```diff
@@ -11,50 +11,50 @@ categories = ["command-line-utilities", "network-programming", "web-programming:
...
 
 [dependencies]
-actix-web = "4"
 actix-files = "0.6"
 actix-multipart = "0.4"
+actix-web = "4"
 actix-web-httpauth = "0.8"
-maud = "0.23"
-yansi = "0.5"
-simplelog = "0.12"
```
svenstaro, line 26: If you want to, you could make another PR to only order these around so that we keep history clean and total changes low.

svenstaro:
Hey @cyqsimon, I'd like to cut a new release of miniserve with this in it after I merge the other PRs soon. Do you think you can finish this one up in the coming days? No problem if it's not in the release, just curious.

Cargo.toml
svenstaro: I ran `cargo sort` on master to minimize the difference here. However, that means you're going to have to rebase once.

cyqsimon:
@svenstaro please go ahead with the release. Life got in the way for me during this past week and I am unable to squeeze out extra time. Sorry about that. I'll try to work on this PR this weekend.

svenstaro:
No worries!

cyqsimon:
Rebased.

data/style.scss
cyqsimon: Removed.

data/style.scss
cyqsimon: Removed.

tests/qrcode.rs
cyqsimon: Fixed. Test wasn't actually effective since I forgot to set `stdout` to `piped` mode.

tests/qrcode.rs
```diff
@@ -1,43 +1,94 @@
...
+    let cmd = {
+        let bin = template.get_program().to_str().expect("not UTF8");
+        let args = template
+            .get_args()
+            .map(|s| s.to_str().expect("not UTF8"))
+            .collect::<Vec<_>>()
+            .join(" ");
+        format!("{} {}", bin, args)
+    };
+    let mut child = bash_command(&cmd).spawn()?;
```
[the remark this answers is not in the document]
cyqsimon: This is fixed now.

Cargo.toml
cyqsimon: Done.

cyqsimon:
Still working on fixing the QR code in the drop-down menu. The previous implementation seems rather jerry-rigged. I'd like to spend some time to do it properly, if I can say so.

Cargo.toml
```diff
@@ -44,7 +44,7 @@ mime = "0.3"
 nanoid = "0.4"
 percent-encoding = "2"
 port_check = "0.1"
-qrcodegen = "1"
+qrcode = "0.12.0"
```
svenstaro, line 47: By the way, there's also [fast_qr](https://github.com/erwanvivien/fast_qr) which might be better maintained and it has fewer dependencies. Maybe worth checking out?

cyqsimon: Sure, I'll try it.

cyqsimon:
Finally I have everything the way I like it. Sorry for the long long wait.

Previously the QR code in the webpage works by having the client send a request to the server with the `qrcode` param set to the URL, which was very ugly IMO. Now the QR code is generated at once with the rest of the page. The URL is worked out entirely using the info from the request.

Also thanks for the `fast_qr` recommendation. It is indeed quite a bit easier to work with than the `qrcode` crate. In particular, the ability to set the margin size when building the SVG is very nice. There is however one downside: `fast_qr` is much more restrictive than `qrcode` when it comes to generating ASCII art. There is no customising the characters used for background/foreground, which makes the previously supported feature of printing an inverted QR code side-by-side basically impossible. That being said, I don't think it's such a deal-breaker, so altogether the benefits outweigh the downsides.

P.s. I did notice some weird artefacting in VScode's integrated terminal. It seems like the background colour was not reset properly by `fast_qr`:
flameshot-20220901-203755
... but it's not present in my other terminal emulators (Alacritty, Konsole, Cool Retro Terminal). Please help me try it out on other terminal emulators if possible.

cyqsimon:
And finally I think I need some help on writing a test for the webpage QR code. If I understand correctly, the `fixtures` crate should be used for this test, but I'm not familiar with the crate at all.

tests/qrcode.rs
```diff
@@ -1,46 +1,84 @@
...
 
+    assert!(!output.contains("QR code for "));
     Ok(())
 }
 
 #[rstest]
-fn show_qrcode_element(#[with(&["-q"])] server: TestServer) -> Result<(), Error> {
-    let body = reqwest::blocking::get(server.url())?.error_for_status()?;
-    let parsed = Document::from_read(body)?;
-    assert!(parsed.find(Attr("id", "qrcode")).next().is_some());
```
svenstaro, line 23: You deleted these two tests but I think they are actually quite helpful in making sure that
1) The QR code is not shown in the web page when not requested and
2) The QR code _is_ shown in the web page when desired.

I should be pretty straight forward to put these tests back in, I think.

cyqsimon: Yikes. I can't remember why I deleted these. I guess I didn't really understand their point at the time. Restored now.

svenstaro:
Thanks, this is great!

Issue #842, opened 2022-07-14
set upload folder


UltraBlackLinux:
Hey there,
I would love to have a feature that allowed me to use the -u flag with a folder, so only that folder can be uploaded to.
Can you add this?
Thanks!

svenstaro:
So something like `--restrict-upload-dir` and then trying to upload into any other directory would give an error? Sounds like a neat little feature. Would you like to take a stab at this?

UltraBlackLinux:
> So something like `--restrict-upload-dir` and then trying to upload into any other directory would give an error? Sounds like a neat little feature. Would you like to take a stab at this?

Maybe not even an error (At least not in the web UI). More that the Upload Button only appears in the predifined directory(s) (A list would definitely be useful)

Sadly I am 100% unfamiliar with rust, and this does probably require more knowledge than just copy-pasting stuff around. Someone else will have to do this.

jonasdiemer:
Hi,

looking for an opportunity for some first Rust experience, I've taken a stab: https://github.com/jonasdiemer/miniserve/tree/restrict-upload-dir

Please consider it a rough prototype - esp. the area in [file_upload.rs](https://github.com/jonasdiemer/miniserve/blob/9e358d7d2534197e5b6170626d5b8f2add76705f/src/file_upload.rs#L176) doesn't feel very idiomatic... 

@svenstaro, I'd be happy if you could take a look and suggest what to add/change/improve. Thanks!

svenstaro:
Make a draft PR and I'll take a look.

PR #858, opened 2022-08-02
Added option restrict-upload-dir

jonasdiemer:
To fix https://github.com/svenstaro/miniserve/issues/842


svenstaro:
Cool, before I do a proper review, could you add a test to check that other directories are now indeed forbidden?

jonasdiemer:
Sure, let me have a look into testing in Rust :)

svenstaro:
You can take a look at the other tests concerning the upload. It should be somewhat straight forward. Just remember: If you messed up and the test doesn't pick it up, this is a security issue so we need to be really careful here. :)

jonasdiemer:
@svenstaro thanks for the suggestion. Fixed an issue (running tests is actually helpful :)) and added a test similar to existing one `uploading_files_is_prevented`.

svenstaro (changes requested):
Good stuff! This will probably require a few rounds of reviews until we get there but i think for your first foray into Rust this is already looking pretty good.

tests/upload_files.rs
```diff
@@ -80,6 +80,44 @@ fn uploading_files_is_prevented(server: TestServer) -> Result<(), Error> {
...
+    let client = Client::new();
+    // Ensure uploading fails and returns an error
+    assert!(client
+        .post(server.url().join("/upload?path=/")?)
+        .multipart(form)
+        .send()?
+        .error_for_status()
+        .is_err());
+
+    // After uploading, check whether the uploaded file is now getting listed.
```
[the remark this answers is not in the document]
jonasdiemer: Copy&paste error at the place where I copied from, fixed.

tests/upload_files.rs
```diff
@@ -80,6 +80,44 @@ fn uploading_files_is_prevented(server: TestServer) -> Result<(), Error> {
...
+    let body = reqwest::blocking::get(server.url())?;
+    let parsed = Document::from_read(body)?;
+    assert!(!parsed.find(Text).any(|x| x.text() == test_file_name));
+
```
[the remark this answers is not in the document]
jonasdiemer: Added another test.

src/file_upload.rs
```diff
@@ -171,6 +171,19 @@ pub async fn upload_file(
         ContextualError::IoError("Failed to resolve path served by miniserve".to_string(), e)
     })?;
 
+
+    // Disallow paths outside of restricted directories
+    // TODO: Probably not the most rust-ic style...
+    if !conf.restrict_upload_dir.is_empty() {
+        let upl_path = upload_path.clone().into_os_string().into_string().unwrap();
```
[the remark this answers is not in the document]
jonasdiemer: Fully agree to how it reads.

One option I am seeing is to use PathBuf for the --restrict-upload-dir argument instead of String and then handle it as a path. @svenstaro, would this make sense?

src/args.rs
```diff
@@ -111,6 +111,10 @@ pub struct CliArgs {
     #[clap(short = 'u', long = "upload-files")]
     pub file_upload: bool,
 
+    /// Restrict upload directories
+    #[clap(long = "restrict-upload-dir")]
```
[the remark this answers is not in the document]
jonasdiemer: Thanks, makes total sense, done.

jonasdiemer:
Reworked the solution to use PathBuf and with this also fixed subdirectories not working.

@svenstaro I'd appreciate some feedback (esp. on improving the style). Thanks again! :)

src/file_upload.rs
```diff
@@ -171,6 +171,18 @@ pub async fn upload_file(
         ContextualError::IoError("Failed to resolve path served by miniserve".to_string(), e)
     })?;
 
+
+    // Disallow paths outside of restricted directories
+    // TODO: Probably not the most rust-ic style...
+    let upload_allowed = conf.restrict_upload_dir.is_empty() || 
+        conf.restrict_upload_dir.iter().any(|s| upload_path.starts_with(s)); 
+
+    if !(upload_allowed) {
```
svenstaro, line 180:
```suggestion
    if !upload_allowed {
```

src/file_upload.rs
```diff
@@ -171,6 +171,18 @@ pub async fn upload_file(
         ContextualError::IoError("Failed to resolve path served by miniserve".to_string(), e)
     })?;
 
+
+    // Disallow paths outside of restricted directories
+    // TODO: Probably not the most rust-ic style...
+    let upload_allowed = conf.restrict_upload_dir.is_empty() ||
```
svenstaro, line 177:
```suggestion
    // Disallow paths outside of restricted directories
    let upload_allowed = conf.restrict_upload_dir.is_empty() || 
```

src/args.rs
```diff
@@ -111,6 +111,10 @@ pub struct CliArgs {
     #[clap(short = 'u', long = "upload-files")]
     pub file_upload: bool,
 
+    /// Restrict upload directories
+    #[clap(long = "restrict-upload-dir", requires = "file-upload", value_hint = ValueHint::FilePath)]
+    pub restrict_upload_dir: Vec<PathBuf>,
```
svenstaro, line 116:
```suggestion
    /// Allowed upload directories
    ///
    /// If this is set, uploads are only allowed into the provided directories. 
    #[clap(long, requires = "file-upload", value_hint = ValueHint::FilePath)]
    pub allowed_upload_dir: Vec<PathBuf>,
```

svenstaro: I believe this explanation and naming makes more sense. Please change it like that throughout your code unless you have a better suggestion for the naming.

svenstaro (changes requested):

tests/upload_files.rs
```diff
@@ -72,14 +72,96 @@ fn uploading_files_is_prevented(server: TestServer) -> Result<(), Error> {
...
+    // After uploading, check whether the uploaded file is NOT getting listed.
     let body = reqwest::blocking::get(server.url())?;
     let parsed = Document::from_read(body)?;
     assert!(!parsed.find(Text).any(|x| x.text() == test_file_name));
 
     Ok(())
 }
 
+// This test runs the server with --restrict-upload-dir argument and 
+// checks that file upload to a different directory is actually prevented.
```
svenstaro, line 84: Prefer docstrings `///` on top of functions instead of normal comments `//`.

tests/upload_files.rs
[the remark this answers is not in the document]
svenstaro: I think perhaps we could merge the tests using another rstest parameter (success/failure)? We do this in some other places in the tests as well. It always depends on much the code would diverge. Can you check how that would look and whether it makes sense in this case?

src/args.rs
jonasdiemer: Yes, this is clearer.

tests/upload_files.rs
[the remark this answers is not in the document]
jonasdiemer: There is not that much duplication in the two tests - I have removed some redundant parts in the `uploading_files_is_restricted` (eg. checking if file exists). So in my view merging would make the tests significantly less easy to follow.

svenstaro:
I'd be happy to merge as-is. Please update the `README.md` with the new output of `miniserve --help` and fix the formatting issue and we're good to merge.

src/config.rs
```diff
@@ -87,6 +87,9 @@ pub struct MiniserveConfig {
     /// Enable file upload
     pub file_upload: bool,
 
+    /// Restrict file upload dirs
```
svenstaro, line 90:
```suggestion
    /// List of allowed upload directories
```

src/file_upload.rs
```diff
@@ -171,6 +171,17 @@ pub async fn upload_file(
         ContextualError::IoError("Failed to resolve path served by miniserve".to_string(), e)
     })?;
 
+
+    // Disallow paths outside of restricted directories
+    let upload_allowed = conf.allowed_upload_dir.is_empty() || 
+        conf.allowed_upload_dir.iter().any(|s| upload_path.starts_with(s));
```
svenstaro, line 177:
```suggestion
    // Disallow paths outside of allowed directories
    let upload_allowed = conf.allowed_upload_dir.is_empty() || 
        conf.allowed_upload_dir.iter().any(|s| upload_path.starts_with(s)); 
```

src/args.rs
```diff
@@ -111,6 +111,12 @@ pub struct CliArgs {
     #[clap(short = 'u', long = "upload-files")]
     pub file_upload: bool,
 
+    /// Allowed upload directories (together with -u)
+    /// 
+    /// If this is set, uploads are only allowed into the provided directories. 
+    #[clap(long, requires = "file-upload", value_hint = ValueHint::FilePath)]
```
svenstaro, line 117: If you like, you can choose a short letter for this as well as it'd be annoying to write it out many times otherwise.

tests/upload_files.rs
```diff
@@ -72,14 +72,96 @@ fn uploading_files_is_prevented(server: TestServer) -> Result<(), Error> {
...
+    let parsed = Document::from_read(body)?;
+    assert!(!parsed.find(Text).any(|x| x.text() == test_file_name));
+
+    Ok(())
+}
+
+/// This tests that we can upload files to the directory specified by --allow-upload-dir
+#[rstest]
+#[case(server(&["-u",  "--allowed-upload-dir", "someDir"]), "someDir")]
+#[case(server(&["-u",  "--allowed-upload-dir", "someDir/some_sub_dir"]), "someDir/some_sub_dir")]
```
svenstaro, line 120: Actually could you add a test case that shows that it also works when multiple `--allowed-upload-dir` are provided?

svenstaro (changes requested):
Found a few tiny things but once we have that, we're good to go. Great job so far!

tests/upload_files.rs
```diff
@@ -72,14 +72,96 @@ fn uploading_files_is_prevented(server: TestServer) -> Result<(), Error> {
...
+fn uploading_files_to_allowed_dir_works(
+    #[case] server: TestServer,
+    #[case] upload_dir: &str,
+) -> Result<(), Error> {
+    let test_file_name = "uploaded test file.txt";
+
+    // Create test directory
+    use std::fs::create_dir_all;
```
svenstaro, line 128: We already have this in another place in this file. Just hoist this import up to global scope and remove it in the other place as well.

src/args.rs
jonasdiemer: I'd go for 'A' (as in allowed...), but I am wondering if we might just want to re-purpose `-u`...

svenstaro: You mean something like `-u` alone would allow for uploading everywhere and then `-u /tmp/lol` would allow uploading only to that dir? If that's what you're thinking, I appreciate the conciseness but I can't help but wonder whether that wouldn't make `-u` pretty hard to parse (is it even still possible?) and somewhat confusing to users perhaps. Thoughts?

tests/upload_files.rs
```diff
@@ -72,14 +72,96 @@ fn uploading_files_is_prevented(server: TestServer) -> Result<(), Error> {
...
+    let form = form.part("file_to_upload", part);
+
+    let client = Client::new();
+    // Ensure uploading fails and returns an error
+    assert!(client
+        .post(server.url().join("/upload?path=/")?)
+        .multipart(form)
+        .send()?
+        .error_for_status()
+        .is_err());
```
svenstaro, line 107: I was just thinking, perhaps check that this is actually forbidden. Otherwise, this test might fail for a 404 and we wouldn't catch it. Maybe also check for the precise error message?

src/args.rs
svenstaro: `-A` is a fine choice, I think.

svenstaro (changes requested):
Hey, I think we're almost there now. Only some tiny stuff now. I'd like to put this into the next miniserve release, too! Do you have some time to finish it up soon so we can land it with the release?

src/args.rs
jonasdiemer: > You mean something like `-u` alone would allow for uploading everywhere and then `-u /tmp/lol` would allow uploading only to that dir? If that's what you're thinking, I appreciate the conciseness but I can't help but wonder whether that wouldn't make `-u` pretty hard to parse (is it even still possible?) and somewhat confusing to users perhaps. Thoughts?

Exactly, that was my thinking (after seeing so many command line shorts already taken ;)). Not sure about the parsing, though, I'd give it a try...

jonasdiemer: Played around a bit, it looks like it would be relatively easy to parse by using `min_values = 0`:

```rust
    #[clap(short = 'u', long,
     value_hint = ValueHint::FilePath, 
     min_values = 0)]
    pub upload-files: Option<Vec<PathBuf>>,
```

This would yield `None` in case the argument is missing, `Some([])` in case the argument is used without parameters and `Some(['some_dir'])` in case a directory is supplied. So doesn't look to complicated in my view.

@svenstaro what do you think? I would remodel the code to have `-u, --upload-files [<ALLOWED_UPLOAD_DIR>...]` and save the extra argument `-A/--allowed_upload-dir`

svenstaro: Does it parse this correctly? `-u --some-other-flag`? Would it try to parse the stuff behind `-u` as a `PathBuf`?

jonasdiemer: Just tried `-u --verbose` and it worked as expected (i.e. verbose output and `allowed_upload_dir == Some([])` (i.e. it would allow uploads everywhere)

tests/upload_files.rs
jonasdiemer: Done.

tests/upload_files.rs
jonasdiemer: Yes, that makes more sense.

src/args.rs
jonasdiemer: I've just committed a version that uses `-u` to define allowed directories. @svenstaro, please check it out and let me know what you think (I am also fine to use the previous version with `-A` as a shortcut).

jonasdiemer:
> Hey, I think we're almost there now. Only some tiny stuff now. I'd like to put this into the next miniserve release, too! Do you have some time to finish it up soon so we can land it with the release?

Except for the decision for using -u or -A (and some cleanup in case we go with -u), it should be ready.

src/args.rs
Atreyagaurav: I like the `-u` with `min_values`. Makes more sense. Might want to try how it does in windows shell. It should work in bash because we've to use `./-path` when a file starts with `-` (e.g. `touch ./-h`) otherwise it'll be parsed as cli argument.

jonasdiemer: Unfortunately, I cannot try on Windows.

Directories starting with `-` would also need the `./` prefix, which doesn't work with my implementation (yet) - will have a look.

svenstaro:
Please run `cargo fmt` to fix CI.

jonasdiemer:
Fixed formatting and also sanitized paths so that specification of `-u ./some_dir' also works.

svenstaro:
There's a small lint to fix there.

jonasdiemer:
Looks like CICD on Windows fails, might have to do with the tests that include a dash in the directory name. Unfortunately, I don't have a windows machine to test on...

Atreyagaurav:
I also don't have a Windows machine to build miniserve to test, but I made a colleague try to make a directory named "-h" and it was a success (I tried both with "New Folder" button and `mkdir` in cmd. `cd -h` & `cd ./-h`worked fine, but `dir ./-h` gave `Invalid switch - "h".`

I don't know if it's a good idea, but we might be able to use `cfg` to only test that on Linux? At least for now...

svenstaro:
I'd be fine if we don't run that test on Windows.

jonasdiemer:
>

tests/upload_files.rs
```diff
@@ -117,23 +117,27 @@ fn uploading_files_is_restricted(#[case] server: TestServer) -> Result<(), Error
...
-#[case(server(&["-u", "someDir/some_sub_dir"]), vec!["someDir/some_sub_dir"])]
+#[case(server(&["-u", "someDir"]), vec!["someDir".to_string()])]
+#[case(server(&["-u", "./-someDir"]), vec!["./-someDir".to_string()])]
+#[case(server(&["-u", "someDir/some_sub_dir"]), vec!["someDir/some_sub_dir".to_string()])]
 #[case(server(&["-u", "someDir/some_sub_dir", "-u", "someDir/some_other_dir"]), 
-       vec!["someDir/some_sub_dir", "someDir/some_other_dir"])]
+       vec!["someDir/some_sub_dir".to_string(), "someDir/some_other_dir".to_string()])]
 fn uploading_files_to_allowed_dir_works(
     #[case] server: TestServer,
-    #[case] upload_dirs: Vec<&str>,
```
svenstaro: Does it work if you take these as `Path` or `OsPath`?

tests/upload_files.rs
```diff
@@ -72,14 +73,103 @@ fn uploading_files_is_prevented(server: TestServer) -> Result<(), Error> {
...
+
+    let client = Client::new();
+    // Ensure uploading fails and returns an error
+    assert_eq!(
+        500,
+        client
+            .post(server.url().join("/upload?path=/")?)
+            .multipart(form)
+            .send()?
+            .status()
```
svenstaro, line 107: I just found this. We should never return 500s. This is a clear case of FORBIDDEN.

jonasdiemer: Yes, fully agree. However, that's a behavior that I've inherited from how miniserve treats other errors (e.g. uploading a file again without `overwrite_files` option), so I would think this should be fixed in a separate PR...

svenstaro: That's a fair argument to make. Would you be interested in making such a follow-up?

svenstaro:
It seems like there's a conflict with the recent QR stuff merge. Could you rebase?

jonasdiemer:
Rebase done, still need to fix the failing test for windows platform.

tests/upload_files.rs
```diff
@@ -72,14 +73,101 @@ fn uploading_files_is_prevented(server: TestServer) -> Result<(), Error> {
...
+#[case(server(&["-u", "someDir"]), vec!["someDir".to_string()])]
+#[case(server(&["-u", "./-someDir"]), vec!["./-someDir".to_string()])]
+#[case(server(&["-u", if cfg!(windows) {r"someDir\some_sub_dir"} else {"someDir/some_sub_dir"}]),
+  vec!["someDir/some_sub_dir".to_string()])]
+#[case(server(&["-u", if cfg!(windows) {r"someDir\some_sub_dir"} else {"someDir/some_sub_dir"}, 
+                "-u", if cfg!(windows) {r"someDir\some_other_dir"} else {"someDir/some_other_dir"}]), 
+       vec!["someDir/some_sub_dir".to_string(), "someDir/some_other_dir".to_string()])]
+fn uploading_files_to_allowed_dir_works(
+    #[case] server: TestServer,
+    #[case] upload_dirs: Vec<String>,
```
svenstaro, line 129: I still don't really understand why you can't use a `std::path::Path` here as it will take care of the conversion for you on the operating systems. Wouldn't that be more convenient?

svenstaro: We should probably change other tests as well. :)

svenstaro: Since it still doesn't work, I'd also be ok with you simply disabling the failing test cases on Windows.

jonasdiemer: Yup, and kind of hard to make progress without access to a windows machine. Not sure I'll find time to setup something (e.g. on Azure/CodeSpaces)... 

@svenstaro, I don't know how I would disable a test-case for a specific platform. Could you give me a pointer?

jonasdiemer: Also noticed, that the feature doesn't work on windows with subdirectories (likely due to `\` vs `/`) - found out by running miniserve.exe from publish_release action on an old windows laptop. Pushed a fix, waiting for Action workflow to verify.

jonasdiemer: Looks like I found a solution - latest commit passes CI on all platforms.

I assume the warning (the following packages contain code that will be rejected by a future version of Rust: winapi v0.2.8) has nothing to do with my PR.

From my perspective, the PR should be ready to merge now - please let me know if I overlooked something.

Re the 500 errors, I might have a stab at that if time permits.

jonasdiemer: Well, the error handling was easier to fix than I thought - fixed it (also for the case when `overwrite_files` is not set, i.e. `DuplicateFileError`).

svenstaro:
At last! Outstanding work getting it to this state. It took a while but I hope you appreciate that for security stuff we need to go the extra mile. :)

jonasdiemer:
Thanks a lot for your guidance (and patience), I leaned a lot and it was fun!

Issue #859, opened 2022-08-03
[Feature Request + Pull Req] Render Readme if present in current Directory


Atreyagaurav:
Hello, I really like this, and I think we can add the ability to render readme as well, like in git repo. I think it'd be nice to have it while sharing nested directories. 

I'm surprised noone did it before, and I thought it will be simple enough to implement if we have a module for conversion of markdown to html already. So, I've tried to implement it myself in my fork. I tried to use `markdown` crate, and it seems to work for simple markdown, but not for the embedded htmls (like in miniserve's readme). 

I've just started to learn rust so please let me know of any suggestions and improvements on this.

PR #860, opened 2022-08-03
Add support for readme rendering

Atreyagaurav:
Fixes https://github.com/svenstaro/miniserve/issues/859

Suggestions are welcome, I'm only implemented the basic render for now.

Other things that might be needed:
- user flag to turn on/off readme rendering
- render html bits inside markdown (might to be fixed from `markdown` crate)
- other formats besides `.md` (`.txt` and , maybe `.org`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ thiserror = "1"
 yansi = "0.5"
 zip = { version = "0.6.2", default-features = false }
 get_if_addrs = "0.5"
-markdown = "0.3.0"
+comrak = "0.14.0"
 
 [features]
 default = ["tls"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,8 @@ tar = "0.4"
 thiserror = "1"
 yansi = "0.5"
 zip = { version = "0.6.2", default-features = false }
+get_if_addrs = "0.5"
+markdown = "0.3.0"
 
 [features]
 default = ["tls"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,6 @@ tar = "0.4"
 thiserror = "1"
 yansi = "0.5"
 zip = { version = "0.6.2", default-features = false }
-get_if_addrs = "0.5"
 comrak = "0.14.0"
 
 [features]
--- a/src/args.rs
+++ b/src/args.rs
@@ -197,6 +197,10 @@ pub struct CliArgs {
     #[cfg(feature = "tls")]
     #[clap(long = "tls-key", requires = "tls-cert", value_hint = ValueHint::FilePath)]
     pub tls_key: Option<PathBuf>,
+
+    /// Enable readme redering in directories
+    #[clap(long = "readme")]
+    pub readme: bool,
 }
 
 /// Checks wether an interface is valid, i.e. it can be parsed into an IP address
--- a/src/args.rs
+++ b/src/args.rs
@@ -198,7 +198,7 @@ pub struct CliArgs {
     #[clap(long = "tls-key", requires = "tls-cert", value_hint = ValueHint::FilePath)]
     pub tls_key: Option<PathBuf>,
 
-    /// Enable README.md redering in directories
+    /// Enable README.md rendering in directories
     #[clap(long)]
     pub readme: bool,
 }
--- a/src/args.rs
+++ b/src/args.rs
@@ -198,8 +198,8 @@ pub struct CliArgs {
     #[clap(long = "tls-key", requires = "tls-cert", value_hint = ValueHint::FilePath)]
     pub tls_key: Option<PathBuf>,
 
-    /// Enable readme redering in directories
-    #[clap(long = "readme")]
+    /// Enable README.md redering in directories
+    #[clap(long)]
     pub readme: bool,
 }
 
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -232,7 +232,7 @@ pub fn directory_listing(
     }
 
     let mut entries: Vec<Entry> = Vec::new();
-    let mut readme: Option<String> = None;
+    let mut readme: Option<PathBuf> = None;
 
     for entry in dir.path.read_dir()? {
         if dir.is_visible(&entry) || conf.show_hidden {
@@ -285,7 +285,11 @@ pub fn directory_listing(
                     ));
 		    // TODO: Pattern match?
 		    if conf.readme && file_name.to_lowercase() == "readme.md"{
-			readme = Some(file_name);
+			let file_path = conf.path.canonicalize().unwrap()
+			    .join(base.as_os_str().to_str().unwrap()
+				  .strip_prefix("/").unwrap())
+			    .join(&file_name);
+			readme = Some(file_path);
 		    }
                 }
             } else {
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -283,14 +283,22 @@ pub fn directory_listing(
                         last_modification_date,
                         symlink_dest,
                     ));
-		    // TODO: Pattern match?
-		    if conf.readme && file_name.to_lowercase() == "readme.md"{
-			let file_path = conf.path.canonicalize().unwrap()
-			    .join(base.as_os_str().to_str().unwrap()
-				  .strip_prefix("/").unwrap())
-			    .join(&file_name);
-			readme = Some(file_path);
-		    }
+                    // TODO: Pattern match?
+                    if conf.readme && file_name.to_lowercase() == "readme.md" {
+                        let file_path = conf
+                            .path
+                            .canonicalize()
+                            .unwrap()
+                            .join(
+                                base.as_os_str()
+                                    .to_str()
+                                    .unwrap()
+                                    .strip_prefix('/')
+                                    .unwrap(),
+                            )
+                            .join(&file_name);
+                        readme = Some(file_path);
+                    }
                 }
             } else {
                 continue;
@@ -381,7 +389,7 @@ pub fn directory_listing(
             HttpResponse::Ok().content_type(mime::TEXT_HTML_UTF_8).body(
                 renderer::page(
                     entries,
-		    readme,
+                    readme,
                     is_root,
                     query_params,
                     breadcrumbs,
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -232,6 +232,7 @@ pub fn directory_listing(
     }
 
     let mut entries: Vec<Entry> = Vec::new();
+    let mut readme: Option<String> = None;
 
     for entry in dir.path.read_dir()? {
         if dir.is_visible(&entry) || conf.show_hidden {
@@ -275,13 +276,17 @@ pub fn directory_listing(
                     ));
                 } else if metadata.is_file() {
                     entries.push(Entry::new(
-                        file_name,
+                        file_name.clone(),
                         EntryType::File,
                         file_url,
                         Some(ByteSize::b(metadata.len())),
                         last_modification_date,
                         symlink_dest,
                     ));
+		    // TODO: Pattern match, or user arg for readme name?
+		    if file_name.to_lowercase() == "readme.md"{
+			readme = Some(file_name);
+		    }
                 }
             } else {
                 continue;
@@ -372,6 +377,7 @@ pub fn directory_listing(
             HttpResponse::Ok().content_type(mime::TEXT_HTML_UTF_8).body(
                 renderer::page(
                     entries,
+		    readme,
                     is_root,
                     query_params,
                     breadcrumbs,
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -283,8 +283,8 @@ pub fn directory_listing(
                         last_modification_date,
                         symlink_dest,
                     ));
-		    // TODO: Pattern match, or user arg for readme name?
-		    if file_name.to_lowercase() == "readme.md"{
+		    // TODO: Pattern match?
+		    if conf.readme && file_name.to_lowercase() == "readme.md"{
 			readme = Some(file_name);
 		    }
                 }
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -7,6 +7,7 @@ use actix_web::dev::ServiceResponse;
 use actix_web::web::Query;
 use actix_web::{HttpMessage, HttpRequest, HttpResponse};
 use bytesize::ByteSize;
+use comrak::{markdown_to_html, ComrakOptions};
 use percent_encoding::{percent_decode_str, utf8_percent_encode};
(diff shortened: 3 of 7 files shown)


Cargo.toml
```diff
@@ -55,6 +55,7 @@ rustls = { version = "0.20", optional = true }
 rustls-pemfile = { version = "1.0", optional = true }
 socket2 = "0.4"
 get_if_addrs = "0.5"
+markdown = "0.3.0"
```
svenstaro, line 58: Perhaps [comrak](https://crates.io/crates/comrak) would be more suitable for the fancier markdowns that people use nowadays?

svenstaro (changes requested):
I like the idea in general but it should fit into the general current design of miniserve. We already supported an `--index` file and perhaps this could be treated as an index file that still needs to be rendered? What do you think?

svenstaro:
Btw quickest review ever.

svenstaro:
You can mark the PR as draft, then you don't need a `[WIP]` prefix.

Atreyagaurav:
@svenstaro Thanks I didn't know draft was public. 

As for index.html, html writing is complicated compared to markdown, hence I didn't check the index file.

Do we have to write the index file with all the paths ourselves, or just the other parts and the file/directories will be the same? If we can only use the readme part and keep the other contents same as default we can add a bool flag to say the index file is a markdown.

I think we can just add a separate add markdown render without less effort (relatively) and it'll be consistent design with how we share codes in other mediums I think it's worth looking into. Instead of generating a whole HTML. (Of course I'm talking here without knowing how index files work)

And we can improve it further. I am just done implementing a flag so it won't be active by default. And I'll look into better renderer.

svenstaro:
> As for index.html, html writing is complicated compared to markdown, hence I didn't check the index file.

I'm not saying we should actually write any additional HTML here. In fact, we most definitely shouldn't create any files and suprise users.
 
> Do we have to write the index file with all the paths ourselves, or just the other parts and the file/directories will be the same? If we can only use the readme part and keep the other contents same as default we can add a bool flag to say the index file is a markdown.

Sorry, I do not understand this at all. Why would we _write_ any additional HTML files? I meant earlier that we should try to combine the `--index` feature and the markdown rendering to create some kind of combined convergent solution here for miniserve as opposed to creating a divergent one (where we end up with two things that kind of do the same but use different flags and semantics).

> I think we can just add a separate add markdown render without less effort (relatively) and it'll be consistent design with how we share codes in other mediums I think it's worth looking into. Instead of generating a whole HTML. (Of course I'm talking here without knowing how index files work)
> 
> And we can improve it further. I am just done implementing a flag so it won't be active by default. And I'll look into better renderer.

Atreyagaurav:
If you think we need to integrate it with index flag then I'll look into how index is rendered and trace it. It'll take time, I'm not that good with html. 

I'll add some logics to make sure when `--readme` flag is there `--index` should also be given/ or defaults to readme.md. 

I still have some changes that I need to push like actually implementing the `--readme` flag, so I'll commit those and if I'm satisfied I'll change this from draft.

Cargo.toml
Atreyagaurav: Yeah, I think this is better than the one I was using. I'll switch it.

Atreyagaurav:
> >   Do we have to write the index file with all the paths ourselves, or just the other parts and the file/directories will be the same? If we can only use the readme part and keep the other contents same as default we can add a bool flag to say the index file is a markdown.

> Sorry, I do not understand this at all. Why would we write any additional HTML files? I meant earlier that we should try to combine the --index feature and the markdown rendering to create some kind of combined convergent solution here for miniserve as opposed to creating a divergent one (where we end up with two things that kind of do the same but use different flags and semantics).

I meant does the user need to write the entirety of the contents on the index file or will the directory listing part will be automatically generated? If it's the later then we can add readme with the same method, but if user has to write the whole index file, then I think it's better to separate that with the readme functionality where we only add the readme→html part to our template. 

I looked into the index file, and looks like it completely replaces the directory listing, and if we go with using `readme` as a index file that needs to be converted then we'll have to add the directory listing to it using the same method anyway (i.e. generate html for directory listing, and then generate html from readme).

So I think it's better to keep them separate after all. It also matches the file's themes, index file is a way to index directory, and readme is just extra information about that directory. And README can appear sparsely, some directories might have it some might not. And irrespective of its presence, the directory listing will always be there.

Something like this:
```
A
├── B
│   └── README.md
├── C
└── D
    └── README.md
```
So even if there is no `README` in root, if you are visiting `A/B` then the `README` will be rendered.

if you provide index-file: it'll be used and directory listing will be disabled (whatever current default is)
if you have readme flag: default directory listing will have readme contents if there is a file named `readme.md` in the currently visiting directory. It'll be ignored if you have index file.

Current features of readme flag:
- Renders README.md after the directory listing
- Renders README.md if it's present in any nested path -- that you're currently visiting.

Cargo.toml
Atreyagaurav: I changed it. Thought comrak seems to be a bit heavy while markdown was lightweight. It does support the html bits and has the same look as github one so I think it's ok.

Atreyagaurav:
@svenstaro  Any thoughts? My feeling as I said before is readme feature added into current render is ok as index file is supposed to be entry point and readme isn't. Those two seems like different things to me.

svenstaro:
Ok, I thought about this. Sorry, took some time. I want to make sure miniserve doesn't become a hodgepodge of features that don't really make sense. 

I like the feature the way you proposed it and I agree it should be a completely separate mechanism from how the index is rendered. A few points:

- Should it conflict with index? If index is also provided, what should happen? Should index always prevail? I suppose it will, because it will also prevail over simple directory listing already but I'd like to align with you here.
- You write "Renders README.md if it's present in any nested path -- that you're currently visiting.". I'm not sure that makes sense or if understand it correctly. I'd expect a README to only be picked up in the _current_ directory that we're in but not in any subpaths of our currently visited directory, that would cause great confusion I think (and it would also be expensive to find all READMEs potentially).

src/args.rs
```diff
@@ -197,6 +197,10 @@ pub struct CliArgs {
     #[cfg(feature = "tls")]
     #[clap(long = "tls-key", requires = "tls-cert", value_hint = ValueHint::FilePath)]
     pub tls_key: Option<PathBuf>,
+
+    /// Enable readme redering in directories
+    #[clap(long = "readme")]
+    pub readme: bool,
```
svenstaro, line 203: Please also generate an updated `README.md` for the repo and add this to the features list as it's a new cool feature.

Atreyagaurav: Ok. I'll do that.

Atreyagaurav:
> Should it conflict with index? If index is also provided, what should happen? Should index always prevail? I suppose it will, because it will also prevail over simple directory listing already but I'd like to align with you here.

I think `--readme` doesn't matter when `--index` is provided, as once there is index file, then it'll be served and no directory listing will be made by miniserve. Considering readme is feature on the directory listing, it'll be ignored same way `-D` flag is ignored, since those are silently ignored we can do the same here?

> You write "Renders README.md if it's present in any nested path -- that you're currently visiting.". I'm not sure that makes sense or if understand it correctly. I'd expect a README to only be picked up in the current directory that we're in but not in any subpaths of our currently visited directory, that would cause great confusion I think (and it would also be expensive to find all READMEs potentially).


Sorry. I wanted to say in case of nested paths we can render different README.md files automatically, in contrast to having to give the path of index file at start. Of course the one that'll be rendered at any moment is the one from current directory.

Also about the test, there is problem from clippy due to too many arguments in the `page` function, do you mind taking a look at it? I added `readme` argument in that page function that exceeded the limit, I don't know if we can modify other arguments to reduce the number, so you might have better idea if we can reduce it, or maybe increase the limit for clippy. When I built it in arch, it was showing only as warning and not error and the exit code of process was 0, so I didn't do anything but looks like it failed in `nightly on ubuntu latest`.

Atreyagaurav:
Does it cancel automatically once one of the tests failed? Would have been nice to know if there are some errors so we could fix in one go. Anyway, as I've said previously I'm waiting for your opinion on what to do about that too many argument problem.

svenstaro:
> Does it cancel automatically once one of the tests failed? Would have been nice to know if there are some errors so we could fix in one go. Anyway, as I've said previously I'm waiting for your opinion on what to do about that too many argument problem.

Do this for now:
```
#[allow(clippy::too_many_arguments)]
```
I think we'll just ignore this as we know what we're doing here.

svenstaro:
Also please rebase your branch so that the merge into your branch goes away as otherwise the history looks kinda funky.

src/renderer.rs
```diff
@@ -165,6 +168,17 @@ pub fn page(
...
                         }
                     }
+            @if readme.is_some() {
+            div {
+                h3 { (readme.as_ref().unwrap().file_name().unwrap()
+                  .to_string_lossy().to_string()) }
+                (PreEscaped
+                 (markdown_to_html(
+                 &std::fs::read_to_string(readme.unwrap())
+                     .unwrap_or_else(|_| "Cannot read File.".to_string()),
```
svenstaro, line 178: I think we should probably not do heavy IO work during template-time. Can you move the file reading into `listing.rs` as well?

svenstaro (changes requested):

src/renderer.rs
Atreyagaurav: I did try that, then we'd have to pass filename (could be variation on capitalization) and the content separately into 2 variables and I didn't want to do that, we can remove the filename from the render but I think it's better to be consistent with github as that's where we got the inspiration from.

src/renderer.rs
```diff
@@ -165,6 +168,17 @@ pub fn page(
                             }
                         }
                     }
+            @if readme.is_some() {
+            div {
+                h3 { (readme.as_ref().unwrap().file_name().unwrap()
+                  .to_string_lossy().to_string()) }
```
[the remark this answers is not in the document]
Atreyagaurav: I tried that, problem was to open the file we need `PathBuf` but to put filename as string we need all these layers. I was surprised at how convoluted it was to get the filename as string.

src/renderer.rs
```diff
@@ -165,6 +168,17 @@ pub fn page(
...
+                (PreEscaped
+                 (markdown_to_html(
+                 &std::fs::read_to_string(readme.unwrap())
+                     .unwrap_or_else(|_| "Cannot read File.".to_string()),
+                 &ComrakOptions::default())));
+            }
+            }
```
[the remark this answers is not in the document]
Atreyagaurav: Done.

src/listing.rs
```diff
@@ -275,13 +276,29 @@ pub fn directory_listing(
...
                     entries.push(Entry::new(
-                        file_name,
+                        file_name.clone(),
                         EntryType::File,
                         file_url,
                         Some(ByteSize::b(metadata.len())),
                         last_modification_date,
                         symlink_dest,
                     ));
+                    // TODO: Pattern match?
```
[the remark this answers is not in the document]
Atreyagaurav: I was thinking of supporting `readme.txt` as well, and `readme.org` like in github. But ended up not doing it. `txt`  should be easy enough to give it a try not sure about `org` file. I've removed the TODO for now. It can be separate issue later.

Atreyagaurav:
Looks like the formatting issue was because I hadn't set my editor to use spaces so there was tabs and it looked ok to me. I've made the other changes and force pushed. Sorry if there is some problem in intermediate commits as I'm not that familiar with rebasing. 

As for the file opening issue, We can make a struct with readme name and the content and pass that instead, and read contents in the listings instead of rendering so we have it under same variable instead of two. I read it from rendering originally to just use one variable.

Atreyagaurav:
@svenstaro  I tried with Readme struct but haben't merged to master: https://github.com/Atreyagaurav/miniserve/commit/04ec70e10400642df96c9ffd4eb4daf19cf44df9

Is this good?

src/renderer.rs
Atreyagaurav: It's done now.

src/renderer.rs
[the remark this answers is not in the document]
Atreyagaurav: I moved this to `new()` of `Readme` struct in `listing.rs` it looks clean on other places now.

Atreyagaurav:
My mind isn't adjusted to rust yet, sorry about that bool use in the struct. I've made the use of Readme struct better, I think. Making a blank struct when I could pass None was pretty stupid. 

Let me know what you think. This should at least move the heavy processing (IO) away from rendering for now.

src/args.rs
```diff
@@ -197,6 +197,10 @@ pub struct CliArgs {
     #[cfg(feature = "tls")]
     #[clap(long = "tls-key", requires = "tls-cert", value_hint = ValueHint::FilePath)]
     pub tls_key: Option<PathBuf>,
+
+    /// Enable README.md redering in directories
```
svenstaro, line 201: "rendering" is still misspelt.

src/renderer.rs
```diff
@@ -165,6 +167,12 @@ pub fn page(
                             }
                         }
                     }
+            @if readme.is_some() {
+                div {
+                    h3 { (readme.as_ref().unwrap().filename) }
```
svenstaro, line 172: Do you still need this construct now that you have the special struct for it?

svenstaro (changes requested):

src/renderer.rs
Atreyagaurav: Since `readme` here is `Option<Readme>` looks like it'll be moved when I use `.unwrap()` so to access two fields separately I need to use `.as_ref()` the first time. Other option would be to save unwrapped value to a variable I guess.

svenstaro: I think we can probably find a simpler solution here. It doesn't prevent me from merging but I might golf it some more.

Atreyagaurav:
Sorry didn't notice the spelling, I somehow only saw the `README.md` part, anyway it's fixed now. The `.as_ref().unwrap()` is just rust I guess, To make things safe. I did it as suggested by the compiler.

src/renderer.rs
Atreyagaurav: Probably, but considering previous code I think we've made it a lot straightforward, so, I'm satisfied.

svenstaro:
Merging as is. Thanks, this is great!

svenstaro:
Oh, I just noticed we don't actually have a test for this. Would you like to add one after the fact, @Atreyagaurav?

Atreyagaurav:
I've never written rust tests, and currently have no idea how to test readme thing, let me read other tests and see what I can do. Might take a week or so.

svenstaro:
Not to worry! I think the tests are actually fairly easy to read and it might be a good learning exercise at any rate.

PR #869, opened 2022-08-10
Optimize for smaller binaries

ZuoFuhong:
Cargo can be configured to automatically strip binaries. 

Before:

image

now:

image


.gitignore
```diff
@@ -7,3 +7,4 @@
 
 /target
 **/*.rs.bk
+.idea/
```
svenstaro, line 10: I'm not sure how I feel about editor-specific ignores here.

Cargo.toml
```diff
@@ -15,6 +15,7 @@ lto = true
 opt-level = 'z'
 codegen-units = 1
 panic = 'abort'
+strip = true
```
svenstaro, line 18: We already strip in post but I suppose turning this on doesn't hurt.

.gitignore
ZuoFuhong: Maybe you are not using the idea editor.

svenstaro: Ok, but at least section it of with something like `# Editor specific ignores below`

ZuoFuhong: Ok, i submitted

svenstaro:
Thanks.

Issue #879, opened 2022-08-24
Questions for tests on the `--readme` flag


Atreyagaurav:
Not really a issue but more of a question here:

So I can add the id on readme contents like: 

```rust
            @if readme.is_some() {
                div id="readme" {
                    h3 id="readme-filename" { (readme.as_ref().unwrap().filename) }
                    div id="readme-contents" {
			(PreEscaped (readme.unwrap().contents));
		    }
                }
            }
```

Now I can test for the presence of those id in the document body, for example:

```rust
#[rstest]
/// Do not show readme contents unless told to
fn hide_readme_contents(server: TestServer) -> Result<(), Error> {
    let body = reqwest::blocking::get(server.url())?.error_for_status()?;
    let parsed = Document::from_read(body)?;
    assert!(parsed.find(Attr("id", "readme")).next().is_none());
    assert!(parsed.find(Attr("id", "readme-filename")).next().is_none());
    assert!(parsed.find(Attr("id", "readme-contents")).next().is_none());

    Ok(())
}
```

This one seems straightforward enough, problem now I'm facing is, how do I simulate the contents of a directory with `readme.md` file?

Do I have to make some files in the tests directory, include it in git, and then use them as args in the cli? Or can I just create dummy files in the code itself??

EDIT: I read the fixures module and saw the `testdir` has the files with contents, so I used it while also adding `readme.md` file with some contents in all the directories.

PR #880, opened 2022-08-26
Add tests for `--readme`

Atreyagaurav:
Fixes https://github.com/svenstaro/miniserve/issues/879

I've tried a bit here to write tests. Idk why the `child.kill()?` isn't working properly so I've made two tests in different servers. 

I also tried to add `readme.md` in `FILES` to just use the testserver, but some other tests failed when I did that.

### DIFF
--- a/tests/readme.rs
+++ b/tests/readme.rs
@@ -6,6 +6,7 @@ use select::predicate::Attr;
 use select::{document::Document, node::Node};
 use std::fs::{remove_file, File};
 use std::io::Write;
+use std::path::PathBuf;
 
 #[rstest]
 /// Do not show readme contents by default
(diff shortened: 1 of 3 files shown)


Atreyagaurav:
I think the tests now will check the contents of the readme is there and the readme div are there. 
@svenstaro let me know if I need to add more tests. Or edit these to make them better.

tests/readme.rs
```diff
@@ -0,0 +1,124 @@
...
+/// Show readme contents when told to if there is readme.md file
+fn show_readme_contents(tmpdir: TempDir) -> Result<(), Error> {
+    tmpdir
+        .child("readme.md")
+        .write_str("Readme Contents.")
+        .expect("Couldn't write to readme.md");
+    let mut child = Command::cargo_bin("miniserve")?
+        .arg("--readme")
+        .arg("--port")
+        .arg("8090")
```
svenstaro, line 37: Instead of the manual server handling that you're doing here, how about injecting the `TestServer` dependency with an argument like in other tests? See for instance `fn uploading_files_works`. Then you also don't need to kill it manually.

tests/readme.rs
```diff
@@ -0,0 +1,124 @@
...
+        .arg(tmpdir.path())
+        .stdout(Stdio::null())
+        .spawn()?;
+
+    sleep(Duration::from_secs(1));
+    let body = reqwest::blocking::get("http://localhost:8090")?.error_for_status()?;
```
svenstaro, line 43: How about `let body = reqwest::blocking::get(server.url())` this instead? It'll automatically get the right URL for you with a free port.

tests/readme.rs
```diff
@@ -0,0 +1,124 @@
...
+            == "Readme Contents."
+    );
+
+    child.kill()?;
+    Ok(())
+}
+
+#[rstest]
+/// Show readme contents when told to if there is readme.md file on directories.
+fn show_readme_contents_directories(tmpdir: TempDir) -> Result<(), Error> {
```
svenstaro, line 72: I don't really see how this test is conceptually different from the one above. Can't you merge these and use the `rstest`-based paramterization (that we use in many other places in the tests already, have a look around) to test more directories?

tests/readme.rs
```diff
@@ -0,0 +1,124 @@
...
+        .arg("--port")
+        .arg("8090")
+        .arg(tmpdir.path())
+        .stdout(Stdio::null())
+        .spawn()?;
+
+    sleep(Duration::from_secs(1));
+    let body = reqwest::blocking::get("http://localhost:8090")?.error_for_status()?;
+    let parsed = Document::from_read(body)?;
+    assert!(parsed.find(Attr("id", "readme")).next().is_some());
```
svenstaro, line 45: It seems to me it might be neat to check whether the actual file listing is still shown above the README.

tests/readme.rs
```diff
@@ -0,0 +1,124 @@
+mod fixtures;
+
+use assert_cmd::prelude::*;
+use assert_fs::fixture::{FileWriteStr, TempDir};
+use assert_fs::prelude::PathChild;
```
svenstaro, line 5: Can you merge this to:
```suggestion
use assert_fs::prelude::*;
```
?
I'm not sure the `prelude` has all the required packages but that's usually what I try first.

svenstaro (changes requested):
Good job! I think with a few minor changes we can easily get this merged.

tests/readme.rs
Atreyagaurav: I tried that first, but I've to write the `readme.md` contents in tempdir, so I needed the tempdir location and saw similar test on `server_request.rs` and followed that example.

tests/readme.rs
Atreyagaurav: Same reason as previous. Needed the `tempdir` so took this from `server_request.rs`

tests/readme.rs
Atreyagaurav:
```
error[E0412]: cannot find type `TempDir` in this scope
  --> tests/readme.rs:28:33
   |
28 | fn show_readme_contents(tmpdir: TempDir) -> Result<(), Error> {
   |                                 ^^^^^^^ not found in this scope
   |
help: consider importing this struct
   |
3  | use assert_fs::TempDir;
   |

error[E0412]: cannot find type `TempDir` in this scope
  --> tests/readme.rs:71:45
   |
71 | fn show_readme_contents_directories(tmpdir: TempDir) -> Result<(), Error> {
   |                                             ^^^^^^^ not found in this scope
   |
help: consider importing this struct
   |
3  | use assert_fs::TempDir;
```
just having `use assert_fs::prelude::*;` instead of those two gives me this error.

svenstaro: Ah yes, nevermind then.

Atreyagaurav:
I'll try to combine the directories test and the root one, as well as add the test for directory listings. 

I just committed this one to see if it passes the timeline, coz idk why adding `readme.md` in the `FILES` in the fixtures was failing before, but it's passing for me locally.

tests/readme.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    Ok(())
+}
+
+#[rstest(
+    dir,
+    content,
+    case("", "Test Hello Yes"),
+    case("/dira/", "This is dira/readme.md"),
+    case("/dirb/", "This is dirb/readme.md")
```
svenstaro, line 31:
```suggestion
    case("/dirb/", "This is dirb/README.md")
```
Let's test uppercase as well.

tests/readme.rs
```diff
@@ -0,0 +1,68 @@
...
+)]
+/// Show readme contents when told to if there is readme.md file
```
svenstaro, line 33:
```suggestion
/// Show readme contents when told to if there is a README.md file
```

svenstaro (changes requested):
Almost there! Looking quite good now.

tests/readme.rs
Atreyagaurav: Any idea how to test both?

for now I think to test uppercase would be problematic, having both "README.md" and "readme.md" (thought might not exactly be undefined behavior) there is no priority for one over another. And fixtures is automatically making the file "readme.md" on all directories, so adding "README.md" in the `FILES` might make things confusing to test.

svenstaro: You could put the uppercased one in a different directory.

Atreyagaurav:
@svenstaro  it's not the most elegant way to test it but I made a different directory containing just the `README.md` file for testing in a different case. Since the fixures adds `readme.md` to all the directories, I made a different directory for it. It should work for now.

Atreyagaurav:
Ok, I think this is way better. Checks different combination of cases, as well as all the directories in the fixures. no hardcoded file paths in the test cases.

Atreyagaurav:
@svenstaro whenever you have time, please have a look. The tests should be ok now.

svenstaro:
Thanks! There are only some minor things but I'll fix them myself after merging.

PR #898, opened 2022-09-11
Added uppy

draxaris1010:
Hi
I've added uppy so you would be able to track the upload progress.


Yusuto:
This is pretty neat!

I suggest adding a way to choose either Uppy or the old file upload when starting the server, and if JavaScript is disabled that it always falls back to the old file upload, I think those would be very nifty things to have.

Also the layout is a bit odd on mobile, maybe there is a way to embed the widget into the site instead of opening a modal?

Oh and since this PR is rather old it would need to be manually merged, but I think I could do that!

chaibiq:
Hi, I'm working on this progress bar for upload files, but not ready to merge. It's use external JavaScript to create xhr and upload files with progress bars.
Untitled

svenstaro:
Closing favor of #1431 which is likely to be merged soon.

PR #910, opened 2022-09-19
Fix `fake-tty` API breaking change

cyqsimon:
It's difficult to create a more trivial commit. Possible, but difficult.


svenstaro:
:D

PR #911, opened 2022-09-19
Add support for README.txt and README files

Atreyagaurav:
Since we added support for readme.md I thought it'd be easy enough to do readme.txt and just plain readme.

I was thinking of adding support for readme.org like in GitHub but didn't find any libraries that can convert it easily.

Also, if I rename "README.md" from miniserve to .txt, it only shows a fraction of the file contents, I can't debug it...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ mime = "0.3"
 nanoid = "0.4"
 percent-encoding = "2"
 port_check = "0.1"
+regex = "1"
 rustls = { version = "0.20", optional = true }
 rustls-pemfile = { version = "1.0", optional = true }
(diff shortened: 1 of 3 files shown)


Atreyagaurav:
Additional points:

It's trivial to add support for README.html, we can just insert the file contents to readme div, should I? I don't know if people use htmls as readme. But it might help if people want some advanced formatting or to convert unsupported formats to html (which preserves more info then converting to markdown).

svenstaro:
Hm I'm not sure about README.html as it might open us up to some weird and unforeseen JavaScript injection attacks in conjunction with uploads. Let's keep it to known formats.

svenstaro:
Implementation looks very good and I appreciate the easily expandable test. :)

PR #935, opened 2022-10-14
WIP: Feature: listen on unix domain sockets

FallingSnow:
### Synopsis
Initial conversations converged on using the `--port` argument but after attempting to implement this, `--interface` seemed more promising based on the current code structure. Also, a unix socket is more similar to an interface than a port so this made sense logically as well.

Interface arguments are now parsed into an `Interface` enum consisting of either a `Interface::Address` (IP address) or `Interface::Path` (Unix socket path). If an interface argument starts with a `.` or `/` it is considered a `Interface::Path`.

Currently the code does not clean up the socket file it creates. I'm not really sure what the desired behavior should be. Perhaps if miniserve creates the socket file, it should also delete it on shutdown. But then again what if another, different process is listening on the socket file... This does not cause any issues with starting miniserve up again since miniserve will use an existing socket file.

You can listen on both unix sockets and tcp sockets at the same time when given more than one `--interface` flag.

### Tests
As for test code there is only one added test to ensure we don't overwrite non-socket files. I tried to get a `GET` test working but it keeps throwing an error about serve path even though it is provided. The non working test is commented out in `tests/bind.rs`.
```rust
01:11:46 [ERROR] Failed to resolve path to be served
01:11:46 [ERROR] caused by: No such file or directory (os error 2)
```

A test for IPv6 binding is failing but seems to work in manual testing. Not sure what is going on there.

### Testing
An easy way to test this is to use the following commands:
```sh
$ cargo run -- -i /tmp/test.socket ./ -v
$ curl -X GET --unix-socket /tmp/test.socket http://localhost
```

You can also test in your browser using socat as a TCP->UNIX proxy:
```
$ socat TCP-LISTEN:8080,fork /tmp/test.socket
```

### Todo
- [ ] Add more tests

---
Let me know if you have any questions or changes you'd like.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,12 @@ Sometimes this is just a more practical and quick way than doing things properly
 This uses the `--media-type` option, which sends a hint for the expected media type to the browser.
 Some mobile browsers like Firefox on Android will offer to open the camera app when seeing this.
 
+### Listen on a unix domain socket
+
+    miniserve -p /tmp/web.socket
+
+You can test access to this socket using curl: `curl -X GET --unix-socket /tmp/web.socket http:/localhost/index.html`
+
 ## Features
 
 - Easy to use
--- a/src/args.rs
+++ b/src/args.rs
@@ -42,9 +42,9 @@ pub struct CliArgs {
     #[clap(long, requires = "index_file")]
     pub spa: bool,
 
-    /// Port to use
+    /// Port or unix socket path to listen on
     #[clap(short = 'p', long = "port", default_value = "8080")]
-    pub port: u16,
+    pub port: String,
 
     /// Interface to listen on
     #[clap(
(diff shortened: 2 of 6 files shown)


svenstaro:
Hey, is this still WIP? Would you like to add some tests and rebase and then I can take a proper look at this?

FallingSnow:
Wasn't looking forward to fixing the tests. Let me try to finish this up by today/tomorrow.

FallingSnow:
Ready for review.

I moved the tests so that the working directory is set to the tmpdir of each test. This required tls files to be copied into a test's tmpdir so it could access them. This was mainly done so each test could create a socket file in their own directory.

If you'd like any changes please let me know.

FallingSnow:
Not sure why checks are failing. Same command succeeds on my machine. `cargo fmt --all -- --check`

Issue #748, opened 2022-03-04
Please add timezone support


Karmenzind:
Thanks a lot

svenstaro:
This is now in there since 0.23.0.

Issue #440, opened 2021-01-31
DateTime format


sonichy:
Change
`Jun 24 at 15:21`
to
`2020/06/24 15:21:16`

Order directory first then file, time desc or filesize desc.

Theme not pretty.

CheungChan:
please add timezone please。

PR #949, opened 2022-10-23
Change DateTime format

IvkinStanislav:
Fix issues [440](https://github.com/svenstaro/miniserve/issues/440) and [748](https://github.com/svenstaro/miniserve/issues/748)

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -1,7 +1,7 @@
 use std::time::SystemTime;
 
 use actix_web::http::StatusCode;
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, Local};
 use chrono_humanize::Humanize;
 use clap::{crate_name, crate_version, ValueEnum};
 use fast_qr::{convert::svg::SvgBuilder, qr::QRCodeError, QRBuilder};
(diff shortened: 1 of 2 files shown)


src/renderer.rs
```diff
@@ -622,13 +620,10 @@ fn page_header(title: &str, file_upload: bool, favicon_route: &str, css_route: &
...
-    src_time.map(DateTime::<Utc>::from).map(|date_time| {
-        (
-            date_time.format("%b %e").to_string(),
-            date_time.format("%R").to_string(),
-        )
-    })
+fn convert_to_utc(src_time: Option<SystemTime>) -> Option<String> {
+    src_time
+        .map(DateTime::<Local>::from)
+        .map(|date_time| date_time.format("%Y/%m/%d %H:%M:%S %:z").to_string())
```
svenstaro, line 626: Let's have these be displayed in the more traditional dash separated format:
```suggestion
        .map(|date_time| date_time.format("%Y-%m-%d %H:%M:%S %:z").to_string())
```
Alternatively, you could also use `to_rfc2822()` which might be more readable to some people. Your call.

svenstaro (changes requested):
Thanks, this is good stuff. I'd just like the format to be a adjusted slightly.

IvkinStanislav:
> Thanks, this is good stuff. I'd just like the format to be a adjusted slightly.

Ready

svenstaro:
Thanks!

PR #955, opened 2022-11-06
Use gcr.io/distroless/cc as a glibc-based container instead of ubuntu

### LINKED ISSUE #939: Create Distroless Container Image (again)
Huhu, out of security perspective.
It would be nice to has an Image without Distro (and all there vulnubilities inside).

This was already, but changed in #904 (to an ubuntu), by coping the binary from builder image to a new `FROM scratch`.

svenstaro:
This should have less useless stuff in the image and fix #939.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,3 +1,3 @@
-FROM docker.io/ubuntu
+FROM gcr.io/distroless/cc
 COPY --chmod=755 miniserve /app/
 ENTRYPOINT ["/app/miniserve"]


lilnasy:
I would've expected a nonroot image to make the cut too. Does miniserve require root?

svenstaro:
What do you mean? You can perfectly well execute this image on podman without root required. In fact that's how I tested it.

lilnasy:
That's true. You can start up this container without needing root access on the host machine, but it's a good practice to run the applications without root access _inside_ the container. I have seen container breakouts that are made possible because of root access within the namespaces. Running as a specific user or even a non-zero user ID could potentially prevent some of them.

svenstaro:
According to the docs, `podman|docker run --user` overwrites the provided user id in the `Containerfile`. As such, I think I'd prefer to leave this completely up to users. Also, if you use `podman`, the default user inside the container is actually mapped to your own user outside it so I think we should go by the principle of least surprise there unless you can point me to some source that strongly argues otherwise.

lilnasy:
No, I think this is fine, leaving it upto the users is a good idea.

Issue #964, opened 2022-11-09
The directory is sorted first


fcwys:
I hope the directory can be sorted first. It is inconvenient to mix directories and files together.

levaitamas:
Hi @fcwys ! Good news, since [v0.10.4](https://github.com/svenstaro/miniserve/releases/tag/v0.10.4) there is a CLI flag  `-D` or `--dirs-first`, which implements this functionality.

fcwys:
It is expected that this option is enabled by default

svenstaro:
I'm not sure this behavior would necessarily be expected by default. I think the default behavior of other applications is pretty mixed in this regard.

Issue #306, opened 2020-05-12
Size of directories is not calculated


NightMachinery:
![image](https://user-images.githubusercontent.com/36224762/81748620-66a56080-94bf-11ea-9acc-51b32f0f64fa.png)

wyhaya:
I don't think this function should be added, this requires recursive directory to get file size and then merge.

it will be very slow.

svenstaro:
Eh let's reopen it. It could be fast enough and it could also be cached. It should be opt-in though for sure so that the host can't be DOSed by default.

wyhaya:
@svenstaro If you encounter a directory like `node_modules`, there may be 10,000-100,000... files, is there any good way?

svenstaro:
Yeah, how about setting a useful timeout per directory?

wyhaya:
I think timeout is not good. If it cannot be calculated quickly, it will time out every time, which means that the server can never calculate the size of some folders. The timeout should be set on some unstable functions, such as network packets.

NightMachinary:
The timeout can be cached as “-1”. Or “NA.” I personally don’t have that many files.

zamu-flowerpot:
I came here to mention KDE's file count and glad it was already been mentioned.

PR #965, opened 2022-11-09
feat/issue-306-pr 

zamu-flowerpot:
This would resolve issue #306 with entry counts for directories. 
This also resolves issue #964 as I had to define how to compare `EntrySize::EntryCount` and `EntrySize::Bytes`.

This adds a single dependency [`walkdir`](https://docs.rs/walkdir/latest/walkdir/) to handle counting the number of items in a directory cross-platform. Also, by default it doesn't follow symbolic links which seems fine.

### DIFF
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -1,4 +1,5 @@
 #![allow(clippy::format_push_string)]
+use std::fmt::Display;
 use std::io;
 use std::path::{Component, Path, PathBuf};
 use std::time::SystemTime;
@@ -10,6 +11,7 @@ use percent_encoding::{percent_decode_str, utf8_percent_encode};
 use regex::Regex;
 use serde::Deserialize;
 use strum::{Display, EnumString};
+use walkdir::WalkDir;
 
 use crate::archive::ArchiveMethod;
 use crate::auth::CurrentUser;
@@ -78,6 +80,24 @@ pub enum EntryType {
     File,
 }
 
+#[derive(Clone, Copy, PartialEq, Eq)]
+/// Possible entry size types
+pub enum EntrySize {
+    /// EntryCount is number of entries in a directory
+    EntryCount(usize),
+    /// Bytes is number of bytes in a file
+    Bytes(bytesize::ByteSize),
+}
+
+impl Display for EntrySize {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            EntrySize::EntryCount(count) => write!(f, "{}", count),
+            EntrySize::Bytes(bytes) => write!(f, "{}",  bytes),
+        }
+    }
+}
+
 /// Entry
 pub struct Entry {
     /// Name of the entry
@@ -89,8 +109,8 @@ pub struct Entry {
     /// URL of the entry
     pub link: String,
 
-    /// Size in byte of the entry. Only available for EntryType::File
-    pub size: Option<bytesize::ByteSize>,
+    /// Size of the entry
+    pub size: EntrySize,
 
     /// Last modification date
     pub last_modification_date: Option<SystemTime>,
@@ -104,7 +124,7 @@ impl Entry {
         name: String,
         entry_type: EntryType,
         link: String,
-        size: Option<bytesize::ByteSize>,
+        size: EntrySize,
         last_modification_date: Option<SystemTime>,
         symlink_info: Option<String>,
     ) -> Self {
@@ -256,12 +276,14 @@ pub fn directory_listing(
                     Err(_) => None,
                 };
 
+                let size = WalkDir::new(entry.path()).into_iter().count();
+
                 if metadata.is_dir() {
                     entries.push(Entry::new(
                         file_name,
                         EntryType::Directory,
                         file_url,
-                        None,
+                        EntrySize::EntryCount(size),
                         last_modification_date,
                         symlink_dest,
                     ));
@@ -270,7 +292,7 @@ pub fn directory_listing(
                         file_name.clone(),
                         EntryType::File,
                         file_url,
-                        Some(ByteSize::b(metadata.len())),
+                        EntrySize::Bytes(ByteSize::b(metadata.len())),
                         last_modification_date,
                         symlink_dest,
                     ));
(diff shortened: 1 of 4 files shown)


zamu-flowerpot:
One thing I didn't look into was setting these behind CLI flags.

svenstaro:
I think this should probably be behind CLI flags as the performance hit can be severe and we don't want that by default. I will do a proper review a little later.

src/renderer.rs
```diff
@@ -511,19 +511,23 @@ fn entry_row(
...
                 }
             }
             td.size-cell {
-                @if let Some(size) = entry.size {
-                    (maud::display(size))
-                }
+                (maud::display(&entry.size))
+                // @if let size = entry.size {
+                //     (maud::display(size))
+                // }
```
svenstaro, line 530: What's with the commented code here?

src/listing.rs
```diff
@@ -302,9 +324,15 @@ pub fn directory_listing(
...
-                .cmp(&e1.size.unwrap_or_else(|| ByteSize::b(0)))
+            match (e1.size, e2.size) {
+                (EntrySize::EntryCount(ref e1_count), EntrySize::EntryCount(ref e2_count)) => e2_count.cmp(e1_count),
+                (EntrySize::Bytes(ref e1_bytes), EntrySize::Bytes(ref e2_bytes)) => e2_bytes.cmp(e1_bytes),
+                (EntrySize::EntryCount(_), EntrySize::Bytes(_)) => std::cmp::Ordering::Greater,
+                (EntrySize::Bytes(_), EntrySize::EntryCount(_)) => std::cmp::Ordering::Less,
+            }
+            // e2.size
+            //     .unwrap_or_else(|| ByteSize::b(0))
+            //     .cmp(&e1.size.unwrap_or_else(|| ByteSize::b(0)))
```
svenstaro, line 335: Also this.

src/listing.rs
```diff
@@ -78,6 +80,24 @@ pub enum EntryType {
...
+
+impl Display for EntrySize {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            EntrySize::EntryCount(count) => write!(f, "{}", count),
+            EntrySize::Bytes(bytes) => write!(f, "{}",  bytes),
+        }
+    }
+}
+
```
svenstaro, line 100: Wouldn't it be cool to also show the combined directory file size?

svenstaro:
Cool stuff! As you already said, a memoization technique would be neat. I think a `HashMap` would be good enough for us. What do you also think about timing out on directories that just take too long? I don't miniserve to ever not feel _snappy_ when generating pages.

src/renderer.rs
zamu-flowerpot: Ah, I didn't remove that since it was the code that would need to be switched on if we were going to throw everything behind a feature flag.

src/listing.rs
zamu-flowerpot: Not sure what you are referring to by "combined directory file size". Are you referring to the total size in bytes of a directory? That might be a bit expensive and probably would require a persistent cache to be anywhere near fast.

zamu-flowerpot:
> Cool stuff! As you already said, a memoization technique would be neat. I think a `HashMap` would be good enough for us. [...] 

A `HashMap` would be fine. I tried my hand at using a `once_cell::OnceCell<HashMap>`, but couldn't figure out how to make a mutable global static. Maybe it'd just be as simple as just constructing it and passing it in as part of the function call to `directory_listing`?

> [...] What do you also think about timing out on directories that just take too long? I don't miniserve to ever not feel _snappy_ when generating pages.

Definitely. We might be able to shortcut a lot of the issues by just counting direct children by changing the size calculation to call `.max_depth(1)`. It still might be slow, but it's on the user if they have a directory with a million files right?

src/listing.rs
```diff
@@ -256,12 +276,14 @@ pub fn directory_listing(
                     Err(_) => None,
                 };
 
+                let size = WalkDir::new(entry.path()).into_iter().count();
```
zamu-flowerpot, line 279: Thinking about it, we might be able to just do `.take(UPPER_LIMIT).count()` instead of `.count()` to shortcut execution time. So, we might have `UPPER_LIMIT` be something like 10001 and if it's more than 10000 we have the Impl Display for EntrySize print `>10000` or something along those lines.

zamu-flowerpot: The other option would have been to add an actual timeout, but I'm not really sure how to best do that in a non-async context.

src/listing.rs
svenstaro: Yes, the byte sizes. I recognize it might be too slow. Hm. If you like to, you could try to experiment with this.

svenstaro:
> > Cool stuff! As you already said, a memoization technique would be neat. I think a `HashMap` would be good enough for us. [...]
> 
> A `HashMap` would be fine. I tried my hand at using a `once_cell::OnceCell<HashMap>`, but couldn't figure out how to make a mutable global static. Maybe it'd just be as simple as just constructing it and passing it in as part of the function call to `directory_listing`?

You're going to need a `OnceCell<Arc<Mutex<HashMap>>>`.

> 
> > [...] What do you also think about timing out on directories that just take too long? I don't miniserve to ever not feel _snappy_ when generating pages.
> 
> Definitely. We might be able to shortcut a lot of the issues by just counting direct children by changing the size calculation to call `.max_depth(1)`. It still might be slow, but it's on the user if they have a directory with a million files right?

How about walking through the tree in threads using rayon? Should be fairly easy and it might also allow you to cancel/timeout threads.

zamu-flowerpot:
> You're going to need a `OnceCell<Arc<Mutex<HashMap>>>`.

Yep, worked like a charm.
  
> How about walking through the tree in threads using rayon? Should be fairly easy and it might also allow you to cancel/timeout threads.

Tried rayon and then encountered https://github.com/BurntSushi/walkdir/issues/21. This lead me to look for something other than [WalkDir](https://github.com/BurntSushi/walkdir) and found [fs_extra](https://github.com/webdesus/fs_extra). 

I just wanted to get this to a MVP and figure out what flags we want to incorporate (and how to wire them from the config/cli to the call site). 

Looking at `fs_extra`s code, it might be doable to parallelize the size scanning using Rayon. If we get all the decedent `DirEntry`s and push them onto a stack, for_each over the stack and cache their size, and then calculate the directory sizes from the cache on request, we should be pretty fast. The biggest issue still comes down to cache invalidation though it wouldn't be too hard to change the cache map to include modified time.

That said another option is to spawn a thread that listens for file system notifications and updates the cache on writes, but I'm only aware of Linux's inotify so I'm not sure if it's really cross-platform.

Let me know if you have any questions or concerns.

svenstaro:
@zamu-flowerpot Hey, would you like to rebase this and clean it up? I'd like to get this feature into the next release.

zamu-flowerpot:
It's been 2 years and I don't particularly have the time to tinker with this. 

I did take a look at it though and fs_extra is long since abandoned, so some solution for indexing the directories is needed. All the other conflicts are minor things.

svenstaro:
> It's been 2 years and I don't particularly have the time to tinker with this.
> 
> I did take a look at it though and fs_extra is long since abandoned, so some solution for indexing the directories is needed. All the other conflicts are minor things.

Fair enough.

svenstaro:
Closing this in favor of #1482.

svenstaro:
I'm aware, that's why there's been no release of this yet. I'm not really happy with the current mechanism. The current goal is to make this cancelable and also for the server to cache it. Also there should be no concurrent calculations of the same directory.

svenstaro:
Well there was a feature request to add folder counting. I'll add a flag for disabling this today for the time being.

svenstaro:
Directory size counting is now disabled by default and can be enabled via `--directory-size`.

PR #991, opened 2022-12-06
[fix] Sorting breaks subdir downloading

### LINKED ISSUE #967: Sorting by any means breaks subdir downloads
When entering a sub directory and pressing any of the sorting buttons, the `download .x` button changes its url from `example.com/cool_folder/?download=zip` to `example.com/?sort=size&order=asc&download=zip`, which forces you to download everything instead of that specific directory. 

This seems to be caused by sorting added a '/?' to the download button, where as no sorting is just '?'.

Vam-Jam:
Adds a small check in `render.rs` `make_link_with_trailing_slash` to do nothing if link is empty, instead of adding a slash where its not needed.

It looks like nothing calls make_link_with_trailing_slash with an empty string aside from downloading when sorting, so this should not break anything else.

You can test the fix out here if needed: https://archive.vamist.dev/

Another alternative fix is to just not include sorting options when downloading, so if you would rather this, let me know.

Fixes #967

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -411,7 +411,7 @@ fn archive_button(
 
 /// Ensure that there's always a trailing slash behind the `link`.
 fn make_link_with_trailing_slash(link: &str) -> String {
-    if link.ends_with('/') {
+    if link.is_empty() || link.ends_with('/') {
         link.to_string()
     } else {
         format!("{}/", link)


svenstaro:
Thanks! Fix is fine as is.

Vam-Jam:
Awesome, thank you for making this awesome project!

Issue #1039, opened 2023-02-19
--title flag also changes the wget command


Yusuto:
When changing the title using `--title`, the value that title is set to will also be set as the domain name that the wget command generated and displayed using the `--show-wget-footer` flag. That means that if I set the title to for example »Star«, that the command displayed at the bottom would now be `wget -r -c -nH -np --cut-dirs=0 -R "index.html*" --ask-password --user user "http://Star/?raw=true"`, I don't think it's intended to be like that.

Instead there should be my IP-Address (and port) there, or a domain name, right? I'm not sure if miniserve can even know that, so maybe there should be another flag for that?

Yusuto:
I think I have an idea of how to implement this now, the link is formed through the `breadcrumbs` that `listing.rs` passes to the page function in `renderer.rs`. The first element of this path is set to either the title, or if it is missing, to the host address retrieved from the request. The former is what we want for the page title and the heading, but for the wget command we definitely want the host.

I want to implement it without it being a botch or so, but I'm very unsure how exactly to do that. If it were my program I think I'd either make `breadcrumbs` always use the title if available and in the wget footer, swap the title for the host, or the other way around, but I'm very unsure what is better. The former seems better because it seems like the breadcrumb names are mainly used for displaying them to the user, but then the latter would also seem reasonable because then all the names strung together with slashes in between and a `http://` at the front would make a link together.

Does anyone have an idea which of the two would be better?

Yusuto:
I discovered this literally a minute after I wrote this, I can just use the `abs_url` passed to the page function to pass to `wget_footer`, instead of passing `title_path`. With some small changes it works just fine! I'm not sure if that counts as a botch though, but I think it should be fine?

PR #1043, opened 2023-02-20
Fix the wget footer

Yusuto:
...by having it use the absolute url instead of the breadcrumbs.
I think the changes should be rather self-explanatory, I tested my changes before opening this PR and also ran Clippy and rustfmt and neither said anything, but please still check what I did, just in case.

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -194,7 +194,7 @@ pub fn page(
                     }
                     div.footer {
                         @if conf.show_wget_footer {
-                            (wget_footer(abs_uri, breadcrumbs[0].name.as_str(), current_user))
+                            (wget_footer(abs_uri, conf.title.as_deref(), current_user.map(|x| &*x.name)))
                         }
                         @if !conf.hide_version_footer {
                             (version_footer())
@@ -268,11 +268,20 @@ fn version_footer() -> Markup {
     }
 }
 
-fn wget_footer(abs_path: &Uri, root_dir_name: &str, current_user: Option<&CurrentUser>) -> Markup {
+fn wget_footer(abs_path: &Uri, root_dir_name: Option<&str>, current_user: Option<&str>) -> Markup {
+    fn escape_apostrophes(x: &str) -> String {
+        x.replace('\'', "'\"'\"'")
+    }
+
     // Directory depth, 0 is root directory
     let cut_dirs = match abs_path.path().matches('/').count() - 1 {
         // Put all the files in a folder of this name
-        0 => format!(" -P {root_dir_name}"),
+        0 => format!(
+            " -P '{}'",
+            escape_apostrophes(
+                root_dir_name.unwrap_or_else(|| abs_path.authority().unwrap().as_str())
+            )
+        ),
         1 => String::new(),
         // Avoids putting the files in excessive directories
         x => format!(" --cut-dirs={}", x - 1),
@@ -280,12 +289,12 @@ fn wget_footer(abs_path: &Uri, root_dir_name: &str, current_user: Option<&Curren
 
     // Ask for password if authentication is required
     let user_params = match current_user {
-        Some(user) => format!(" --ask-password --user {}", user.name),
+        Some(user) => format!(" --ask-password --user '{}'", escape_apostrophes(user)),
         None => String::new(),
     };
 
     let command =
-        format!("wget -rcnHp{cut_dirs} -R 'index.html*'{user_params} '{abs_path}?raw=true'");
+        format!("wget -rcnHp -R 'index.html*'{cut_dirs}{user_params} '{abs_path}?raw=true'");
     let click_to_copy = format!("navigator.clipboard.writeText(\"{command}\")");
 
     html! {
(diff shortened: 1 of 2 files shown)


src/renderer.rs
```diff
@@ -267,11 +267,11 @@ fn version_footer() -> Markup {
...
 
-fn wget_footer(title_path: &str, current_user: Option<&CurrentUser>) -> Markup {
+fn wget_footer(abs_path: &str, current_user: Option<&CurrentUser>) -> Markup {
     let count = {
-        let count_slashes = title_path.matches('/').count();
-        if count_slashes > 0 {
-            count_slashes - 1
+        let count_slashes = abs_path.matches('/').count();
+        if count_slashes >= 4 {
+            count_slashes - 4
```
svenstaro, line 274: That's a bit sketchy. Why 4?

Yusuto: Oh, that change is because `abs_path` has more slashes than the breadcrumbs, since it includes the protocol. But wait, why did I increase it by three, not just two? It should be `http://test.net/my-folder/.git` compared to `test.net/my-folder/.git`, only two more slashes... I made it so the `--cut-dirs` flag has the same value as before the change, but I can't explain exactly why it is that way, Imma investigate

Yusuto: Oh I know why! `abs_path` includes an extra slash at the end which the breadcrumbs didn't, so the two slashes from the protocol plus the one at the end made me increase it by three. But that means that there is now two slashes at the end of the path in the command, I need to fix that.

Yusuto: I think I made the fix and it should be part of the PR now, I made an embarrassing typo in the commit message though which I discovered too late because I wasn't careful before pushing to my fork, I fixed the typo in an amend and force pushed that to my fork, I hope that's not causing any issues.

svenstaro:
Please squash your commits.

svenstaro:
Fixes #1039.

Yusuto:
> Please squash your commits.

I think I squashed my commits, I hope I did what you mean and didn't mess anything up with the force pushes

src/renderer.rs
```diff
@@ -267,11 +267,11 @@ fn version_footer() -> Markup {
     }
 }
 
-fn wget_footer(title_path: &str, current_user: Option<&CurrentUser>) -> Markup {
+fn wget_footer(abs_path: &str, current_user: Option<&CurrentUser>) -> Markup {
     let count = {
-        let count_slashes = title_path.matches('/').count();
-        if count_slashes > 0 {
-            count_slashes - 1
+        let count_slashes = abs_path.matches('/').count();
```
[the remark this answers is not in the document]
Yusuto: It's for the `--cut-dirs` of the wget command which I think makes it so you don't download into a bunch of unnecessary subfolders, I will add the comment and add it to the last comment (so another force push) in a bit!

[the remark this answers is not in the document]
Yusuto: Bit of a panicked mess but I think I got it now!

svenstaro:
It's not necessary for this MR btw, but if you'd like to, you could write a small test to make sure the behavior for this footer is always functional. :)

Yusuto:
> It's not necessary for this MR btw, but if you'd like to, you could write a small test to make sure the behavior for this footer is always functional. :)

Oh I'd love to do that Yip!
Would it make sense to just test the function `wget_footer` by giving it inputs and seeing if it matches a certain output? That's how I'd do it, but I'd consult a testing manual just in case!

Also I couldn't add the tests right now, should I add them in a separate PR, add there here later (squashed or unsquashed?), or for now not at all?

svenstaro:
You could also add the tests in this MR. Don't worry, I'll wait. I just want to release something this week ideally. 

I think in this case a unit test of this particular function in the same file makes sense as well as an integration test. we have plenty of examples for both so it shouldn't be that hard. :)

Yusuto:
Yip, sounds gud, Thanks! I'll get to it either later today or tomorrow morning :sunny:

Yusuto:
I'm done I think!

- [x] Fixed footer
- [x] Made footer click to copy (I didn't pay attention to accessibility though)
- [x] Added some fancy improvements
- [x] Added and updated tests

If any issues arise I'll get to them tomorrow, but if nawt then *I think* this MR is ready to merge!

src/listing.rs
```diff
@@ -163,12 +163,12 @@ pub fn directory_listing(
...
-        req.connection_info().scheme(),
-        req.connection_info().host(),
-        req.uri()
-    );
+    let abs_uri = http::Uri::builder()
+        .scheme(req.connection_info().scheme())
+        .authority(req.connection_info().host())
+        .path_and_query(req.uri().to_string())
+        .build()
+        .unwrap();
```
svenstaro, line 171: Yup, this is much more sane.

src/renderer.rs
```diff
@@ -683,3 +695,54 @@ pub fn render_error(
...
+                r#"<div class="downloadDirectory"><p>Download folder:</p><a class="cmd" title="Click to copy!" style="cursor: pointer;" onclick="navigator.clipboard.writeText(&quot;wget -rcnHp -R 'index.html*' {0}/?raw=true'&quot;)">wget -rcnHp -R 'index.html*' {0}/?raw=true'</a></div>"#,
+                x
+            )
+        };
+        let uri = |x| Uri::try_from(x).unwrap();
+
+        let to_be_tested: String = wget_footer(
+            &uri("https://github.com/svenstaro/miniserve/"),
+            Some("Miniserve"),
+            None,
```
svenstaro, line 716: Why not make all of these test cases into their own unit tests? That way, you can also make it self-documenting by varying the test names.

svenstaro (changes requested):

src/renderer.rs
```diff
@@ -267,26 +268,37 @@ fn version_footer() -> Markup {
...
-fn wget_footer(title_path: &str, current_user: Option<&CurrentUser>) -> Markup {
-    let count = {
-        let count_slashes = title_path.matches('/').count();
-        if count_slashes > 0 {
-            count_slashes - 1
-        } else {
-            0
-        }
+fn wget_footer(abs_path: &Uri, root_dir_name: Option<&str>, current_user: Option<&str>) -> Markup {
+    let escape_apostrophes = |x: &str| x.replace('\'', "'\"'\"'");
```
svenstaro, line 272: Let's not do the JS fat arrow function thing here. Why not just make this a proper little inner function instead?

src/renderer.rs
```diff
@@ -267,26 +268,37 @@ fn version_footer() -> Markup {
...
+
+    // Directory depth, 0 is root directory
+    let cut_dirs = match abs_path.path().matches('/').count() - 1 {
+        // Put all the files in a folder of this name
+        0 => format!(
+            " -P '{}'",
+            escape_apostrophes(
```
svenstaro, line 279: How do you run into the case that this is required anyway?

svenstaro:
Hey, do you plan on fixing up the rest soonish? I'd like to do a release. No worries if you need longer but then it might not make the release.

Yusuto:
> Hey, do you plan on fixing up the rest soonish? I'd like to do a release. No worries if you need longer but then it might not make the release.

Oh, I'm not exactly sure what is left to be fixed, it's only the tests, right?

I wrote a comment earlier that I'm very unsure about how to split them apart, whatever approach I try seems to be just kind of not really optimal, or at least not significantly better than just lumping it all in one function and if something fails manually figuring out what it is, so I'm a bit at loss of how to continue there I'm afraid

svenstaro:
Merging as is, thanks! I split the tests the way I imagined it.

Yusuto:
Oké, Thank you! This is my first MR I ever made, I'm very happy to get it accepted ^^

svenstaro:
Thanks for the contribution :) I'd be happy for more contributions of course.

Yusuto:
No problem!
I will definitely consider it, Yeh! More advanced uploading (like uploading folders and seeing progress) would be pretty heccin practical I think for example! Although I'm also very interested in media processing and am currently looking into that, I'll also look at issues if I get the time ^^

PR #1093, opened 2023-04-03
File & directory deletion feature

### LINKED ISSUE #397: How about add "delete file" feature?
After I do some research on many mini http server, miniserve is a great work, which have authentication. But I also have found many mini http servers did not have "delete file" feature including miniserve.

I think "delete file" will be a good point to a mini http server, especially it can combine the authntication.

cyqsimon:
Closes #397.

## Todo
- [x] CLI options
- [x] Backend API
- [x] Frontend controls
- [x] Tests

### DIFF
--- a/data/style.scss
+++ b/data/style.scss
@@ -252,6 +252,10 @@ table thead th.date {
     width: 21em;
 }
 
+table thead th.actions {
+    width: 4em;
+}
+
 table tbody tr:nth-child(odd) {
     background: var(--odd_row_background);
 }
@@ -275,6 +279,22 @@ td.size-cell {
 td.date-cell {
     display: flex;
     justify-content: space-between;
+}
+
+td.actions-cell button {
+    padding: 0.1rem 0.3rem;
+    border-radius: 0.2rem;
+    border: none;
+}
+
+td.actions-cell .rm_form {
+    display: flex;
+    place-content: center;
+}
+
+td.actions-cell .rm_form button {
+    background: var(--rm_button_background);
+    color: var(--rm_button_text_color);
 }
 
 .history {
--- a/tests/fixtures/mod.rs
+++ b/tests/fixtures/mod.rs
@@ -42,6 +42,11 @@ pub static DIRECTORIES: &[&str] = &["dira/", "dirb/", "dirc/"];
 #[allow(dead_code)]
 pub static HIDDEN_DIRECTORIES: &[&str] = &[".hidden_dir1/", ".hidden_dir2/"];
 
+/// Files nested at different levels under the same root directory
+#[allow(dead_code)]
+pub static NESTED_FILES_UNDER_SINGLE_ROOT: &[&str] =
+    &["someDir/alpha", "someDir/some_sub_dir/bravo"];
+
 /// Name of a deeply nested file
 #[allow(dead_code)]
 pub static DEEPLY_NESTED_FILE: &str = "very/deeply/nested/test.rs";
@@ -72,10 +77,14 @@ pub fn tmpdir() -> TempDir {
         }
     }
 
-    tmpdir
-        .child(DEEPLY_NESTED_FILE)
-        .write_str("File in a deeply nested directory.")
-        .expect("Couldn't write to file");
+    let mut nested_files = NESTED_FILES_UNDER_SINGLE_ROOT.to_vec();
+    nested_files.push(DEEPLY_NESTED_FILE);
+    for file in nested_files {
+        tmpdir
+            .child(file)
+            .write_str("File in a deeply nested directory.")
+            .expect("Couldn't write to file");
+    }
     tmpdir
 }
 
--- a/tests/fixtures/mod.rs
+++ b/tests/fixtures/mod.rs
@@ -4,7 +4,9 @@ use assert_fs::prelude::*;
 use port_check::free_local_port;
 use reqwest::Url;
 use rstest::fixture;
+use std::path::Path;
 use std::process::{Child, Command, Stdio};
+use std::sync::LazyLock;
 use std::thread::sleep;
 use std::time::{Duration, Instant};
 
@@ -43,13 +45,24 @@ pub static DIRECTORIES: &[&str] = &["dira/", "dirb/", "dirc/"];
 pub static HIDDEN_DIRECTORIES: &[&str] = &[".hidden_dir1/", ".hidden_dir2/"];
 
 /// Files nested at different levels under the same root directory
+///
+/// This is not a `&[&str]` because in percent-encoding, path segments and full paths
+/// are encoded differently.
 #[allow(dead_code)]
-pub static NESTED_FILES_UNDER_SINGLE_ROOT: &[&str] =
-    &["someDir/alpha", "someDir/some_sub_dir/bravo"];
+pub static NESTED_FILES_UNDER_SINGLE_ROOT: LazyLock<Vec<&Path>> = LazyLock::new(|| {
+    vec![
+        Path::new("someDir/alpha"),
+        Path::new("someDir/some_sub_dir/bravo"),
+    ]
+});
 
-/// Name of a deeply nested file
+/// Path to a deeply nested file
+///
+/// This is not a `&str` because in percent-encoding, path segments and full paths
+/// are encoded differently.
 #[allow(dead_code)]
-pub static DEEPLY_NESTED_FILE: &str = "very/deeply/nested/test.rs";
+pub static DEEPLY_NESTED_FILE: LazyLock<&Path> =
+    LazyLock::new(|| Path::new("very/deeply/nested/test.rs"));
 
 /// Test fixture which creates a temporary directory with a few files and directories inside.
 /// The directories also contain files.
@@ -77,8 +90,8 @@ pub fn tmpdir() -> TempDir {
         }
     }
 
-    let mut nested_files = NESTED_FILES_UNDER_SINGLE_ROOT.to_vec();
-    nested_files.push(DEEPLY_NESTED_FILE);
+    let mut nested_files = NESTED_FILES_UNDER_SINGLE_ROOT.clone();
+    nested_files.push(&DEEPLY_NESTED_FILE);
     for file in nested_files {
         tmpdir
             .child(file)
--- a/tests/rm_files.rs
+++ b/tests/rm_files.rs
@@ -3,10 +3,14 @@ mod fixtures;
 use anyhow::bail;
 use assert_fs::fixture::TempDir;
 use fixtures::{server, server_no_stderr, tmpdir, TestServer};
+use percent_encoding::utf8_percent_encode;
 use reqwest::blocking::Client;
 use reqwest::StatusCode;
 use rstest::rstest;
-use std::path::Path;
+use std::{
+    iter,
+    path::{Component, Path},
+};
 use url::Url;
 
 use crate::fixtures::{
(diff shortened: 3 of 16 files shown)


cyqsimon:
https://github.com/svenstaro/miniserve/blob/82dd82615ade9265630a178165c66002b0b9314c/src/main.rs#L329-L330

I am curious why the upload feature is implemented like so. I guess it's good for unforeseen future extensibility, but there is semantic duplication between the `/upload` endpoint and the `POST` method.

The real question I want to ask, is should I implement the deletion feature in a similar manner? I.e. something like this:

```rust
app.service(web::resource("/delete").route(web::delete().to(file_rm::rm_file)));
```

svenstaro:
Are you suggesting it might have been better to just have `POST` always try to upload stuff? It would be simpler for sure but as you are saying, it's not great if we ever want a second `POST` for something. I think this is currently an ok compromise.

I see how this is a bit of a conundrum with delete. I also want to add WebDAV at some point and that will introduce even more semantics. Perhaps it's the best to go with what you suggested just now and eat the semantic duplication for the time being. It's not like it can't be changed later.

bgusach:
Hi,

unfortunately I can't directly help... just wanted to say that I appreciate the effort and if it gets done at some point, my life would be somehow better :)

cyqsimon:
> Hi,
> 
> unfortunately I can't directly help... just wanted to say that I appreciate the effort and if it gets done at some point, my life would be somehow better :)

Thanks for reminding me. I have too many things on hand and honestly just forgot about this whole matter. I will find some time to work on it this week, hopefully.

data/style.scss
```diff
@@ -269,6 +273,22 @@ td.date-cell {
...
+
+td.actions-cell .rm_form {
+    display: flex;
+    place-content: center;
+}
+
+td.actions-cell .rm_form button {
+    background: var(--rm_button_background);
+    color: var(--rm_button_text_color);
+}
```
cyqsimon, line 290: I suck at visual design. So any styling improvement suggestion is welcomed.

cyqsimon:
Rebased.

cyqsimon:
Rebased on `master`.

cyqsimon:
Rebased on `master`.

svenstaro:
Could you add some tests? Since this is going to be deleting things, it needs some proper tests.

cyqsimon:
> Could you add some tests? Since this is going to be deleting things, it needs some proper tests.

Yeah I'm working on that on and off. The tests are actually written but several are failing. My original intention was to combine the tests and the fixes into one commit, but I guess there's no harm in splitting them up. I'll push the tests first and then look to fix them.

cyqsimon:
If you can spare some time, please take a look at the test suite; make sure they are testing the correct behaviour. Thanks.

svenstaro:
> Yeah I'm working on that on and off. The tests are actually written but several are failing. My original intention was to combine the tests and the fixes into one commit, but I guess there's no harm in splitting them up. I'll push the tests first and then look to fix them.

No worries. There's no rush to get this in. Just take your time and make sure things look good.

> If you can spare some time, please take a look at the test suite; make sure they are testing the correct behaviour. Thanks.

I will try to squeeze a review in but no promises, I'm fairly low on time currently.

cyqsimon:
Rebased on #1454.

cyqsimon:
This should be ready now, finally. Everything works as expected, as far as I can tell. But you may wish to merge the other PR first.

tests/fixtures/mod.rs
```diff
@@ -4,7 +4,9 @@ use assert_fs::prelude::*;
 use port_check::free_local_port;
 use reqwest::Url;
 use rstest::fixture;
+use std::path::Path;
 use std::process::{Child, Command, Stdio};
+use std::sync::LazyLock;
```
cyqsimon, line 9: The use of `LazyLock` does raise the MSRV to 1.80, if you want to run tests. But miniserve doesn't have a MSRV policy so I thought it's probably fine?

If you don't like this, I can always swap in `once_cell::sync::Lazy`.

svenstaro: I'm totally cool with you using modern Rust. :)

svenstaro:
Needs a rebase and then I will do a proper review. :)

cyqsimon:
Actually, let me do an interactive rebase to reorganise things. Then it'd be a bit tidier.

qtekk:
hey, how's the feature going?

cyqsimon:
> hey, how's the feature going?

Implementation is complete. Just waiting on review.

@svenstaro You got time to take a look?

svenstaro:
I will take some time. Might take a look on the weekend.

svenstaro:
Thoughts:
- Do we want any kind of confirmation? Currently you might miss-click and a file is just gone.
- Do we want to warn in case a path given to `-R` is not contained in the served paths?

cyqsimon:
Good points.

>     * Do we want any kind of confirmation? Currently you might miss-click and a file is just gone.

That would be nice. I would also want a toggle/checkbox that allows turning off confirmation, in case where you want to delete several things en-mass.

>     * Do we want to warn in case a path given to `-R` is not contained in the served paths? For instance, `miniserve -R /tmp /some/dir` doesn't do anything.

I think there's some misunderstanding. The paths accepted by `-u` and `-R` are actually URL paths relative to `/`, not file system paths relative to CWD. This means:

1. the specified directory should never escape the serve root, excepting symlink shenanigans (all covered by tests)
2. all paths are valid because new directories can be created on the fly; even something like `-R ../..` is legal because navigating up the serve root is prevented, so effectively it's the same as `-R .`
3. `-R /foo/bar` and `-R foo/bar` are equivalent

So I believe the implementation needs no changes, but it would be prudent to note this in the CLI help document though. I'll make that happen. Also it might be a good idea to log a message at `INFO` level noting this behaviour.

>     * Should we abort if we provide a `path` to a file but also `-R`? We don't do this with `-u` but it might make sense as a followup? So for instance, `miniserve -R /tmp /some/file`.

Oh I actually didn't know we support serving a single file. That's a nice feature.

Hmm yeah this is very interesting indeed. Because the semantics of `-u` and `-R` aren't really compatible with this mode of operation.

But first of all, should we allow removing (or for `-u`, overwriting) in this mode at all? Personally I believe the answer is yes, but I'd like to hear your opinions too.

And if the answer to the previous question is yes, obviously `-u` and `-R` with arguments would make no sense (since a file cannot have subdirectories). Should this be a warning or a hard error?

cyqsimon:
For points 2 and 3, the changes would apply to both `-u` and `-R`. So I think it's better organisationally if these form a separate PR, queued after this one.

tests/rm_files.rs
```diff
@@ -0,0 +1,234 @@
...
+
+#[rstest]
+fn rm_disabled_by_default(server: TestServer) -> anyhow::Result<()> {
+    let paths = [FILES, DIRECTORIES]
+        .concat()
+        .into_iter()
+        .map(Path::new)
+        .chain(iter::once(DEEPLY_NESTED_FILE.as_ref()))
+        .collect::<Vec<_>>();
+    assert_rm_err(server.url(), &paths, true)
```
svenstaro, line 121: Hm, for all of these days I wonder whether it wouldn't be nicer and more obvious to use rstest more in the way that it was intended and have test-cases generated? Currently all paths are tested in the same test which is not really the intention of rstest. Maybe take a look at how we do things in `upload_files.rs` and try to follow that?

svenstaro (changes requested):
Review done. I think we're almost there. :)

svenstaro:
Excuse the conflicts, should be fairly easy to rebase, though.

svenstaro:
> Good points.
> 
> > * Do we want any kind of confirmation? Currently you might miss-click and a file is just gone.
> 
> That would be nice. I would also want a toggle/checkbox that allows turning off confirmation, in case where you want to delete several things en-mass.
> 
> I'll give this a shot, but I'm not great at front end coding. We'll see.

Yeah, I think that would be safer and probably also more convenient. I can't think of many cases where I want to just click something and it'll be deleted right then and there without confirmation. I suspect users have also grown to expect confirmations for deletions.

> 
> > * Do we want to warn in case a path given to `-R` is not contained in the served paths? For instance, `miniserve -R /tmp /some/dir` doesn't do anything.
> 
> I think there's some misunderstanding. The paths accepted by `-u` and `-R` are actually URL paths relative to `/`, not file system paths relative to CWD. This means:
> 
>     1. the specified directory should never escape the serve root, excepting symlink shenanigans (all covered by tests)
> 
>     2. all paths are valid because new directories can be created on the fly; even something like `-R ../..` is legal because navigating up the serve root is prevented, so effectively it's the same as `-R .`
> 
>     3. `-R /foo/bar` and `-R foo/bar` are equivalent
> 
> 
> So I believe the implementation needs no changes, but it would be prudent to note this in the CLI help document though. I'll make that happen. Also it might be a good idea to log a message at `INFO` level noting this behaviour.

You are correct, of course. I didn't consider that the paths can become valid during runtime. I think documenting this more clearly would make sense for sure as it had even me confused :D

> 
> > * Should we abort if we provide a `path` to a file but also `-R`? We don't do this with `-u` but it might make sense as a followup? So for instance, `miniserve -R /tmp /some/file`.
> 
> Oh I actually didn't know we support serving a single file. That's a nice feature.
> 
> Hmm yeah this is very interesting indeed. Because the semantics of `-u` and `-R` aren't really compatible with this mode of operation.
> 
> But first of all, should we allow removing (or for `-u`, overwriting) in this mode at all? Personally I believe the answer is yes, but I'd like to hear your opinions too.

I'm not really sure it makes sense. How would we even present that to the user? Currently serving a single file just gives you that file directly without any UI. I'm not sure supporting `-u` or `-R` would make any sense outside of an API-esque usage scenario.

> 
> And if the answer to the previous question is yes, obviously `-u` and `-R` with arguments would make no sense (since a file cannot have subdirectories). Should this be a warning or a hard error?

svenstaro:
> For points 2 and 3, the changes would apply to both `-u` and `-R`. So I think it's better organisationally if these form a separate PR, queued after this one.

Agreed, no need to do it in this one.

svenstaro:
@cyqsimon wanna push this over the finish line?

cyqsimon:
> @cyqsimon wanna push this over the finish line?

Yeah but I'm on vacation at the moment and connectivity is shoddy. Thanks for the reminder though, will try to find some time for this.

svenstaro:
Any chance you could take a look? We're sooo close. :)

cyqsimon:
> Any chance you could take a look? We're sooo close. :)

Yeah okay tomorrow maybe. Or maybe over the weekend.

Corporate toil is getting the better of me =(

svenstaro:
> > Any chance you could take a look? We're sooo close. :)
> 
> Yeah okay tomorrow maybe. Or maybe over the weekend.
> 
> Corporate toil is getting the better of me =(

I know the feeling bro. Get our before you're burnt out.

svenstaro:
Friendly ping :)

Karotte128:
Any updates on this PR? I really like this feature...

svenstaro:
@cyqsimon I just tagged v0.30.0 but I'd be happy to include this in the next release if you can find some time to work on it. :)

NOBLES5E:
more than 2 years passed ...

NOBLES5E:
@svenstaro Since @cyqsimon seems to be occupied with other commitments recently, would you accept if I take over this PR, address the review feedback, and get it merged ASAP? This feature has been in development for over 2 years and there seems to be significant community interest in having file deletion functionality available.

svenstaro:
I think at this point it would be reasonable to have somebody else take this over, yes. Feel free to grab it.

NOBLES5E:
@svenstaro great, here you go https://github.com/svenstaro/miniserve/pull/1518

PR #1160, opened 2023-07-05
Allow parameters to be provided via env vars

### LINKED ISSUE #829: Configuration with Environment variables
It would be really nice, to configurate miniserve with env variables.

(In k8s it is possible to store Password safe in "Secrets" and use them as envFrom with secretRef - so basic auth credentials could be stored secure.)

svenstaro:
Fixes #829.


onedr0p (approved):
LGTM!!

PR #1169, opened 2023-07-12
simplify theme selection using data attributes

baod-rate:
Instead of complex combinations of class-based selectors, use a data-attribute to select the color scheme. Also simplifies the default-theme case by actually providing a default CSS rule that always applies.  These changes simplify the (S)CSS a lot, but had to add a bit of rust code to build the final stylesheet.


src/renderer.rs
```diff
@@ -56,26 +56,24 @@ pub fn page(
             (page_header(&title_path, conf.file_upload, &conf.favicon_route, &conf.css_route))
 
             body #drop-container
-                .(format!("default_theme_{}", conf.default_color_scheme))
-                .(format!("default_theme_dark_{}", conf.default_color_scheme_dark)) {
+                // .(format!("default_theme_{}", conf.default_color_scheme))
+                // .(format!("default_theme_dark_{}", conf.default_color_scheme_dark)) {
+            {
```
svenstaro, line 61: What's with this?

svenstaro (changes requested):
Hey, cool idea and I think it makes sense. Just a small nit.

src/renderer.rs
baod-rate: Whoops. This change means the default theme is encoded directly in the .css and so we don't need to set a class on the body. I forgot to delete these lines

svenstaro:
Alright I'm going to give you some more time to fix up up so you like it. :)

baod-rate:
It turns out what I was thinking of isn't possible. But I switched to using `grass::include!` and removed the `build.rs`, so there's at least one less place you have to list out the themes (back down to two)

src/renderer.rs
```diff
@@ -667,8 +679,8 @@ pub fn render_error(
                     <script>
                         // read theme from local storage and apply it to body
                         var theme = localStorage.getItem('theme');
-                        if (theme != null && theme != 'default') {
-                            document.body.classList.add('theme_' + theme);
+                        if (theme && theme != "default") {
+                            body.setAttribute("data-theme", name)
                         }
```
svenstaro, line 684: Isn't this what `updateColorScheme()` does already?

src/renderer.rs
```diff
@@ -56,26 +56,22 @@ pub fn page(
...
                         var theme = localStorage.getItem('theme');
-
-                        if (theme != null && theme != 'default') {
-                            body.classList.add('theme_' + theme);
-                        }
+                        updateColorScheme(theme)
 
                         // updates the color scheme by replacing the appropriate class
                         // on body and saving the new theme to local storage
                         function updateColorScheme(name) {
```
svenstaro, line 70: Might better to hoist this function up to make it easier to read.

data/style.scss
```diff
@@ -215,13 +204,7 @@ nav .theme li a:hover {
...
 }
 
-@each $theme in $themes {
-    body.theme_#{$theme} nav .theme li.theme_#{$theme} a {
-        @extend %active_theme_link;
-    }
-}
-
-#{body_not_themed()} nav .theme li.theme_default a {
+body:not([data-theme]) nav .theme li[data-theme="default"] a {
```
svenstaro, line 207: Please document what this does.

svenstaro (changes requested):
Had one more look through it. Small set of requests but almost good to go! :)

src/renderer.rs
baod-rate: Yup. I just figured there was probably a good reason the script wasn't in the header so both the content and error pages could use it. I didn't realize these changes broke the error page theming so this needed fixing anyways

baod-rate:
Moved all the javascript into the `head` to avoid duplicate code between the error and content pages. Also added a minor improvement: `updateColorScheme` is now a listener for the [`storage` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event) so theme changes are synchronized across tabs (:

Also, I *think* I fixed an issue with `error_page_middleware`. While testing the error pages, it seemed like the check to trigger `map_error_page` would never succeed because the content-type header (`text/plain; charset=utf-8`) would never equal `mime::TEXT_PLAIN_UTF_8.essence_str()` (i.e. `text/plain`).

svenstaro (approved):
Good job!

Issue #1190, opened 2023-08-01
Wget Footer Command Fails When Directory Contains Apostrophe's in the Name


d-air1:
Example folder title "The Rabbit's Foot"

Generated wget command 
`wget -rcnHp -R 'index.html*' 'http://localhost:9991/The%20Rabbit's%20Foot/?raw=true'`

The issue is here is that the apostrophe after the word `Rabbit` acts as a closing quote. This makes the apostrophe after the word `true` a dangling starting quote.

Perhaps switch to double quotes for the outside of the command?

d-air1:
I don't mind taking a stab at it, but I'm not really familiar with rust.

d-air1:
Closing since the PR was merged.

PR #1191, opened 2023-08-02
Encode single quote in directory path name

d-air1:
Well I took a stab at it. Took me a minute just to be able to read rust code, since I don't have any experience with rust. However, with just a couple line changes I did get it to work. Not sure if this is how you would have liked it to have been done though.

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -267,8 +267,9 @@ fn wget_footer(abs_path: &Uri, root_dir_name: Option<&str>, current_user: Option
         None => String::new(),
     };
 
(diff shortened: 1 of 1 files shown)


d-air1:
This is about issue #1190

svenstaro:
I'll merge after you fix the formatting issue.

d-air1:
I'm assuming that running `cargo fmt` was the formatting issue you were speaking of? If so, I have just done so.

d-air1:
I see that one of the CI tests failed and gave a useful error message, so I went ahead and made that change.

PR #1193, opened 2023-08-06
Add pretty urls

nlopes:
This adds a new flag namely `--pretty-urls` that when enabled will serve the equivalent `.html` if it exists.

Very much the same approach that [`netlify`
uses](https://docs.netlify.com/site-deploys/post-processing/).

It can be quite useful when having hrefs like `/about` serve `/about.html`.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -41,6 +41,14 @@ pub struct CliArgs {
     #[arg(long, requires = "index", env = "MINISERVE_SPA")]
     pub spa: bool,
 
+    /// Activate Pretty URLs mode
+    ///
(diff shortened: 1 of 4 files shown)


nlopes:
Hi 👋 

I don't believe this was a feature ever requested or in the issue tracker (that I was able to find) but it's one that I use myself.
Happy to keep this as a branch on my own but I thought it would be a nice little feature for `miniserve`.

Feel free to completely discard if you don't think it is something you want to support.

svenstaro (changes requested):
Seems like a cool feature! Good code too, just small nits.

svenstaro (approved):

svenstaro:
Good work!

Issue #1236, opened 2023-09-21
[feature request] Show exact byte size of files in some way


oldherl:
Currently only the "human readable" size of files is shown in the listing, e.g. 2.0 GB
However sometimes I want to know the exact byte size (e.g. 2000111222 bytes) of the file.

Would like to see the exact byte size as a hovering tooltip, or replace the human readable sizes with the exact size on some command line switches.

svenstaro:
Sounds like a good idea. Would you like to try to implement it?

oldherl:
I have no experience with Rust, but I might be able to do that when I have some time to spare. Don't expect any time soon, though.

oldherl:
> is that what you want? @oldherl 

Yes. Is that your fork?

Lzzzzzt:
> > is that what you want? @oldherl
> 
> Yes. Is that your fork?

yes, I'm about to submit a PR.

svenstaro:
This is now implemented!

PR #1261, opened 2023-10-28
Feature: Show size in byte

Lzzzzzt:
Add a cli option that show file size in bytes rather than KB or MB, etc.
solve  the #1236.

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -150,7 +150,7 @@ pub fn page(
                                 }
                             }
                             @for entry in entries {
-                                (entry_row(entry, sort_method, sort_order, false, conf.show_size_in_byte))
+                                (entry_row(entry, sort_method, sort_order, false, conf.show_exact_bytes))
                             }
                         }
                     }
@@ -238,7 +238,7 @@ pub fn raw(entries: Vec<Entry>, is_root: bool, conf: &MiniserveConfig) -> Markup
                             }
                         }
                         @for entry in entries {
-                            (entry_row(entry, None, None, true, conf.show_size_in_byte))
+                            (entry_row(entry, None, None, true, conf.show_exact_bytes))
                         }
                     }
                 }
@@ -522,7 +522,7 @@ fn entry_row(
     sort_method: Option<SortingMethod>,
     sort_order: Option<SortingOrder>,
     raw: bool,
-    show_size_in_byte: bool,
+    show_exact_bytes: bool,
 ) -> Markup {
     html! {
         tr {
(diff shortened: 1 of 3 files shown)


svenstaro:
Sorry for ignoring this for so long! Would you perhaps like to rebase this? I promise I'll review it quickly. :)

Lzzzzzt:
> Sorry for ignoring this for so long! Would you perhaps like to rebase this? I promise I'll review it quickly. :)

OK

Lzzzzzt:
> Sorry for ignoring this for so long! Would you perhaps like to rebase this? I promise I'll review it quickly. :)

I'm done

svenstaro:
Ok, so I'm thinking: Perhaps this should be a frontend option like the sorting method and sorting order? It would be consistent since it's a display thing like those other options. There could be a button in the "Size" row that would allow you to toggle this (and it'd change the query parameter).

We can still have something like `--default-size-display` which is either `human` or `exact`. What do you think about that?

Lzzzzzt:
> Ok, so I'm thinking: Perhaps this should be a frontend option like the sorting method and sorting order? It would be consistent since it's a display thing like those other options. There could be a button in the "Size" row that would allow you to toggle this (and it'd change the query parameter).
> 
> We can still have something like `--default-size-display` which is either `human` or `exact`. What do you think about that?

Agreed! Keeping it consistent with sorting options and adding a toggle in the “Size” row makes sense. I’m good with the --default-size-display flag too.

Lzzzzzt:
> There could be a button in the "Size" row that would allow you to toggle this (and it'd change the query parameter).

It seems there is no suitable place for the button in the table header, I think the `Size` string in the entry could be a button to change the display behavior. And this will change other entries' display behavior, not only the clicked one.

svenstaro:
Do you want to rebase this and for the time being just exchange `--show_size_in_byte` for `--size-display` with an `enum` that is either `human` or `exact`? I'd like to get this into the next release if possible. Let's ignore the frontend stuff for the time being.

Lzzzzzt:
> Do you want to rebase this and for the time being just exchange `--show_size_in_byte` for `--size-display` with an `enum` that is either `human` or `exact`? I'd like to get this into the next release if possible. Let's ignore the frontend stuff for the time being.

Ok

Lzzzzzt:
> Do you want to rebase this and for the time being just exchange `--show_size_in_byte` for `--size-display` with an `enum` that is either `human` or `exact`? I'd like to get this into the next release if possible. Let's ignore the frontend stuff for the time being.

I've rebased and changed cli args, could you review the code?

src/renderer.rs
```diff
@@ -554,13 +555,19 @@ fn entry_row(
 
                         @if !raw {
                             @if let Some(size) = entry.size {
-                                span.mobile-info.size {
-                                    (build_link("size", &format!("{}", size), sort_method, sort_order))
+                                @if show_exact_bytes {
+                                    span.mobile-info.size {
+                                        (maud::display(format!("{}B", size.as_u64())))
```
svenstaro, line 560: Shall we display it the same as the human format with `{} B` instead of `{}B`?

PR #1285, opened 2023-12-06
Prevent illegal request path from crashing program

cyqsimon:
I experienced a service crash today due to a request that contains an illegal character in its path. The request seems to be a malicious vulnerability scan, but it shouldn't crash the program nonetheless. This PR makes sure this doesn't happen.

Logs:

```
Dec 06 04:07:59 HOSTNAME miniserve[1012]: thread 'actix-rt|system:0|arbiter:1' panicked at 'called `Result::unwrap()` on an `Err` value: http::Error(InvalidUri(InvalidUriChar))', /build/path/miniserve-0.24.0/src/listing.rs:171:10
```

### DIFF
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -162,12 +162,17 @@ pub fn directory_listing(
 
     let base = Path::new(serve_path);
     let random_route_abs = format!("/{}", conf.route_prefix);
-    let abs_uri = http::Uri::builder()
-        .scheme(req.connection_info().scheme())
-        .authority(req.connection_info().host())
-        .path_and_query(req.uri().to_string())
-        .build()
-        .unwrap();
+    let abs_uri = {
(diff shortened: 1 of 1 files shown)


svenstaro:
Thanks! Could you perhaps also add a test for this? For extra points, proptest would be perfect for finding stuff such as this!

cyqsimon:
Ugh. That's not going to be trivial because I don't actually know what the offending request was. And going the distance to learn and add a new testing framework needs more time than I'm willing to commit. Sorry ¯\\\_(ツ)\_\/¯.

And to be honest I find it a bit overkill to test for what amounts to an "avoid using unwrap" anyways.

svenstaro:
Ok don't worry. I'll add the test case.

PR #1290, opened 2023-12-11
Remove header Content-Encoding when archiving

### LINKED ISSUE #1187: .tar.gz download is broken and only downloads uncompressed tar
When using `--enable-tar-gz` and clicking the "Download .tar.gz" it still only downloads a uncompressed tar file. It has a `.tar.gz` file extension, but trying to open the `.tar.gz` fails, because it's actually not compressed, so uncompression fails (renaming to `.tar` or opening it otherwise without trying to uncompress it works). The checksum of the downloaded file is exactly the same as when using `--enable-tar` and clicking the "Download .tar" button, so it is indeed just a normal `.tar`.

Version: 0.24.0
OS: Gentoo Linux
Browser: Tested in Firefox and Chromium, so it doesn't look like it's a browser-issue

5long:
Content-Encoding is a representation header which kinda means "same content, presented differently to different clients" or "encoded & decoded on-the-fly, guided by content negotiation".

In the case of downloading an archive, MDN docs explicitly says that (quoted from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding):

> If the original media is encoded in some way (e.g. a zip file) then this
> information would not be included in the Content-Encoding header.

Thus this patch. Also fixes #1187.

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -2,7 +2,6 @@ use std::fs::File;
 use std::io::{Cursor, Read, Write};
 use std::path::{Path, PathBuf};
 
-use actix_web::http::header::ContentEncoding;
 use libflate::gzip::Encoder;
 use serde::Deserialize;
 use strum::{Display, EnumIter, EnumString};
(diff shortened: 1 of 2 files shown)


svenstaro:
Seems good to me, thanks!

PR #1308, opened 2024-01-09
Set default sorting order and method with arguments

ElliottLandsborough:
My intention is to add a way to sort by `descending` `date` (or anything else) by default.

I think the best way to do this is with a command line argument but I am open to suggestions.

This is useful for anyone who wants the most recently created/edited files at the top of the list when the page first loads.

Without this change, a user would have to do an _entire whole extra single click_ to sort the files by date 😢

### DIFF
--- a/tests/navigation.rs
+++ b/tests/navigation.rs
@@ -1,12 +1,13 @@
 mod fixtures;
 mod utils;
 
-use fixtures::{server, Error, TestServer, DEEPLY_NESTED_FILE, DIRECTORIES};
+use fixtures::{server, Error, TestServer, FILES, DEEPLY_NESTED_FILE, DIRECTORIES};
 use pretty_assertions::{assert_eq, assert_ne};
 use rstest::rstest;
 use select::document::Document;
 use std::process::{Command, Stdio};
 use utils::get_link_from_text;
+use utils::get_link_hrefs_from_text_with_prefix;
 
 #[rstest(
     input,
@@ -144,6 +145,33 @@ fn can_navigate_using_breadcrumbs(
     // current dir is not linked
     let current_dir_link = get_link_from_text(&parsed, "nested");
     assert_eq!(None, current_dir_link);
+
+    Ok(())
+}
+
+#[rstest]
+#[case(server(&["--default-sorting-method", "date", "--default-sorting-order", "desc"]))]
+#[case(server(&["--default-sorting-method", "date", "--default-sorting-order", "asc"]))]
+/// We can specify the default sorting order
+fn can_specify_default_sorting_order(#[case] server: TestServer) -> Result<(), Error> {
+    let slash = String::from("/");
+    let base_url = server.url();
+    let nested_url = base_url.join(&slash)?;
+
+    let resp = reqwest::blocking::get(nested_url.as_str())?;
+    let body = resp.error_for_status()?;
+    let parsed = Document::from_read(body)?;
+
+    let links = get_link_hrefs_from_text_with_prefix(&parsed, "/");
+    let first_created_file = slash + FILES.first().unwrap();
+
+    if links.first().unwrap() == &first_created_file {
+        assert_eq!("/very/?sort=date&order=asc", links.last().unwrap());
+    }
+
+    if links.last().unwrap() == &first_created_file {
+        assert_eq!("/very/?sort=date&order=desc", links.first().unwrap());
+    }
 
     Ok(())
 }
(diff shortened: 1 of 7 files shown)


svenstaro:
Cool! Could you add a description for what it is supposed to do and why? For example, why aren't the current sorters not sufficient? I suppose the idea is to provide a sorting default?

ElliottLandsborough:
> Cool! Could you add a description for what it is supposed to do and why? For example, why aren't the current sorters not sufficient? I suppose the idea is to provide a sorting default?

I have updated the description - you are right, I want to be able to set the default sort type and order.

The chevrons _may_ have been pointing in the wrong direction too so I have included the change. This is not a big deal to me and it is probably open to interpretation.

svenstaro:
Code looks perfect. Can I somehow convince you to write a test for this?

ElliottLandsborough:
> Code looks perfect. Can I somehow convince you to write a test for this?

I've added a test, fixed some bugs, updated some comments.

tests/navigation.rs
```diff
@@ -147,3 +147,33 @@ fn can_navigate_using_breadcrumbs(
 
     Ok(())
 }
+
+#[rstest]
+#[case(server(&["--default-sorting-method", "date", "--default-sorting-order", "desc"]))]
```
svenstaro, line 152: Can you add a second case where you order the other way and demonstrate that this works?

svenstaro (changes requested):
Looks great, just one comment about the tests.

tests/navigation.rs
```diff
@@ -147,3 +147,33 @@ fn can_navigate_using_breadcrumbs(
...
+
+    let resp = reqwest::blocking::get(nested_url.as_str())?;
+    let body = resp.error_for_status()?;
+    let parsed = Document::from_read(body)?;
+
+    // can go to intermediate dir with specified sort
+    let intermediate_dir_link =
+        get_link_from_text(&parsed, "very").expect("Intermediate dir link not found.");
+    assert_eq!("/very/?sort=date&order=desc", intermediate_dir_link);
+
```
svenstaro, line 177: Wouldn't it be better to have a test that checks whether the default sorting works as expected? Sort it one way and then compare it against an expected order and then sort it another way and expect the order to be reversed.

svenstaro:
I'm happy to merge this and release a new version of miniserve once the test is a bit more useful.

ElliottLandsborough:
> I'm happy to merge this and release a new version of miniserve once the test is a bit more useful.

Tests have been updated 😃

tests/utils/mod.rs
```diff
@@ -12,3 +12,19 @@ pub fn get_link_from_text(document: &Document, text: &str) -> Option<String> {
         .next()?;
     Some(a_elem.attr("href")?.to_string())
 }
+
+/// Return the href attributes of all links that start with the specified prefix `text`.
+pub fn get_link_hrefs_from_text_with_prefix(document: &Document, text: &str) -> Vec<String> {
```
svenstaro, line 17:
```suggestion
pub fn get_link_hrefs_from_text_with_prefix(document: &Document, prefix: &str) -> Vec<String> {
```

tests/utils/mod.rs
```diff
@@ -12,3 +12,19 @@ pub fn get_link_from_text(document: &Document, text: &str) -> Option<String> {
         .next()?;
     Some(a_elem.attr("href")?.to_string())
 }
+
+/// Return the href attributes of all links that start with the specified prefix `text`.
```
svenstaro, line 16:
```suggestion
/// Return the href attributes of all links that start with the specified `prefix`.
```

svenstaro (changes requested):

tests/navigation.rs
```diff
@@ -147,3 +148,30 @@ fn can_navigate_using_breadcrumbs(
 
     Ok(())
 }
+
+#[rstest]
+#[case(server(&["--default-sorting-method", "date", "--default-sorting-order", "desc"]))]
+#[case(server(&["--default-sorting-method", "date", "--default-sorting-order", "asc"]))]
+/// We can specify the default sorting order
+fn can_specify_default_sorting_order(#[case] server: TestServer) -> Result<(), Error> {
+    let slash = String::from("/");
```
svenstaro, line 157: I'm not sure how I feel about this since we only really need it at a single place. Maybe just create this where it needs to be and remove this?

svenstaro:
I'm making the last few bits of changes myself. Thanks for the contribution!

svenstaro:
Happy to hear! Check my changes to your code as well. I think I improved it somewhere and it's now pretty neat.

PR #1310, opened 2024-01-11
Ignore readonly metadata and handle errors instead

### LINKED ISSUE #994: Unable to upload a/o create folder on any other disk than C:\ on Win10
Hi,
I'm using miniserve on a Windows 10 laptop

When miniserve's path to serve is on C:\ disk, upload of file a/o directory creation works fine, and is possible, at least as soon as appropriate access rights are granted on local NTFS disk.

But if the path to serve is on another local disk (eg "miniserve D:\Temp\ -u -U" ), upload of files a/o directory creation systematically fail with the error message "Insufficient permissions to create file in \\?\D:\Temp", and this even if full control has been granted to everyone at OS level in this target folder.
On the other hand, it works "sometimes" even on disk D:
For example, if I create a new folder D:\Test (at Win10 level),  then launch "miniserve D:\Test -u -U", I'm able to upload files a/o create directories... while access rights are exactly the same at OS level for D:\Test and D:\Temp....

Moreover, in case access rights are NOT granted to users in target dir, upload and directory creation are (hopefully...) not possible, but the error message is not at all the same, but rather "Failed to create mysubdir / caused by: Accès refusé. (os error 5)", or "Failed to create \\?\D:\Test\P1080514.JPG / caused by: Accès refusé. (os error 5)"

ViRb3:
Fixes #994, check the issue for more information. Tested on Windows, and also confirmed that real permission error throws 401 as requested. Logs show like this:

> Thu, 11 Jan 2024 22:19:25 +0000 [ERROR] Insufficient permissions to create file in \\?\C:\Users\User\Downloads\test\test.pdf

### DIFF
--- a/src/file_op.rs
+++ b/src/file_op.rs
@@ -4,6 +4,7 @@ use std::{
     io::Write,
     path::{Component, Path, PathBuf},
 };
+use std::io::ErrorKind;
 
 use actix_web::{http::header, web, HttpRequest, HttpResponse};
 use futures::TryStreamExt;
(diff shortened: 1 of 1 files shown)


ViRb3:
@svenstaro Sorry I'll be away from my PC for a while, can you please fix the formatting so this PR doesn't stay hanging? Thanks

svenstaro:
Thanks! We don't have a test for it but I trust your fix. :)

LaurentGrenet:
I just tested... and I confirm it works fine.
Many thanks to both of you !

PR #1315, opened 2024-01-14
Optional enable compression middleware to reduce response size  (disabled by default)

zuisong:
#523 
#497 

This PR adds an optional section to turn on compress response, turned off by default, and a note in the help docs about the impact of turning this on.

```
/// Enable compress response
///
/// WARNING: Enabling this option may slow down transfers due to CPU overhead, so it is disabled by default.
/// 
/// Only enable this option if you know that your users have slow connections or if you want to minimize your server's bandwidth usage.
```

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -214,6 +214,18 @@ pub struct CliArgs {
     #[arg(short = 'z', long = "enable-zip", env = "MINISERVE_ENABLE_ZIP")]
     pub enable_zip: bool,
 
+    /// Enable compress response
+    ///
+    /// WARNING: Enabling this option may slow down transfers due to CPU overhead, so it is disabled by default.
+    ///
+    /// Only enable this option if you know that your users have slow connections or if you want to minimize your server's bandwidth usage.
+    #[arg(
+        short = 'C',
+        long = "compress-response",
+        env = "MINISERVE_COMPRESS_RESPONSE"
+    )]
+    pub compress_response: bool,
+
     /// List directories first
     #[arg(short = 'D', long = "dirs-first", env = "MINISERVE_DIRS_FIRST")]
     pub dirs_first: bool,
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -202,6 +203,10 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), ContextualError> {
             .app_data(stylesheet.clone())
             .wrap_fn(errors::error_page_middleware)
             .wrap(middleware::Logger::default())
+            .wrap(Condition::new(
+                miniserve_config.compress_response_enabled,
```
svenstaro, line 207: Let's be consistent with ourselves here and remove the extra import at the top.
```suggestion
            .wrap(middleware::Condition::new(
                miniserve_config.compress_response_enabled,
```

svenstaro (changes requested):
Thanks! Just a tiny code change.

Did you perform some benchmarks by any chance?

zuisong:
I didn't do benchmark, but my server egress bandwidth is very low, and I'm transmitting plain text files, which really need compression.

src/args.rs
```diff
@@ -214,6 +214,17 @@ pub struct CliArgs {
...
+    ///
+    /// WARNING: Enabling this option may slow down transfers due to CPU overhead, so it is disabled by default.
+    ///
+    /// Only enable this option if you know that your users have slow connections or if you want to minimize your server's bandwidth usage.
+    #[arg(
+        long = "enable-compress-response",
+        env = "MINISERVE_ENABLE_COMPRESS_RESPONSE"
+    )]
+    pub enable_compress_response: bool,
+
```
svenstaro, line 227: What do you think about just calling this `--compress-response`? The `--enable` prefixes are only there for archiving. Also, let's also give it `short = 'C'`.

svenstaro (changes requested):

src/args.rs
zuisong: I think that's a great idea. I've actually considered doing something like that before.

svenstaro:
Super nice, thanks!

PR #1327, opened 2024-01-26
Fix inaccurate uses of `sanitize_path`

cyqsimon:
Found this while working on #1093 (finally). So I thought I'd clean it up first.


svenstaro:
Seems fine. I'm a bit unhappy that our tests didn't catch this but oh well. Thanks!

PR #1329, opened 2024-01-29
add `--disable-indexing` cli flag to completely disable directory indexing

### LINKED ISSUE #1328: Add cli flag to disable indexing
Sometimes I use miniserve to serve some files over the public internet temporarily, usually for file sharing. And sometimes I forget to turn off the server when I'm done. It would be nice if I could disable indexing so that its a little bit harder to enumerate all the files I have in the folder that is being served.

I'd be willing to implement this myself. Just let me know of any gotchas I should look out for.

dyc3:
closes #1328

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -294,6 +294,12 @@ pub struct CliArgs {
     /// Enable README.md rendering in directories
     #[arg(long, env = "MINISERVE_README")]
     pub readme: bool,
+
+    /// Disable indexing
+    /// This will prevent miniserve from generating directory listings
+    /// and will return a 404 instead.
+    #[arg(long, env = "MINISERVE_DISABLE_INDEXING")]
+    pub disable_indexing: bool,
 }
 
 /// Checks whether an interface is valid, i.e. it can be parsed into an IP address
--- a/src/args.rs
+++ b/src/args.rs
@@ -296,9 +296,10 @@ pub struct CliArgs {
     pub readme: bool,
 
     /// Disable indexing
+    ///
     /// This will prevent miniserve from generating directory listings
     /// and will return a 404 instead.
-    #[arg(long, env = "MINISERVE_DISABLE_INDEXING")]
+    #[arg(short = 'I', long, env = "MINISERVE_DISABLE_INDEXING")]
     pub disable_indexing: bool,
 }
 
--- a/src/args.rs
+++ b/src/args.rs
@@ -297,8 +297,8 @@ pub struct CliArgs {
 
     /// Disable indexing
     ///
-    /// This will prevent miniserve from generating directory listings
-    /// and will return a 404 instead.
+    /// This will prevent directory listings from being generated
+    /// and return an error instead.
     #[arg(short = 'I', long, env = "MINISERVE_DISABLE_INDEXING")]
     pub disable_indexing: bool,
 }
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -376,6 +376,15 @@ pub fn directory_listing(
                 .body(actix_web::body::BodyStream::new(rx)),
         ))
     } else {
+        if conf.disable_indexing {
+            return Ok(ServiceResponse::new(
+                req.clone(),
+                HttpResponse::NotFound()
+                    .content_type(mime::TEXT_PLAIN_UTF_8)
+                    .body("File not found."),
+            ));
+        }
+
         Ok(ServiceResponse::new(
             req.clone(),
             HttpResponse::Ok().content_type(mime::TEXT_HTML_UTF_8).body(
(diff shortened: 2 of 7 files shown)


svenstaro:
Please add a test for this. It's a fairly impactful change.

src/args.rs
```diff
@@ -294,6 +294,12 @@ pub struct CliArgs {
     /// Enable README.md rendering in directories
     #[arg(long, env = "MINISERVE_README")]
     pub readme: bool,
+
+    /// Disable indexing
+    /// This will prevent miniserve from generating directory listings
+    /// and will return a 404 instead.
+    #[arg(long, env = "MINISERVE_DISABLE_INDEXING")]
```
svenstaro, line 301:
```suggestion
    #[arg(short = 'I', long, env = "MINISERVE_DISABLE_INDEXING")]
```

src/args.rs
```diff
@@ -294,6 +294,12 @@ pub struct CliArgs {
     /// Enable README.md rendering in directories
     #[arg(long, env = "MINISERVE_README")]
     pub readme: bool,
+
+    /// Disable indexing
+    /// This will prevent miniserve from generating directory listings
+    /// and will return a 404 instead.
```
svenstaro, line 300:
```suggestion
    /// Disable indexing
    ///
    /// This will prevent directory listings from being generated
    /// and return an error instead.
```

src/listing.rs
```diff
@@ -376,6 +376,15 @@ pub fn directory_listing(
...
     } else {
+        if conf.disable_indexing {
+            return Ok(ServiceResponse::new(
+                req.clone(),
+                HttpResponse::NotFound()
+                    .content_type(mime::TEXT_PLAIN_UTF_8)
+                    .body("File not found."),
+            ));
+        }
+
```
svenstaro, line 387: Doesn't putting it here result in the index still being downloadable via requesting an archive? This seems to me like a potential security problem or at least something people might not expect. I think it might be safer to prohibit index generation entirely in `main.rs`.

svenstaro (changes requested):

dyc3:
I added a couple tests. I'm not super familiar with your codebase, so let me know if there's some better assertions I could use.

src/listing.rs
dyc3: I wasn't aware that was a thing. I'll take a look.

tests/serve_request.rs
```diff
@@ -321,3 +321,34 @@ fn serves_requests_static_file_check(
...
+    Ok(())
+}
+
+#[rstest]
+#[case(server(&["--disable-indexing"]))]
+fn serves_file_requests_when_indexing_disabled(#[case] server: TestServer) -> Result<(), Error> {
+    reqwest::blocking::get(format!("{}/{}", server.url(), FILES[0]))?.error_for_status()?;
+
+    Ok(())
+}
```
svenstaro, line 354: Solid tests! I would just add a test that checks that archiving is also not possible when indexing is disabled as otherwise archiving would get around this as explained above.

src/listing.rs
dyc3: After looking at it closer, I think it'll be more clean to just move this check all the way to the top of this function. All file listings seem to go through this handler.

src/args.rs
svenstaro: I don't see this being used. I'd prefer this wording.

dyc3: Ah whoops. I'll fix that.

src/listing.rs
```diff
@@ -161,6 +161,14 @@ pub fn directory_listing(
...
 
     let conf = req.app_data::<crate::MiniserveConfig>().unwrap();
+    if conf.disable_indexing {
+        return Ok(ServiceResponse::new(
+            req.clone(),
+            HttpResponse::NotFound()
+                .content_type(mime::TEXT_PLAIN_UTF_8)
+                .body("File not found."),
+        ));
+    }
```
svenstaro, line 171: I think this is probably fine but I'm wondering: Can we just disable indexing entirely in `main.rs`?

dyc3: Technically yes, and if you want to see what that looks like, you can check out [7b5b1a0](https://github.com/svenstaro/miniserve/pull/1329/commits/7b5b1a0e7fec2c0577eb0dc2a93e9fde48228a67). Reiterating what I said [in this comment](https://github.com/svenstaro/miniserve/pull/1329?notification_referrer_id=NT_kwDOABuZp7I5MjQ5MTgzMzUyOjE4MDg4MDc#discussion_r1471362494) I think it would be less complicated to do it like this.

svenstaro:
Excellent stuff, tests are spot on, too!

PR #1331, opened 2024-01-29
Refactor errors

cyqsimon:
## Motivation

I noticed that the `ResponseError` impl for `ContexualError` was not very complete. Many variants simply default to 500-ISE by wildcard match.

So I tried to fix that, but in the process I realised that some error variants are only encountered at program startup and not at runtime. Therefore I think it's much more appropriate to put them in separate enums.

## Change summary

- Split `ContexualError` into `StartError` & `RuntimeError` & `AuthParseError`
- Made sure every `RuntimeError` variant has an accurate status code

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -11,7 +11,7 @@ use thiserror::Error;
 use crate::{renderer::render_error, MiniserveConfig};
 
 #[derive(Debug, Error)]
-pub enum StartError {
+pub enum StartupError {
     /// Any kind of IO errors
     #[error("{0}\ncaused by: {1}")]
     IoError(String, std::io::Error),
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -11,7 +11,23 @@ use thiserror::Error;
 use crate::{renderer::render_error, MiniserveConfig};
 
 #[derive(Debug, Error)]
-pub enum ContextualError {
+pub enum StartError {
+    /// Any kind of IO errors
+    #[error("{0}\ncaused by: {1}")]
+    IoError(String, std::io::Error),
+
+    /// In case miniserve was invoked without an interactive terminal and without an explicit path
+    #[error("Refusing to start as no explicit serve path was set and no interactive terminal was attached
+Please set an explicit serve path like: `miniserve /my/path`")]
+    NoExplicitPathAndNoTerminal,
+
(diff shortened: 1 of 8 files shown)


cyqsimon:
Btw I intend to queue #1093 after this one.

src/errors.rs
```diff
@@ -11,7 +11,23 @@ use thiserror::Error;
 use crate::{renderer::render_error, MiniserveConfig};
 
 #[derive(Debug, Error)]
-pub enum ContextualError {
+pub enum StartError {
```
svenstaro, line 14: Let's go with this instead.
```suggestion
pub enum StartupError {
```

src/errors.rs
```diff
@@ -75,28 +75,25 @@ pub enum ContextualError {
...
+            E::DuplicateFileError => S::CONFLICT,
+            E::UploadForbiddenError => S::FORBIDDEN,
+            E::InvalidPathError(_) => S::BAD_REQUEST,
+            E::InsufficientPermissionsError(_) => S::FORBIDDEN,
+            E::ParseError(_, _) => S::BAD_REQUEST,
+            E::ArchiveCreationError(_, err) => err.status_code(),
+            E::ArchiveCreationDetailError(_) => S::INTERNAL_SERVER_ERROR,
+            E::InvalidHttpCredentials => S::UNAUTHORIZED,
+            E::InvalidHttpRequestError(_) => S::BAD_REQUEST,
+            E::RouteNotFoundError(_) => S::NOT_FOUND,
```
svenstaro, line 96: Oh, I really like this! I didn't know we were hiding so many errors!

svenstaro (changes requested):
I really like this! Maybe just rename the `Start` stuff to `Startup` and we should be good.

src/errors.rs
cyqsimon: Yeah this is the one thing I ***REALLY*** dislike about non-specific wildcard match clauses. When you add a new enum variant, there's no indication from the LSP that you probably also want to update the match block. I'd say it's almost as bad as [`enum_glob_use`](https://rust-lang.github.io/rust-clippy/master/#/enum_glob_use).

svenstaro:
Very nice stuff!

PR #1364, opened 2024-03-16
add folder and file symbol

chaibiq:
Hi, i just add  folder dan file symbol, it might look for better UI. 

Clip

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -503,7 +503,7 @@ fn entry_row(
                             }
                         }@else {
                             a.directory href=(parametrized_link(&entry.link, sort_method, sort_order, raw)) {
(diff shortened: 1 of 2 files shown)


src/renderer.rs
```diff
@@ -503,7 +503,7 @@ fn entry_row(
                             }
                         }@else {
                             a.directory href=(parametrized_link(&entry.link, sort_method, sort_order, raw)) {
-                                (entry.name) "/"
+                                "📁 " (entry.name) "/"
```
svenstaro, line 506: I like this! However, could you please put this into the SCSS part rather than directly here? Take a look at how `symlink-symbol` is done.

chaibiq: sure,  I will take a look there.

PR #1379, opened 2024-04-02
Update style.scss

chaibiq:
fix wrapping text in mobile view when the file name too long, the box of file size become inconsistent.
PR


svenstaro:
Thanks!

PR #1389, opened 2024-04-12
Improve documentation for the --header parameter

orwithout:
A more detailed explanation of the use of the parameter --header

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -235,7 +235,9 @@ pub struct CliArgs {
     #[arg(short = 't', long = "title", env = "MINISERVE_TITLE")]
     pub title: Option<String>,
 
-    /// Set custom header for responses
+    /// Inserts custom headers in the responses.
+    /// Specify headers as 'Header1:Value1;Header2:Value2' for multiple headers in a single use,
+    /// or use --header multiple times, each with one 'Header:Value' pair.
     #[arg(
         long = "header",
         value_parser(parse_header),
--- a/src/args.rs
+++ b/src/args.rs
@@ -240,7 +240,7 @@ pub struct CliArgs {
     ///
     /// Example:
     /// --header "Header1:Value1" --header "Header2:Value2"
-    /// (If a header is already set or previously inserted, it will not be overwritten)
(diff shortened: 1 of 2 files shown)


src/args.rs
```diff
@@ -235,7 +235,8 @@ pub struct CliArgs {
     #[arg(short = 't', long = "title", env = "MINISERVE_TITLE")]
     pub title: Option<String>,
 
-    /// Set custom header for responses
+    /// Inserts custom headers into the responses. Specify each header as a 'Header:Value' pair.
+    /// This parameter can be used multiple times to add multiple headers.
```
svenstaro, line 239: Since you're already here:
1. Perhaps provide an example?
1. Maybe also document the behavior what happens when a header conflicts with a previous one?

svenstaro (changes requested):
Great stuff! Now that I think about it, we could make even improve _that_. I left a comment.

src/args.rs
orwithout: Thanks for the feedback, @svenstaro! I've updated the PR:

- Added an example in the README demonstrating how to use the `--header` option for enabling HSTS, which helps in achieving an A+ SSL Labs rating.
- Clarified the behavior when header conflicts occur: new headers will not overwrite existing values if already set.

Please let me know if further adjustments are needed.

orwithout (commented):
Thank you for the review. Here are the updates made:

- **Improved Documentation**: Enhanced clarity on `--header` usage, particularly regarding how conflicts with existing headers are handled.

- **HSTS Example**: Added an example in the README for enabling HSTS using `--header`, essential for achieving an A+ rating on SSL Labs.

- **Fullchain Certificates Usage**: Documented the use of fullchain certificates with the `--tls-cert` option to improve HTTPS security.

These changes aim to clarify usage and enhance security features. Please let me know if additional modifications are necessary.

svenstaro:
Thanks!

orwithout:
> Thanks!

Glad to contribute! And it means a lot to me!

PR #1415, opened 2024-05-13
add read-only webdav support

ahti:
Adds read-only webdav support by (when indexing isn't disabled) routing OPTIONS/PROPFIND requests to a `DavHandler` from the `webdav-server` crate. Hidden files are excluded via a custom `DavFileSystem`.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,13 +6,18 @@ use std::time::Duration;
 use actix_files::NamedFile;
 use actix_web::{
     dev::{fn_service, ServiceRequest, ServiceResponse},
-    http::header::ContentType,
+    guard,
+    http::{header::ContentType, Method},
     middleware, web, App, HttpRequest, HttpResponse, Responder,
 };
 use actix_web_httpauth::middleware::HttpAuthentication;
 use anyhow::Result;
 use clap::{crate_version, CommandFactory, Parser};
 use colored::*;
+use dav_server::{
+    actix::{DavRequest, DavResponse},
+    DavConfig, DavHandler, DavMethodSet,
+};
 use fast_qr::QRBuilder;
 use log::{error, warn};
 
@@ -27,9 +32,11 @@ mod file_utils;
 mod listing;
 mod pipe;
 mod renderer;
+mod webdav_fs;
 
 use crate::config::MiniserveConfig;
 use crate::errors::{RuntimeError, StartupError};
+use crate::webdav_fs::RestrictedFs;
 
 static STYLESHEET: &str = grass::include!("data/style.scss");
 
@@ -84,6 +91,12 @@ async fn run(miniserve_config: MiniserveConfig) -> Result<(), StartupError> {
 
     if miniserve_config.no_symlinks && miniserve_config.path.is_symlink() {
         return Err(StartupError::NoSymlinksOptionWithSymlinkServePath(
+            miniserve_config.path.to_string_lossy().to_string(),
+        ));
+    }
+
+    if miniserve_config.webdav_enabled && miniserve_config.path.is_file() {
+        return Err(StartupError::WebdavWithFileServePath(
             miniserve_config.path.to_string_lossy().to_string(),
         ));
     }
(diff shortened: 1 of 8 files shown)


src/webdav_fs.rs
```diff
@@ -0,0 +1,72 @@
+use dav_server::{davpath::DavPath, fs::*, localfs::LocalFs};
+use futures::{future::ready, StreamExt, TryFutureExt};
+use std::path::{Component, Path};
```
svenstaro, line 3: Give this module a description.

src/webdav_fs.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[derive(Clone)]
+pub struct RestrictedFs {
+    local: Box<LocalFs>,
+    allow_hidden: bool,
+}
```
svenstaro, line 9: Give this a proper description of what this is supposed to do and why we need it.

src/webdav_fs.rs
```diff
@@ -0,0 +1,72 @@
...
+        Box::new({
+            RestrictedFs {
+                local,
+                allow_hidden,
+            }
+        })
+    }
+}
+
+fn check_path(path: &DavPath) -> bool {
```
svenstaro, line 23: Describe what the idea here is.

svenstaro (changes requested):
The webdav functionality should likely be put behind a flag. I suppose most people wouldn't want to run it with webdav activated.

Also, can you produce a test that shows that webdav at least works in _some_ fashion?

svenstaro:
Please rebase this!

src/main.rs
```diff
@@ -308,7 +315,9 @@ fn configure_header(conf: &MiniserveConfig) -> middleware::DefaultHeaders {
 /// This is where we configure the app to serve an index file, the file listing, or a single file.
 fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
     let dir_service = || {
-        let mut files = actix_files::Files::new("", &conf.path);
+        // use routing guard so propfind and options requests fall through to the webdav handler
```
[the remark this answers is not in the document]
ahti: The WebDAV spec [requires](http://www.webdav.org/specs/rfc4918.html#HEADER_DAV):
> All DAV-compliant resources must return the DAV header with compliance-class "1" on all OPTIONS responses.

Actix-files [only handles HEAD and GET](https://github.com/actix/actix-web/blob/0fafb486d4877df3d505ea6533765af499d99a5e/actix-files/src/service.rs#L105) either way, so I don't think this needs to change.

src/main.rs
```diff
@@ -376,6 +385,36 @@ fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
...
+                .hide_symlinks(conf.no_symlinks)
+                .strip_prefix(conf.route_prefix.to_owned())
+                .build_handler();
+
+            app.app_data(web::Data::new(dav_server.clone()));
+
+            // order is important: the dir service above is checked first, but has a method guard,
+            // so options and propfind go here. if this service was registered first, it would swallow the gets/heads
+            app.service(
+                web::resource("/{tail:.*}")
```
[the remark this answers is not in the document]
ahti: Using just "/.*" for the path, actix-web prints a bunch of "[WARN] Tail segments must have names. Consider `.../{tail}*`. This may become a panic in the future.".

svenstaro (changes requested):
I'd like to see you use a real webdav client to check whether this does something useful. I think [reqwest_dav](https://github.com/niuhuan/reqwest_dav) would be suitable. Also I'd like to see you handle:
- Single file serving mode ([new_file](https://docs.rs/dav-server/latest/dav_server/localfs/struct.LocalFs.html#method.new_file) seems to already to this)
- `--route-prefix`/`--random-route`
- `--no-symlinks`
- Optional: `--mkdir`
- Optional: `--upload-files`

svenstaro (approved):
Excellent work, merging as-is!

Issue #1369, opened 2024-03-21
Save uploaded file atomically


ubruhin:
It seems when uploading a file, the file is immediately created and any incoming data is written into it. However, this means the file is incomplete/invalid during the upload and there's no way to know as a user (or, in my case a tool processing the uploaded files) whether a file is complete or not. I didn't check what happens when an upload is aborted/fails, is the incomplete file then even left over?

Anyway, IMHO the upload would be much safer if the file is saved atomically. Just as a simple idea, the upload could be performed into a temporary file (e.g. ending with `~`) and renamed when the upload succeeded. This way it is visible for humans and detectable by tools which files are incomplete.

svenstaro:
I like this suggestion and agree with you reasoning. I won't have time to tend to this, though. I'd be very happy to review a pull request, though.

chaibiq:
Hi, I have been trying add and to test upload files with progress bar. I use external vanilla JS and CSS file to use XHR to upload files. If it's complete, I will submit PR after merge to main code without external file.
Here is it look like now: 
![318087142-53a9410f-a05f-4b09-975b-7299db14575f](https://github.com/svenstaro/miniserve/assets/18382453/ef7e2bac-b6e7-4cf6-9c78-5f10b26f2c49)

PR #1431, opened 2024-06-06
feat: Added HTML and Javascript progress bar when uploading files

AlecDivito:
# Overview

After running this application, i noticed that the upload functionality is pretty bad when it comes to user experience. When searching the Github issues I found #1369 which seemed to link to a branch that adds basic support for a progress bar upload feature. I thought it would be fun to try and implement it because I enjoy how simple this tool is and would like the functionality for myself.

This PR adds the ability to view uploads as they progress.

## Issues

While implementing the solution, I found sometimes Actix could return very very early after a file upload is sent (for example, when a request required authentication). When this happened, the browser fails to register the response the server sends back **unless** the entire web request is consumed. I do believe that this may block this PR from being merged :(

## Implementation

Here is a video showing the upload implementation. My computer is a bit fast so it's a bit hard to see it actually working.

https://github.com/svenstaro/miniserve/assets/12245650/89c7b480-dd55-4920-a411-71c413de4899

I've been able to create a simple example, showing it for slower uploads. Note that I had to edit my code to use `setInterval` so some of the text doesn't work as expected.

https://github.com/svenstaro/miniserve/assets/12245650/f7ed021b-72a5-4052-9dbd-00f14f078db4

You review would be greatly appreciated. Thanks.


Atreyagaurav:
This actually looks excellent. UI wise, it's good. I tested it out and saw that it looks good in PC, but on phone it get's full width and slightly obscured on the bottom side (have to scroll even for single entry).

One significant problem I saw was that, you have added the capacity to cancel the upload here, which is great specially since you can cancel all or file wise basis. But it seems to just cancel the file in middle, and not remove that file from the server. So I think if we're to add the option to cancel, then we need to remove those partial files.

So, I think a way to save the files with `.part` extension and then move upon completion, or another request to the server that upload was cancelled so it removed the file should come along with the feature.

svenstaro:
@AlecDivito Hey, sorry for letting this sit for so long! I actually thing this is great. I'd like to give it a proper review but first what do you think about the feedback by @Atreyagaurav?

AlecDivito:
I think the comments are great, I've just been busy, but, I'd be happy to address the comments! By all means wait until the changes are in.

> This actually looks excellent. UI wise, it's good. I tested it out and saw that it looks good in PC, but on phone it gets full width and slightly obscured on the bottom side (have to scroll even for single entry).

I believe I did this intentionally on mobile. I'll review my design to address this.

> So I think if we're to add the option to cancel, then we need to remove those partial files.

This is me not knowing the code base well enough haha.

> So, I think a way to save the files with .part extension and then move upon completion, or another request to the server that upload was cancelled so it removed the file should come along with the feature.

Not a bad idea, I'll give it a go.

@Atreyagaurav Thank you for the review :)

AlecDivito:
Addressed some of the comments:

1. Mobile view now only takes up about 80px at the bottom of the screen. However the user can toggle the uploader so it takes up the full hight.
2. Uploading uses a temporary file. Users can optionally send a SHA256 hash which if included in the `/upload` request header, will be used to check to see if the correct file was uploaded. If incorrect the file is deleted.

I had trouble finding where exactly I could get request cancellations for actix, it actually seems this is one of the web servers in Rust that doesn't easily provide that data. Therefore, I'm relying on SHA256 hashs to determine if a request was cancelled and thus, if a file should be deleted.

I would rather stay away from adding a new endpoint to clean up aborted downloads as that would increase the scope of this feature to a level I'm not comfortable with implementing.

I also didn't want to break existing functionality of the upload endpoint, so if the SHA256 isn't included in the request, I don't check for it. I believe this was the pre-existing functionality so it should be ok...

https://github.com/user-attachments/assets/3d3c771a-6ea7-486f-b152-14bf5b2fc13f

Anyways, here's a demo, @svenstaro when you have the time please let me know what you think.

src/file_op.rs
```diff
@@ -4,50 +4,115 @@ use std::io::ErrorKind;
...
+        if let Some(hasher) = hasher.as_mut() {
+            hasher.update(&bytes)
+        }
+        if let Err(e) = temp_file.write_all(&bytes).await {
+            error = Some(RuntimeError::IoError(
+                "Failed to write to file".to_string(),
+                e,
+            ));
+            break;
+        }
```
svenstaro, line 78: Why do we even store the error here? Couldn't we just return `drop(temp_file);` and then `return` the error?

src/file_op.rs
```diff
@@ -4,50 +4,115 @@ use std::io::ErrorKind;
...
+        if let Some(FileHash::SHA256(expected_hash)) = file_hash {
+            let actual_hash = hex::encode(hasher.finalize());
+            if &actual_hash != expected_hash {
+                let _ = tokio::fs::remove_file(&temp_path).await;
+                return Err(RuntimeError::UploadHashMismatchError);
+            }
+        }
+    }
+
+    if let Err(e) = tokio::fs::rename(&temp_path, &file_path).await {
```
svenstaro, line 105: Maybe an `info!()` log would be good here to inform users running with `--verbose` that the upload is being moved?

src/file_op.rs
```diff
@@ -4,50 +4,115 @@ use std::io::ErrorKind;
...
+        let _ = tokio::fs::remove_file(&temp_path).await;
+        return Err(RuntimeError::IoError(
+            format!(
+                "Failed to move temporary file {} to {}",
+                temp_path.display(),
+                file_path.display()
+            ),
```
svenstaro, line 112: I think this should work:
```suggestion
            format!("Failed to move temporary file {temp_path:?} to {file_path:?}"),
```

src/file_op.rs
```diff
@@ -218,6 +290,27 @@ pub async fn upload_file(
...
+            .and_then(|h| h.to_str().ok())
+        {
+            match hash_funciton.to_ascii_uppercase().as_str() {
+                "SHA256" => {
+                    file_hash = Some(FileHash::SHA256(hash.to_string()));
+                }
+                _ => {}
+            }
+        }
+    }
```
svenstaro, line 311: Just a stylistic thing but how about this?

```rust
if let (Some(hash), Some(hash_function)) =
    (
        req.headers().get("X-File-Hash").and_then(|h| h.to_str().ok()),
        req.headers().get("X-File-Hash-Function").and_then(|h| h.to_str().ok())
    )
{
    match hash_function.to_ascii_uppercase().as_str() {
        "SHA256" => file_hash = Some(FileHash::SHA256(hash.to_string())),
        _ => unreachable!() // Which shouldn't ever get here, right?
    }
}
```

tests/upload_files.rs
```diff
@@ -12,7 +13,16 @@ mod fixtures;
 use crate::fixtures::{server, tmpdir, Error, TestServer};
 
 #[rstest]
-fn uploading_files_works(#[with(&["-u"])] server: TestServer) -> Result<(), Error> {
+#[case("", "")]
+#[case(
```
svenstaro, line 17: We haven't done any case naming in the project yet but perhaps here it would be appropriate?
```suggestion
#[case::no_hash("", "")]
#[case::sha256_hash(...)
```

tests/upload_files.rs
```diff
@@ -33,7 +43,12 @@ fn uploading_files_works(#[with(&["-u"])] server: TestServer) -> Result<(), Erro
         .mime_str("text/plain")?;
     let form = form.part("file_to_upload", part);
 
-    let client = Client::new();
+    let mut headers = HeaderMap::new();
+    headers.insert("X-File-Hash", sha.parse()?);
+    headers.insert("X-File-Hash-Function", sha_func.parse()?);
```
svenstaro, line 48: I think it would be more proper to make this test behave in the same way a client leaving out the headers would do it. I suppose it would therefore be more realistic to make the parameters into `Option<String>`s?

Frankly I'm not sure how this currently passes since we're setting the headers but we just leave the values as an empty string.

svenstaro (changes requested):
> I also didn't want to break existing functionality of the upload endpoint, so if the SHA256 isn't included in the request, I don't check for it. I believe this was the pre-existing functionality so it should be ok...

I think that's a good idea. People using curl shouldn't have to go through hoops to accomplish this.

All in all, I reviewed this and I think it's stellar work. There's some questions and a few minor things but I think the overall structure is solid.

src/file_op.rs
```diff
@@ -4,50 +4,115 @@ use std::io::ErrorKind;
...
     overwrite_files: bool,
+    file_hash: Option<&FileHash>,
 ) -> Result<u64, RuntimeError> {
     if !overwrite_files && file_path.exists() {
         return Err(RuntimeError::DuplicateFileError);
     }
 
-    let file = match File::create(&file_path).await {
-        Err(err) if err.kind() == ErrorKind::PermissionDenied => Err(
+    let named_temp_file = match tokio::task::spawn_blocking(|| NamedTempFile::new()).await {
```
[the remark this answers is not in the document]
AlecDivito: I think this is a good idea. It should be clear up front from reading the documentation and command line parameters that uploads are placed in a temporary directory before being moved. Allowing the user to configure this would be a smart idea.

I do believe though that you might want to make this change a major release then because, like you said, it semantically changes the upload functionality.

I'll let you decide that, and I'll add the option into the project.

src/file_op.rs
AlecDivito: We could do that, but we would be duplicating code because we'd need to call `drop(temp_file)` and `tokio::fs::remove_file(...)` anyways. I also noticed that I should have been calling `flush` on the file as well which will also update the existing error if it doesn't exist.

You'll be able to see my changes soon, but I don't think I support this suggestion.

tests/upload_files.rs
AlecDivito: It passes because the endpoint isn't programmed to fail if incorrect values are passed in, instead, it will continue on silently. It might be worthwhile to update it so that if the headers are in the HTTP request, then it's enforced they have valid values.

AlecDivito:
Sorry for the wait, I'm busy during the week.

I've addressed the majority of the comments and added quite a bit of documentation so hopefully it will be easier to understand the entire upload flow now. Here's a description of some of the changes from the last push:

1. Added comments to javascript and upload file functions
2. Added ability to set `SHA512` as well
3. Added in extra tests
4. Updated logic for getting SHA value from headers and be more strict on their validation
5. Remove list sorting from javascript (it never worked 🥲)
6. Updated concurrency in javascript so that promises take from a shared queue of upload tasks

Overall, thanks for reviewing, after taking another look at my code I did notice some bugs and I've tried to remove them.

src/config.rs
```diff
@@ -275,9 +278,16 @@ impl MiniserveConfig {
             .transpose()?
             .unwrap_or_default();
 
+        let temp_upload_directory = args.temp_upload_directory.as_ref().take().map(|v| if v.exists() && v.is_dir() {
+            Ok(v.clone())
+        } else {
+            Err(anyhow!("Upload temporary directory must exist and be a directory. Validate that path {v:?} meets those requirements"))
+        }).transpose()?;
```
svenstaro, line 285: I think this would be better done at the clap level in `args.rs` with a `value_parser()` function that checks whether the path exists.

src/args.rs
```diff
@@ -26,6 +26,20 @@ pub struct CliArgs {
...
+    ///
+    /// This value will only be used **IF** file uploading is enabled. If this option is
+    /// not set, the operating system default temporary directory will be used.
+    #[arg(
+        long = "temp-directory",
+        value_hint = ValueHint::FilePath,
+        requires = "allowed_upload_dir",
+        env = "MINISERVER_TEMP_UPLOAD_DIRECTORY")
+    ]
+    pub temp_upload_directory: Option<PathBuf>,
```
svenstaro, line 41: A thought I had about this: If the dir is chosen so that it is in a served path, should we serve it? Or hide (and 403 any direct file hits) it or show it in some special way?

svenstaro (changes requested):
Super cool, we're very close now. Some comments to address and some clippy to fix.

src/args.rs
AlecDivito: Personally, I would just show the directory and the file uploads that are happening, because if someone decides to set the upload directory to something visible, they are probably doing so on purpose.

I think that this should be addressed _if_ someone posts an issue about it.

AlecDivito:
I addressed your comments and fixed the clippy issues. I hope this is good enough to merge! 🎉🎉🎉

Let me know if you have any other comments!

src/args.rs
svenstaro: That's a fair point!

svenstaro:
Super nice. It was almost a year in the making but we got it done! Merging.

PR #1442, opened 2024-07-14
Don't show mkdir option when the directory is not upload allowed

### LINKED ISSUE #1249: Hide "Create Directory" panel for folders without write-permission
Imo it would make sense to hide it in the same way as the upload-panel, since it is non-functional anyway.

Atreyagaurav:
Fixes #1249


svenstaro:
Ah yes, that makes sense. Thanks!

PR #1453, opened 2024-09-14
new flag: Enable renaming during file upload if duplicate exists

### LINKED ISSUE #1452: No rename option, only overwrite

Atreyagaurav:
Fixes #1452

I think this will work as a flag to rename files. It won't work when the file contains multiple extension like `.tar.gz` which is hard to do, as we could make it work with that, but any file with multiple `.` in the text for no meaning will mess it up. But I can make that change if you think that's acceptable.

Also, it will break compatibility, but we could combine the overwrite and rename flags into a single flag like `--duplicate-files` that takes enum (rename/overwrite).

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,10 +312,15 @@ Options:
 
           [env: MINISERVE_RAW_MEDIA_TYPE=]
 
-  -o, --overwrite-files
-          Enable overriding existing files during file upload
+  -o, --on-duplicate-files <ON_DUPLICATE_FILES>
+          What to do if existing files with same name is present during file upload
 
-          [env: MINISERVE_OVERWRITE_FILES=]
+          If you enable renaming files, the renaming will occur by adding numerical suffix to the filename before the final extension. For example file.txt will be uploaded
+          as file-1.txt, the number will be increased until an available filename is found.
+
+          [env: MINISERVE_ON_DUPLICATE_FILES=]
+          [default: error]
+          [possible values: error, overwrite, rename]
 
   -r, --enable-tar
           Enable uncompressed tar archive generation
@@ -543,7 +548,7 @@ You can provide `-i` multiple times to bind to multiple interfaces at the same t
 
 This is mostly a note for me on how to release this thing:
 
-- Make sure `CHANGELOG.md` is up to date.
+- Make sure [CHANGELOG.md](./CHANGELOG.md) is up to date.
 - `cargo release <version>`
 - `cargo release --execute <version>`
 - Releases will automatically be deployed by GitHub Actions.
--- a/src/args.rs
+++ b/src/args.rs
@@ -262,7 +262,7 @@ pub struct CliArgs {
     /// What to do if existing files with same name is present during file upload
     ///
     /// If you enable renaming files, the renaming will occur by
-    /// adding numerical suffix to the filename before the final
+    /// adding a numerical suffix to the filename before the final
     /// extension. For example file.txt will be uploaded as
     /// file-1.txt, the number will be increased until an available
     /// filename is found.
--- a/src/args.rs
+++ b/src/args.rs
@@ -16,6 +16,14 @@ pub enum MediaType {
     Video,
 }
 
+#[derive(Debug, ValueEnum, Clone, Default, Copy)]
+pub enum DuplicateFile {
+    #[default]
+    Error,
+    Overwrite,
+    Rename,
+}
+
 #[derive(ValueEnum, Clone)]
 pub enum SizeDisplay {
     Human,
@@ -251,13 +259,20 @@ pub struct CliArgs {
     )]
     pub media_type_raw: Option<String>,
 
-    /// Enable overriding existing files during file upload
+    /// What to do if existing files with same name is present during file upload
+    ///
+    /// If you enable renaming files, the renaming will occur by
+    /// adding numerical suffix to the filename before the final
+    /// extension. For example file.txt will be uploaded as
+    /// file-1.txt, the number will be increased until an available
+    /// filename is found.
     #[arg(
         short = 'o',
-        long = "overwrite-files",
-        env = "MINISERVE_OVERWRITE_FILES"
+        long = "on-duplicate-files",
+        env = "MINISERVE_ON_DUPLICATE_FILES",
+        default_value = "error"
     )]
-    pub overwrite_files: bool,
+    pub on_duplicate_files: DuplicateFile,
 
     /// Enable uncompressed tar archive generation
     #[arg(short = 'r', long = "enable-tar", env = "MINISERVE_ENABLE_TAR")]
--- a/src/file_op.rs
+++ b/src/file_op.rs
@@ -131,6 +131,7 @@ async fn save_file(
                     } else {
                         file_path.with_file_name(format!("{}-{}", file_name, i))
                     };
+                    // If we have a file name that doesn't exist yet then we'll use that.
                     if !fp.exists() {
                         file_path = fp;
                         break;
--- a/src/file_op.rs
+++ b/src/file_op.rs
@@ -27,8 +27,8 @@ use tokio::io::AsyncWriteExt;
 use tokio::sync::RwLock;
 
 use crate::{
-    config::MiniserveConfig, errors::RuntimeError, file_utils::contains_symlink,
-    file_utils::sanitize_path,
+    args::DuplicateFile, config::MiniserveConfig, errors::RuntimeError,
+    file_utils::contains_symlink, file_utils::sanitize_path,
 };
 
 enum FileHash {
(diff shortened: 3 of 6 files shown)


Atreyagaurav:
I also thought about just renaming when `overwrite` is false instead of error. Which is probably a reasonable thing to do as well. Let me know what you think.

svenstaro:
Thanks for this! I'd really like to see some tests here, though. It just occurred to me that we have no tests for conflicting files without any flags set nor did we have one for `--overwrite-files`. Would be up for making a bunch of tests showing
1. Overwriting files is not possible without extra flags
2. `--overwrite-files` works as expected
3. `--rename-duplicate` works as expected
?

> I also thought about just renaming when overwrite is false instead of error. Which is probably a reasonable thing to do as well. Let me know what you think.

I think we want an error here by default. It's safer.

src/args.rs
```diff
@@ -198,6 +198,15 @@ pub struct CliArgs {
     #[arg(short = 'o', long = "overwrite-files", env = "OVERWRITE_FILES")]
     pub overwrite_files: bool,
 
+    /// Enable renaming files during file upload if duplicate exists
```
svenstaro, line 201: This should give the user an indication of exactly _how_ they will be renamed.

src/file_op.rs
```diff
@@ -21,11 +21,34 @@ use crate::{
...
+                } else {
+                    file_path.with_file_name(format!("{}-{}", file_name, i))
+                }
+            });
+            for fp in filepaths {
+                if !fp.exists() {
+                    file_path = fp;
+                    break;
+                }
+            }
```
svenstaro, line 48: This is a little bit dense and not super self-descriptive. Could you put in some comments describing the idea and/or rewrite the code a bit to be more self-descriptive?

Also, shouldn't always just go `.n` perhaps? I'm not really sure how other systems do this but maybe we can follow some convention that users could have come to expect from elsewhere?

svenstaro (changes requested):

src/file_op.rs
Atreyagaurav: It doesn't go .n, it adds the number after the filename and before the extension. It's how Firefox download works. 
`file.txt` becomes `file-1.txt` if it already exists (or `file-2.txt` and so on), the dense and complicated looking bits is because rust needs complicated logic to extract file-stem and extension... I'll see if I can extract this logic to a function and explain it there. 

The core logic is:
get file_stem (without extension)
get file_extension 

Loop for N from 1 to ∞:
- if file has extension: do file_stem-N.extension
- if file has no extension: do file_stem-N
- check if new name already exists: if not, exit else increment N

Atreyagaurav:
Thanks for the review, I'll push with the tests once I have a bit more free time.

src/file_op.rs
svenstaro: I know what the code does, I just found it a little hard to read and would like some code comments at the very least to make sure the desired logic is abundantly clear. :)

src/args.rs
Atreyagaurav: Done

src/file_op.rs
Atreyagaurav: I made a for loop with the numbers instead, it should be easier to read. Also added comments.

Atreyagaurav:
I have added the tests for the three conditions mentioned previously.

I copied the tests from the `upload_file` test, and modified them. The test are for failing to upload a duplicate file, checking overwrite with new contents, and checking rename flag that creates an additional file.

svenstaro:
@Atreyagaurav Looking great! What do you think about the enum thing that we discussed above?

Atreyagaurav:
I added it as enum, and made the default be error. So only breaking change is 
`-o` not needs a value as it is no longer bool. and `--overwrite-files` is not available. I named the new flag `on-duplicate-files`

I have the tests updated as well. And I also manually tested.

svenstaro:
Hey, I'd like to get this into the next release if possible. Could you rebase this and then it should be good to merge?

src/args.rs
```diff
@@ -194,9 +202,20 @@ pub struct CliArgs {
...
-    /// Enable overriding existing files during file upload
-    #[arg(short = 'o', long = "overwrite-files", env = "OVERWRITE_FILES")]
-    pub overwrite_files: bool,
+    /// What to do if existing files with same name is present during file upload
+    ///
+    /// If you enable renaming files, the renaming will occur by
+    /// adding numerical suffix to the filename before the final
+    /// extension. For example file.txt will be uploaded as
+    /// file-1.txt, the number will be increased untill a available
+    /// slot is found.
```
svenstaro, line 211:
```suggestion
    /// What to do if existing files with same name is present during file upload
    ///
    /// If you enable renaming files, the renaming will occur by
    /// adding numerical suffix to the filename before the final
    /// extension. For example file.txt will be uploaded as
    /// file-1.txt, the number will be increased until an available
    /// filename is found.
```

README.md
```diff
@@ -289,6 +289,14 @@ Options:
 
           [env: OVERWRITE_FILES=]
 
+  -R, --rename-duplicate
+          Enable renaming files during file upload if duplicate exists
+
+          The renaming will occur by adding numerical suffix to the filename before the final extension. For example file.txt will be uploaded as file-1.txt, the number will
+          be increased untill a available slot is found.
+
+          [env: RENAME_DUPLICATE_FILES=]
```
svenstaro, line 298: This needs to be updated as it's not synced with the text in the `--help` output.

Atreyagaurav:
Hi, it looks like a lot of things were changed while doing the multipart file upload. Would have been nice to merge it before that. I'll have to look into the conflicts a little more deeply and understand the new process as well, do you think you can wait like a week for it?

tests/upload_files.rs
```diff
@@ -275,3 +278,177 @@ fn set_media_type(
...
+    Ok(())
+}
+
+fn write_file_contents(path: PathBuf, filename: &str, contents: &str) -> PathBuf {
+    let file_path = path.join(filename);
+    let mut file = File::create(&file_path).unwrap();
+    file.write_all(contents.as_bytes())
+        .expect("Couldn't write file");
+    file_path
+}
```
svenstaro, line 449: Do we really need this function and can't we use `std::fs::write()` instead?

tests/upload_files.rs
```diff
@@ -275,3 +278,177 @@ fn set_media_type(
...
+    let test_file_path = write_file_contents(
+        server.path().to_path_buf(),
+        test_file_name,
+        test_file_contents,
+    );
+
+    // Before uploading, make sure the file is there.
+    let body = reqwest::blocking::get(server.url())?.error_for_status()?;
+    let parsed = Document::from_read(body)?;
+    assert!(parsed.find(Text).any(|x| x.text() == test_file_name));
```
svenstaro, line 300: I think you can get rid of all of this in every test because you can use `fixtures::FILES` and try to overwrite them. They already exist and this should allow you to cut the tests by a lot. We use `fixtures::FILES` in many places in the test suite.

svenstaro (changes requested):

svenstaro:
> Hi, it looks like a lot of things were changed while doing the multipart file upload. Would have been nice to merge it before that. I'll have to look into the conflicts a little more deeply and understand the new process as well, do you think you can wait like a week for it?

Yeah, sorry about that. Life happened. Anyway, I can certainly wait a week. :)

svenstaro:
Hey, @Atreyagaurav, do you think we can push this over the finish line soon?

Atreyagaurav:
Since there were conflicts and other changes in master, I just redid everything on the new master and force-pushed. Should be fine. For the tests, I took the examples from other tests in the `tests/file_upload.rs` to make a file in the function itself, instead of using the fixtures. Might be better to redo all of them in a separate issue.

tests/upload_files.rs
```diff
@@ -268,6 +268,161 @@ fn uploading_files_to_allowed_dir_works(
...
+        .multipart(form)
+        .send()?
+        .error_for_status()?;
+
+    // After uploading, uploaded file is still getting listed.
+    let body = reqwest::blocking::get(server.url())?;
+    let parsed = Document::from_read(body)?;
+    assert!(parsed.find(Text).any(|x| x.text() == test_file_name));
+    // and assert the contents is the same as before
+    assert_file_contents(&test_file_path, test_file_contents_new);
```
svenstaro, line 367: I think these comments need some adjustment. Seems copy-pasted.

src/file_op.rs
```diff
@@ -109,13 +109,35 @@ pub async fn recursive_dir_size(dir: &Path) -> Result<u64, RuntimeError> {
...
+                    // format until available name is found (e.g. file-1.txt, file-2.txt, etc)
+                    let fp = if let Some(ext) = &file_ext {
+                        file_path.with_file_name(format!("{}-{}.{}", file_name, i, ext))
+                    } else {
+                        file_path.with_file_name(format!("{}-{}", file_name, i))
+                    };
+                    if !fp.exists() {
+                        file_path = fp;
+                        break;
+                    }
```
svenstaro, line 137:
```suggestion
                    };
                    // If we have a file name that doesn't exist yet then we'll use that. 
                    if !fp.exists() {
                        file_path = fp;
                        break;
                    }
```

tests/upload_files.rs
```diff
@@ -268,6 +268,161 @@ fn uploading_files_to_allowed_dir_works(
...
+#[case(server(&["-u", "-o", "error"]))]
+#[case(server(&["-u", "--on-duplicate-files", "error"]))]
+fn uploading_duplicate_file_is_prevented(#[case] server: TestServer) -> Result<(), Error> {
+    let test_file_name = "duplicate test file.txt";
+    let test_file_contents = "Test File Contents";
+    let test_file_contents_new = "New Uploaded Test File Contents";
+
+    // create the file
+    let test_file_path = server.path().join(test_file_name);
+    let _ = std::fs::write(&test_file_path, test_file_contents);
```
svenstaro, line 282: Might be nicer to do:
```suggestion
    std::fs::write(&test_file_path, test_file_contents)?;
```
here and in other tests.

svenstaro (changes requested):
I think we're super close now. Sorry for taking some time to re-review this.

svenstaro:
Whoops.

Atreyagaurav:
Thank you! 

Sorry I was in a meeting and couldn't rebase.

PR #1458, opened 2024-10-11
Small README improvements

lahwaacz:
Just a couple of things I found while setting it up. miniserve is great btw! :tada:


svenstaro:
Thanks!

PR #1471, opened 2025-02-05
Change font-weight from 300 to normal to improve readability

shaicoleman:
font-weight 300 is almost unreadable for me
image


svenstaro:
I like this better as well, thanks!

Issue #1424, opened 2024-05-23
Running with multiple replicas


ffzzhong:
hi, I plan to run miniserve in k8s with multiple replicas behind a nginx ingress controller, in this pattern, multiple replicas make the service more robust and there will be automatic load balancing inside the cluster.

I found if I run only 1 replica, everything works fine, but if I run with multiple replicas, say 2 or 3, then page sometimes rendered without CSS, it throws `Failed to load resource: the server responded with a status of 404 ()` because it's trying to look for the `https://my.service.domain/xxxxxxxx` static CSS file, I guess this path is only available on one replica but not all of them.

my entrypoint is straightforward and easy, just following the guide, `/app/miniserve /data-folder`

Am I missing some configuration? as I think it could be a typical use case that running multiple replicas, Any suggestion about how to make it work?

svenstaro:
The static CSS file is inside the binary. It's not a real file and as such should exist in all instances. miniserve is entirely stateless and should as such be an ideal server to run for multiple replicas. I'm actually pretty confused why it would sometimes work and sometimes not.

ffzzhong:
@svenstaro  yup i think so and I'm confused too, i'm using the latest miniserve image: https://hub.docker.com/layers/svenstaro/miniserve/latest/images/sha256-6bdbc5052be355dbe803080c3a67c07520c944107be9e58edc71944b038bf162?context=explore
![image](https://github.com/svenstaro/miniserve/assets/12509779/be3399ba-6566-4c48-804a-9e480bb8be43)

svenstaro:
Depending on what the issue and how you run it and how you're accessing it, it might help to use `--route-prefix`.

ffzzhong:
@svenstaro  actually I had already tried adding the --route-prefix, also configured ingress/livenessprobe/readinessprobe accordingly, the instances launched successfully but still got the same issue
the root cause might be like what @ahti mentioned?

svenstaro:
Yes, that's very likely the cause. I'll look at it soon.

PR #1472, opened 2025-02-06
move favicon and css to stable, non-random routes

ahti:
Favicon and stylesheet are currently served at random paths in order to avoid collisions. However, this causes problems with [running multi-replica deployments](https://github.com/svenstaro/miniserve/issues/1424) as well as caching, because the path is not stable across multiple instances/restarts of miniserve.

In order to address this, this PR moves the paths to stable routes prefixed by "__miniserve_internal", which should provide enough protection from accidental collisions.


svenstaro:
Fixes #1424.

PR #1473, opened 2025-02-06
Add date pill and sort links for mobile views

Flat:
Adds a mobile view pill for the relative date, and creates clickable links to sort via those pills.

It was annoying to have to turn my phone sideways to be able to sort by the hidden date and size columns, and I think this is a decent solution while keeping the overall mobile feel.


svenstaro:
Great idea, thanks!

PR #1483, opened 2025-03-09
Fix: size display for dirs with spaces in the name

Prajjwal:
DirSize requests for dirs with a space in the name contain a urlencoded name (eg. { DirSize: "/hello%20world" }), leading to a 400 Bad Request. The frontend then injects the error response body directly into the page.

1. Handle failures by returning a default string if this request fails.
2. Decode the payload before the request is issued.

**Before:**

image

**After, upon request failure:**

image

**After, with uridecoded payloads:**

image

## Notes

- Decoding this on the server along with a test is probably a better way of doing this.

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -682,9 +682,9 @@ fn page_header(
                             },
                             method: 'POST',
                             body: JSON.stringify({
-                                DirSize: dir
+                                DirSize: decodeURI(dir)
                             })
-                        }).then(resp => resp.text())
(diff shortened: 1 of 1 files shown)


svenstaro:
Hey, thanks for this. I went ahead and fixed this in the backend so now we can either take percent-encoded paths or plain paths. I used your line of code for JS-side error handling.

Prajjwal:
Excellent, thank you.

PR #1492, opened 2025-04-18
Adding an option to serve files from a different server

jankeymeulen:
Adding an option to specify an absolute hostname to serve the files from somewhere else.

In my usecase, I wanted to have an authenticated instance for browsing and an unauthenticated one to serve files (with indexing disabled). Instead of having to rewrite the URL when sharing a link, this option would allow this without manual intervention.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -377,9 +377,13 @@ pub struct CliArgs {
     #[arg(long, default_value_t = SizeDisplay::Human, env = "MINISERVE_SIZE_DISPLAY")]
     pub size_display: SizeDisplay,
 
-    /// Optional base URL (e.g., 'http://external.example.com:8081') to prepend to file links
-    #[arg(long = "file-base-url", env = "MINISERVE_FILE_BASE_URL")]
-    pub file_base_url: Option<String>,
+    /// Optional external URL (e.g., 'http://external.example.com:8081') prepended to file links in listings.
+    /// Allows serving files from a different URL than the browsing instance. Useful for setups like:
+    /// one authenticated instance for browsing, linking files (via this option) to a second,
+    /// non-indexed (-I) instance for direct downloads. This obscures the full file list on
+    /// the download server, while users can still copy direct file URLs for sharing.
+    #[arg(long = "file-external-url", env = "MINISERVE_FILE_EXTERNAL_URL")]
+    pub file_external_url: Option<String>,
 }
 
 /// Checks whether an interface is valid, i.e. it can be parsed into an IP address
--- a/src/args.rs
+++ b/src/args.rs
@@ -383,6 +383,8 @@ pub struct CliArgs {
     /// one authenticated instance for browsing, linking files (via this option) to a second,
     /// non-indexed (-I) instance for direct downloads. This obscures the full file list on
     /// the download server, while users can still copy direct file URLs for sharing.
+    /// The external URL is put verbatim in front of the relative location of the file, including the protocol.
+    /// The user should take care this results in a valid URL, no further checks are being done.
     #[arg(long = "file-external-url", env = "MINISERVE_FILE_EXTERNAL_URL")]
     pub file_external_url: Option<String>,
 }
--- a/src/args.rs
+++ b/src/args.rs
@@ -376,6 +376,10 @@ pub struct CliArgs {
     /// Show served file size in exact bytes
     #[arg(long, default_value_t = SizeDisplay::Human, env = "MINISERVE_SIZE_DISPLAY")]
     pub size_display: SizeDisplay,
+
+    /// Optional base URL (e.g., 'http://external.example.com:8081') to prepend to file links
+    #[arg(long = "file-base-url", env = "MINISERVE_FILE_BASE_URL")]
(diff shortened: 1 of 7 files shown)


svenstaro:
Good idea. I think most services call this "external URL" or "public URL" or something like that. What do you think?

jankeymeulen:
Yeah I wanted to stress out that this would only be applicable to the links to the files, maybe calling it --file-external-url then?

svenstaro:
Hm how does this interact with `--show-wget-footer`? Should that change it?

svenstaro:
> Yeah I wanted to stress out that this would only be applicable to the links to the files, maybe calling it --file-external-url then?

I like `--file-external-url`.

svenstaro:
It's a fairly niche use-case but I'm ok having this in. Could you try to add some more description to the help text as to why a user would even want to use this?

svenstaro:
Also please fix CI :)

jankeymeulen:
> Hm how does this interact with `--show-wget-footer`? Should that change it?

Indeed interesting point. The wget command would still point to the page on the "original" server and fetch the index pages from there, the links would point to the external server, and thanks to the "-H" option that is there already, be fetched from there. I think this behaviour makes sense?

svenstaro:
> > Hm how does this interact with `--show-wget-footer`? Should that change it?
> 
> Indeed interesting point. The wget command would still point to the page on the "original" server and fetch the index pages from there, the links would point to the external server, and thanks to the "-H" option that is there already, be fetched from there. I think this behaviour makes sense?

Makes sense. Did you test that behavior in practice? Would be good to validate assumptions.

svenstaro:
I'd also love a test showing this actually does anything. Should be pretty easy. :)

jankeymeulen:
Sure, you were totally right to ask for a test. It didn't work. :-) Changed the wget command to explicitely include the -H, as this is obviously needed. Changed the tests as well then and I had cargo fmt complain about something so changed that as well.

You can test on http://45.59.100.15:5555/ which is the indexing server and  http://45.59.100.15:55555/ which is the "external" server.

svenstaro:
Are you happy with this? I'll then give this another review.

jankeymeulen:
Yeah it does what it needs to for my usecase, and I'm happy that there are no more linters and tests complaining. :-D

src/args.rs
```diff
@@ -376,6 +376,15 @@ pub struct CliArgs {
...
     pub size_display: SizeDisplay,
+
+    /// Optional external URL (e.g., 'http://external.example.com:8081') prepended to file links in listings.
+    ///
+    /// Allows serving files from a different URL than the browsing instance. Useful for setups like:
+    /// one authenticated instance for browsing, linking files (via this option) to a second,
+    /// non-indexed (-I) instance for direct downloads. This obscures the full file list on
+    /// the download server, while users can still copy direct file URLs for sharing.
+    #[arg(long = "file-external-url", env = "MINISERVE_FILE_EXTERNAL_URL")]
+    pub file_external_url: Option<String>,
```
svenstaro, line 387: Maybe also document the behavior if given a relative such as `hello`. Is this even intended behavior?

svenstaro (changes requested):
We have some traversal attack tests already. I reckon we need to add some traversal attack tests for this new flag as well because the logic that builds the path isn't covered by the old set of tests.

svenstaro:
Should this flag conflict with `--enable-webdav`? Does it make sense to have both?

src/renderer.rs
```diff
@@ -303,9 +309,16 @@ fn wget_footer(abs_path: &Uri, root_dir_name: Option<&str>, current_user: Option
         None => String::new(),
     };
 
+    // Add the -H option to span hosts when serving files from another instance
+    let span_hosts_option = if file_external_url.is_some() {
+        " -H"
+    } else {
+        " -nH"
```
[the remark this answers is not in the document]
jankeymeulen: It's the default behaviour actually. Given how I was confused myself with wget params I'd thought to make things a bit more explicit.

jankeymeulen:
> We have some traversal attack tests already. I reckon we need to add some traversal attack tests for this new flag as well because the logic that builds the path isn't covered by the old set of tests.

I'll have to read up on those a bit then first, I'm not entirely sure I understand the problem. The code added here doesn't change anything to the actual file serving logic, only to the hyperlinks on a given page. Any traversal attack should be covered already by test covering the serving or files or dirs?

src/args.rs
jankeymeulen: Will add the following:
    /// The external URL is put verbatim in front of the relative location of the file, including the protocol.
    /// The user should take care this results in a valid URL, no further checks are being done.

jankeymeulen:
> Should this flag conflict with `--enable-webdav`? Does it make sense to have both?

I suppose 99% of the usecases of webdav for miniserve would be to mount as a filesystem, right? I'm trying hard to imagine why someone would want to use webdav in the scenario I'm proposing this change for, and I'm not even sure webdav (fs mounts) would support different hosts at all.

Would you be fine with just a note in both the --file-external-url and --enable-webdav args?

PR #1495, opened 2025-05-02
Fix --no-symlinks not filtering files and dirs nested in symlinks

ahti:
Currently, if you can guess the name of a symlinked dir, and the name of files or directories nested inside, miniserve will not prevent you from accessing them, even when `--no-symlinks` is passed.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -385,8 +385,23 @@ fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
             .prefer_utf8(true)
             .redirect_to_slash_directory()
             .path_filter(move |path, _| {
-                // deny symlinks if conf.no_symlinks
-                !(no_symlinks && base_path.join(path).is_symlink())
+                if !no_symlinks {
+                    // no_symlinks not enabled => nothing to filter
+                    return true;
+                }
+
+                // append path to base_path component by component and check for symlink at each step
+                let mut full_path = base_path.clone();
+                for component in path.components() {
+                    full_path.push(component);
+                    if full_path.is_symlink() {
+                        // path contains symlink component while no_symlink is active => filter
+                        return false;
+                    }
+                }
+
+                // path didn't include a symlink component => don't filter
+                true
             })
     };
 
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -385,8 +385,20 @@ fn configure_app(app: &mut web::ServiceConfig, conf: &MiniserveConfig) {
...
+                let mut full_path = base_path.clone();
+
+                for component in path.components() {
+                    full_path.push(component);
+                    if full_path.is_symlink() {
+                        return false;
+                    }
+                }
+
+                true
```
svenstaro, line 401: This logic is now complex enough that I'd like a few comments sprinkled in here.

tests/fixtures/mod.rs
```diff
@@ -45,6 +45,13 @@ pub static DEEPLY_NESTED_FILE: &str = "very/deeply/nested/test.rs";
 /// Name of a symlink pointing to a directory
 pub static DIRECTORY_SYMLINK: &str = "dir_symlink/";
 
+/// Name of a directory inside a symlinked directory
+#[allow(unused)]
+pub static SYMLINK_NESTED_DIR: &str = "dir_symlink/nested";
+
+/// Name of a file inside a directory inside a symlinked directory
+pub static SYMLINK_NESTED_FILE: &str = "dir_symlink/nested/file";
```
svenstaro, line 53: Maybe make the names even more obvious? `DIR_BEHIND_SYMLINKED_DIR` and `FILE_NESTED_BEHIND_SYMLINKED_DIR` perhaps.

tests/serve_request.rs
```diff
@@ -152,6 +152,33 @@ fn serves_requests_no_hidden_files_without_flag(server: TestServer) -> Result<()
...
+#[case(true, server(&["--no-symlinks"]))]
+fn serves_requests_nested_in_symlinks(
+    #[case] no_symlinks: bool,
+    #[case] server: TestServer,
+) -> Result<(), Error> {
+    let expected_status = if no_symlinks {
+        StatusCode::NOT_FOUND
+    } else {
+        StatusCode::OK
+    };
```
svenstaro, line 166: Put these into the `cases` so you don't have to have this conditional here. It also makes it more obvious as to what we expect in each case.

svenstaro (changes requested):
Looks good. Just a few nits.

PR #1502, opened 2025-06-24
fix filtering symlinks in webdav methods

ahti:
This allows `--enable-webdav` and `--no-symlinks` to be used together, and adds the necessary filtering.

### DIFF
--- a/src/webdav_fs.rs
+++ b/src/webdav_fs.rs
@@ -10,32 +10,77 @@ use dav_server::{
     },
     localfs::LocalFs,
 };
-use futures::{StreamExt, TryFutureExt, future::ready};
-use std::path::{Component, Path};
+use futures::StreamExt;
+use std::ffi::OsStr;
+#[cfg(target_family = "unix")]
+use std::os::unix::ffi::OsStrExt;
+use std::path::{Component, Path, PathBuf};
+use tokio::fs;
 
 /// A dav_server local filesystem backend that can be configured to deny access
 /// to files and directories with names starting with a dot.
 #[derive(Clone)]
 pub struct RestrictedFs {
     local: Box<LocalFs>,
+    base_path: PathBuf,
     show_hidden: bool,
+    no_symlinks: bool,
 }
 
 impl RestrictedFs {
     /// Creates a new RestrictedFs serving the local path at "base".
     /// If "show_hidden" is false, access to hidden files is prevented.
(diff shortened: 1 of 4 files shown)


src/webdav_fs.rs
```diff
@@ -10,83 +10,165 @@ use dav_server::{
...
-                        })
-                    }));
-                dyn_stream
-            }))
-        } else {
-            Box::pin(ready(Err(DavFsError::NotFound)))
-        }
+        Box::pin(async move {
+            if !self.is_path_allowed(path).await {
+                return Err(DavFsError::NotFound);
```
svenstaro, line 108: Maybe log something here so it's a bit easier to debug?

src/webdav_fs.rs
```diff
@@ -10,83 +10,165 @@ use dav_server::{
...
+        })
     }
 
     fn metadata<'a>(&'a self, path: &'a DavPath) -> DavFsFuture<'a, Box<dyn DavMetaData>> {
-        if !path_has_hidden_components(path) || self.show_hidden {
-            self.local.metadata(path)
-        } else {
-            Box::pin(ready(Err(DavFsError::NotFound)))
-        }
+        Box::pin(async move {
```
svenstaro, line 156: So many `pin`s D:

src/webdav_fs.rs
```diff
@@ -10,83 +10,165 @@ use dav_server::{
...
+                                let entry_path = base_path.join(os_string);
+                                if let Ok(md) = fs::symlink_metadata(&entry_path).await {
+                                    if md.file_type().is_symlink() {
+                                        return None;
+                                    }
+                                }
+                            }
+                            Some(Ok(e))
+                        }
+                        Err(e) => Some(Err(e)),
```
svenstaro, line 146: Hmmm is this right? Won't this result in weird entries? Also, should we log here?

src/webdav_fs.rs
```diff
@@ -10,83 +10,165 @@ use dav_server::{
...
         options: DavOpenOptions,
     ) -> DavFsFuture<'a, Box<dyn DavFile>> {
-        if !path_has_hidden_components(path) || self.show_hidden {
-            self.local.open(path, options)
-        } else {
-            Box::pin(ready(Err(DavFsError::NotFound)))
-        }
+        Box::pin(async move {
+            if !self.is_path_allowed(path).await {
+                Err(DavFsError::NotFound)
```
svenstaro, line 94: Maybe log something here?

svenstaro (changes requested):
Almost there!

Issue #62, opened 2019-04-08
Search bar ?


boastful-squirrel:
What would you think about having a search bar to search for files in the listing ? I think that would be pretty useful for mobiles, when you can't easily CTRL+F. If that's an idea you like, I think we should make it compatible with both Javascript and no-Javascript users. 

I was thinking:

The search bar has a "Search" button. When clicking on it, the searched term is put as query parameter `?search=foobar`, then the server generates a page corresponding to that filtering. That would be compatible with no-javascript user. If Javascript is enabled, we simply hide the search button, and the filtering is performed in live using Javascript.

svenstaro:
Not quite sure I like this. Especially the JS/non-JS part. Essentially that would have us implementing the same search (albeit not much code) server side and then in JS for optimum performance in both scenarios.

Also, in general, I'm not convinced we need a search. However, if you went ahead and whipped something up, I'd be happy to give it a spin and change my mind. I might not end up accepting that, though.

boastful-squirrel:
No I didn't start, I was waiting for your input. I'll close this then

PR #1505, opened 2025-07-07
Minimalist search/filter feature

7sDream:
I recently needed a minimalist HTTP static file server, and I found that miniserve fits my needs very well.

However, it lacks a feature to filter by filename, which I really need. Coincidentally, Rust is a language I've been enjoying recently, so I tried adding this feature: add a `?search=xxx` query in url will filter output items by their filename.

This code is written purely based on my personal needs. For example, I don't need a search bar on the page, query parameters are sufficient for me...

I've already looked at the discussion in #62 and understand that search might not be a feature you want to add, so this PR doesn't need to be merged, you can close it if you want. I created it just so that the next user who tries to search for this feature can find a usable fork/branch with this feature and can continue from here if they want.

Thank you again for your beautiful work.

===

Screenshot:

Without filter:
without filter

Without filter:
with filter

### DIFF
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -1,4 +1,4 @@
-use std::time::SystemTime;
+use std::{borrow::Cow, time::SystemTime};
 
 use actix_web::http::{StatusCode, Uri};
 use chrono::{DateTime, Local};
@@ -30,14 +30,18 @@ pub fn page(
     conf: &MiniserveConfig,
     current_user: Option<&CurrentUser>,
 ) -> Markup {
+    let (sort_method, sort_order, search) = (
+        query_params.sort,
+        query_params.order,
+        query_params.search.as_deref(),
+    );
+
     // If query_params.raw is true, we want render a minimal directory listing
     if query_params.raw.is_some() && query_params.raw.unwrap() {
-        return raw(entries, is_root, conf);
+        return raw(entries, search, is_root, conf);
     }
 
     let upload_route = format!("{}/upload", &conf.route_prefix);
-    let (sort_method, sort_order) = (query_params.sort, query_params.order);
-
     let upload_action = build_upload_action(&upload_route, encoded_dir, sort_method, sort_order);
     let mkdir_action = build_mkdir_action(&upload_route, encoded_dir);
 
@@ -85,7 +89,7 @@ pub fn page(
                                 // wrapped in span so the text doesn't shift slightly when it turns into a link
                                 span { bdi { (el.name) } }
                             } @else {
-                                a href=(parametrized_link(&el.link, sort_method, sort_order, false)) {
+                                a href=(parametrized_link(&el.link, sort_method, sort_order, false, search)) {
                                     bdi { (el.name) }
                                 }
                             }
@@ -142,7 +146,7 @@ pub fn page(
                                     td colspan="3" {
                                         p {
                                             span.root-chevron { (chevron_left()) }
-                                            a.root href=(parametrized_link("../", sort_method, sort_order, false)) {
+                                            a.root href=(parametrized_link("../", sort_method, sort_order, false, search)) {
                                                 "Parent directory"
                                             }
                                         }
@@ -150,7 +154,7 @@ pub fn page(
                                 }
                             }
                             @for entry in entries {
-                                (entry_row(entry, sort_method, sort_order, false, conf.show_exact_bytes))
+                                (entry_row(entry, sort_method, sort_order, false, search, conf.show_exact_bytes))
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


7sDream:
Sure, I can add some docs. But I have some other work on my hands, may not have time to write docs until tomorrow. Will ping you after finish.

As for the CI error, I think it's not introduced by this PR, the log shows: 

```
 ---- auth_multiple_accounts_pass::case_1 stdout ----
[miniserve stdout] miniserve v0.31.0
Error: rve stderr] Mon, 07 Jul 2025 17:46:33 +0000 [ERROR] Failed to bind server to [::]:45067
Error: rve stderr] Mon, 07 Jul 2025 17:46:33 +0000 [ERROR] caused by: Address already in use (os error 98)
[miniserve stderr] Error: Failed to bind server to [::]:45067
[miniserve stderr] caused by: Address already in use (os error 98)
```

Which to me it's like a CI enviroment issue, and the macOS CI is passed. Maybe ubuntu machine will be ok if we run it again.

7sDream:
@svenstaro 

I think this is ready for review.

I add an example in README, not sure if this is the way you expect to add docs. Let me know if there are other place to add docs.

7sDream:
Hello,

While using my branch today, I noticed that there are times when I want to filter only files while keeping all directories visible. This functionality is particularly useful when searching for files without knowing their exact folder location.

I’ve temporarily implemented the following syntax for filtering:

- `f:<keyword>` – Filters files only; all directories are displayed.
- `d:<keyword>` – Filters directories only; all files are displayed. (I’m unsure if this is widely needed, but I added it for symmetry.)
- `<keyword>` (without a prefix) – Filters both files and directories.

Here’s an example:

PixPin_2025-07-10_05-17-46

I’m not sure if this is a accpetable feature. if it is, should I include this in the current PR for you to review together?

Please let me know if you’re interested in this or have any feedbacks, thanks.

svenstaro (changes requested):
Cool work! Implemention looks fine, too.

What do you think about adding a simple search form to the index page? Perhaps it should be enabled by default and made opt-out?

I know you said
> This code is written purely based on my personal needs. For example, I don't need a search bar on the page, query parameters are sufficient for me...

but I think it would be a lot more discoverable and user friendly with a bar on the page.

svenstaro:
Sorry for taking a while to get to this! I will review this soon. Just pushed the v0.32.0 release. Could you rebase this?

7sDream:
Sure, rebased.

svenstaro:
Oh shit, yeah, sorry, I totally forgot about this. Would be glad to get this in. Thanks for the poke.

src/renderer.rs
7sDream: CI clippy lint error: too many arguments 😢

svenstaro: Eh fuggit, you can add a clippy exception here.

7sDream: Done

tests/serve_request.rs
```diff
@@ -45,6 +49,40 @@ fn serves_requests_with_no_options(reqwest_client: Client, tmpdir: TempDir) -> R
...
+#[case(server(None::<&str>), "keyword-that-matches-nothing", &[])]
+#[case(server(None::<&str>), ".hidden", &[])]
+#[case(server(&["--hidden"]), ".hidden", &[".hidden space dir/", ".hidden_dir1/", ".hidden_file1", ".hidden_file2"])]
+fn serves_requests_with_search_query(
+    #[case] server: TestServer,
+    #[case] search: &'static str,
+    #[case] result: &'static [&'static str],
+) -> Result<(), Error> {
+    let body = reqwest::blocking::get(format!("{}/?search={}", server.url(), search))?
+        .error_for_status()?;
```
svenstaro: We now have a `reqwest_client` fixture:
```suggestion
fn serves_requests_with_search_query(
    #[case] server: TestServer,
    #[case] search: &'static str,
    #[case] result: &'static [&'static str],
    reqwest_client: Client,
) -> Result<(), Error> {
    let body = reqwest_client(format!("{}/?search={}", server.url(), search))?
        .error_for_status()?;
```

tests/serve_request.rs
```diff
@@ -45,6 +49,40 @@ fn serves_requests_with_no_options(reqwest_client: Client, tmpdir: TempDir) -> R
...
+    "test \" \' & < >.csv",
+    "test.html",
+    "test.mkv",
+    "test.txt",
+])]
+#[case(server(None::<&str>), "d", &["dir space/", "dir_symlink/", "dira/", "dirb/", "someDir/", "😀.data"])]
+#[case(server(None::<&str>), "dira", &["dira/"])]
+#[case(server(None::<&str>), "keyword-that-matches-nothing", &[])]
+#[case(server(None::<&str>), ".hidden", &[])]
+#[case(server(&["--hidden"]), ".hidden", &[".hidden space dir/", ".hidden_dir1/", ".hidden_file1", ".hidden_file2"])]
```
svenstaro: Good tests and test cases!

Issue #124, opened 2019-05-23
Uploading files and writing with specific permissions and user/group


sadnblueish:
I have the same folder shared via Samba which requires specific user and file/directory permissions. If I upload through miniserve, it uses current user and creates file with 644 permissions. Anyway to change this behaviour to use "nobody.nogroup" and file permissions to 775?

svenstaro:
Good idea, yes. Do you feel up for implementing this?

sadnblueish:
Yes sure.

TTTPOB:
any progress on this feature?

svenstaro:
Would you expect it to be possible to choose the permissions on a per-upload basis or would it be fine to supply the desired permissions as a CLI flag once when starting miniserve initially?

TTTPOB:
> Would you expect it to be possible to choose the permissions on a per-upload basis or would it be fine to supply the desired permissions as a CLI flag once when starting miniserve initially?

 For me, I think the latter one would really help

lilydjwg:
I relied on umask to create files with the right permissions but https://github.com/svenstaro/miniserve/commit/413a63a60307bdf60229670b0f858963604d62a3 broke this. I guess I'll make a pr to support a `--chmod` option somehow.

PR #1506, opened 2025-07-08
Add a --chmod option to set file mode on uploaded files

lilydjwg:
see also #124.

I haven't looked into tests / docs yet. It works for me so I send it first to get early feedback.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -517,6 +517,7 @@ pub fn parse_header(src: &str) -> Result<HeaderMap, httparse::Error> {
     Ok(header_map)
 }
 
+#[cfg(unix)]
 pub fn parse_file_mode(src: &str) -> Result<u16, std::num::ParseIntError> {
     u16::from_str_radix(src, 8)
 }
--- a/src/args.rs
+++ b/src/args.rs
@@ -227,16 +227,16 @@ pub struct CliArgs {
 
     /// Set unix file permissions of uploaded files
     ///
-    /// This takes an octal number, for example 0600.
+    /// This takes an octal number, for example 0600. By default 0666 & ~umask is used to simulate
+    /// the system's default behavior.
     #[cfg(unix)]
     #[arg(
         long = "chmod",
         value_parser(parse_file_mode),
         env = "MINISERVE_CHMOD",
-        default_value = "0600",
         requires = "allowed_upload_dir"
     )]
-    pub chmod: u32,
+    pub chmod: Option<u32>,
 
     /// Enable recursive directory size calculation
     ///
--- a/src/args.rs
+++ b/src/args.rs
@@ -233,7 +233,8 @@ pub struct CliArgs {
         long = "chmod",
         value_parser(parse_file_mode),
         env = "MINISERVE_CHMOD",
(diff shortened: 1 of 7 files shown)


tests/upload_files.rs
```diff
@@ -535,3 +535,46 @@ fn assert_file_contents(file_path: &Path, contents: &str) {
     let file_contents = std::fs::read_to_string(file_path).unwrap();
     assert!(file_contents == contents)
 }
+
+#[cfg(unix)]
+#[rstest]
+#[case(server(&["-u"]), 0o600)]
+#[case(server(&["-u", "--chmod", "660"]), 0o660)]
+#[case(server(&["-u", "--chmod", "644"]), 0o644)]
+fn chmod(#[case] server: TestServer, #[case] expected_mode: u32) -> Result<(), Error> {
```
svenstaro, line 544: Seems like a good test but please add a doc comment explaining the intention. I'd also love to see a second test that tests the chmod error in case you don't have permissions for the chmod but I recognize it might be hard. If you don't find an easy way to test that, it's fine.

src/args.rs
```diff
@@ -225,6 +225,18 @@ pub struct CliArgs {
...
+    ///
+    /// This takes an octal number that is similar to the one the chmod use
+    #[cfg(unix)]
+    #[arg(
+        long = "chmod",
+        value_parser(parse_file_mode),
+        env = "MINISERVE_CHMOD",
+        default_value = "0600"
+    )]
+    pub chmod: u32,
```
svenstaro, line 238: This doesn't make make sense without `-u` right? So let's add `requires = "allowed_upload_dir",`

svenstaro (changes requested):
Hey, this is excellent! Make sure to also update the README, please.

src/args.rs
```diff
@@ -225,6 +225,18 @@ pub struct CliArgs {
...
 
+    /// chmod permissions of uploaded files
+    ///
+    /// This takes an octal number that is similar to the one the chmod use
+    #[cfg(unix)]
+    #[arg(
+        long = "chmod",
+        value_parser(parse_file_mode),
+        env = "MINISERVE_CHMOD",
+        default_value = "0600"
```
svenstaro, line 236: So far, the default behavior was that we uploaded with default permissions of the underlying system. This changes it and that's a problem. If this option is not set, we shouldn't also apply permissions that differ from the system's default.

lilydjwg: > So far, the default behavior was that we uploaded with default permissions of the underlying system.

This isn't the case since 413a63a60307bdf60229670b0f858963604d62a3. The file is created with 0600 since then (and it is a problem for me, thus this pr).

I can change it default to what umask says.

svenstaro: Oh crap. Well, looks like no one noticed. Might as well leave your change as it is, then.

svenstaro:
The CI failure looks legit. Wanna look into it?

lilydjwg:
I hope it works on all three platforms this time...

svenstaro:
This looks great! Thanks! Merging as-is. Sorry for taking some time on this.

Issue #1350, opened 2024-02-20
Wget footers :  hostname:port is repeated 


jgranduel:
The possibility to use wget for downloading a folder is a very valuable feature. I've tried `miniserve-0.26` today and I see that hostname:port is written twice,like below:

```
Download folder:
 'https://<IP>:8080https://<IP>:8080/folder/?raw=true'
```

I think, but I am not 100% sure, that it happens when I use autogenerated certificates
```
./miniserve --tls-cert miniserve.cert --tls-key miniserve.key  -W   /path/to/my/folder
````

**Comment by maxlivi** [comment]
Same here with  letsencrypt certicate:

wget -rcnHp -R 'index.html*' --cut-dirs=1 --ask-password --user 'user' 'https://IP:8080https://IP:8080/?raw=true'

The QR code has the same issue

**Comment by svenstaro** [comment]
Ah, this is odd. I don't really have the time to look into this right now. Would you like to see whether you can find and fix this?

**Comment by jgranduel** [comment]
Hi, 
thanks for your answer. I am not a Rust coder, but I tried to follow the functions calls. I think the error appears in `listing.rs` .
```
let abs_uri = {
    let res = Uri::builder()
        .scheme(req.connection_info().scheme())
        .authority(req.connection_info().host())
        .path_and_query(req.uri().to_string())
        .build();

    let abs_uri_string = req.uri().to_string();
    println!("listing abs_uri = {}", abs_uri_string);

    println!("listing scheme {}", req.connection_info().scheme().to_string());
    println!("listing host {}", req.connection_info().host().to_string());
    println!("listing uri {}", req.uri().to_string());

    match res {
        Ok(uri) => uri,
        Err(err) => return Ok(ServiceResponse::from_err(err, req.clone())),
    }
};
let abs_uri_string = abs_uri.to_string();
println!("listing abs_uri = {}", abs_uri_string);
```
I get :
```
git\miniserve> cargo build
   Compiling miniserve v0.27.1 (C:\git\miniserve)
    Finished dev [unoptimized + debuginfo] target(s) in 4.88s
    
git\miniserve> .\target\debug\miniserve.exe --tls-cert miniserve.cert --tls-key miniserve.key --auth-file ./authentications -z -W .
miniserve v0.27.1
Bound to [::]:8080, 0.0.0.0:8080
Serving path ...git\miniserve
Available at (non-exhaustive list):
    https://127.0.0.1:8080
    [...]

Quit by pressing CTRL-C
[...]

PR #1515, opened 2025-08-05
Fix duplicate url in footer when TLS enabled as mentioned in #1350

xtay573269555:
Duplicate url introduced by `abs_uri` in function `directory_listing`.


svenstaro:
Thanks!

PR #1561, opened 2026-04-19
Fix missing UTF-8 charset on text file responses

### LINKED ISSUE #1500: Specify `charset=utf-8` for not only plain text
I was trying to serve a javascript file and it had decoding problems. 

Previous issue #263 addressed this for plain text files `text/plain`; I think we should do the same for other text files like `text/javascript` or `text/markdown`.

lawrence3699:
Fixes #1500.

`actix-files` is currently only appending `charset=utf-8` to some text responses, so UTF-8 `.js` and `.md` files are still served without a charset while `.txt` already gets one.

This adds a small response middleware on the file-serving scope that appends `charset=utf-8` to `text/*` responses that don't already declare a charset.

Validation:
- cargo test served_text_files_include_utf8_charset --test encoding -- --nocapture
- cargo test --test serve_request
- cargo fmt --check

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,10 +7,15 @@ use actix_files::NamedFile;
 use actix_web::middleware::from_fn;
 use actix_web::{
     App, HttpRequest, HttpResponse, Responder,
+    body::MessageBody,
     dev::{ServiceRequest, ServiceResponse, fn_service},
     guard,
-    http::{Method, header::ContentType},
-    middleware, web,
+    http::{
+        Method,
+        header::{self, ContentType},
+    },
+    middleware::{self, Next},
+    web,
 };
 use actix_web_httpauth::middleware::HttpAuthentication;
 use anyhow::Result;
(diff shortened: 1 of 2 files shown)


svenstaro:
Is there an upstream bug report for this? I don't really want to fix this downstream if I don't have to.

pzhlkj6612:
> Is there an upstream bug report for this? ...

Didn't find it.

<br>

> `actix-files` is currently only appending `charset=utf-8` to some text responses ...

We have the `.prefer_utf8(true)` call:

https://github.com/svenstaro/miniserve/blob/eee8e19f39a92d6151d5dd5745e350aa36d565c3/src/main.rs#L411-L434

`pub fn prefer_utf8(mut self, value: bool)`: [actix-files/src/named.rs#L397-L404](https://github.com/actix/actix-web/blob/86eeea7f92cb14250a4529bb1938c1bfa26c0f27/actix-files/src/named.rs#L397-L404)

`pub fn into_response(self, req: &HttpRequest)` calls `equiv_utf8_text(...)`: [actix-files/src/named.rs#L477-L481](https://github.com/actix/actix-web/blob/86eeea7f92cb14250a4529bb1938c1bfa26c0f27/actix-files/src/named.rs#L477-L481)

Then, `pub(crate) fn equiv_utf8_text(ct: Mime)` only cares about part of MIME types: [actix-files/src/encoding.rs#L3-L40](https://github.com/actix/actix-web/blob/86eeea7f92cb14250a4529bb1938c1bfa26c0f27/actix-files/src/encoding.rs#L3-L40)

<br>

The function `prefer_utf8(...)` looks a bit misleading to me.

lawrence3699:
Closing this because the maintainer indicated they do not want to carry this as a downstream fix without an upstream bug report, and I do not have a confirmed upstream issue to anchor it.

svenstaro:
Well, I don't think we definitely shouldn't fix this here but I at least want to see an upstream discussion about it.

PR #1570, opened 2026-06-28
Implement `--inline` flag to serve files inline (`content-disposition:inline`)

### LINKED ISSUE #1564: `miniserve` currently does not support a way to override content-disposition headers
Hi, in some cases, I would like to serve a single `file.pdf`.

```bash
miniserve file.pdf -p <port>
```

To serve that without requiring the users to automatically download the pdf file, I would need to override the `Content-Disposition` header to `inline`. But currently, `miniserve` hard code that header to `attachment`.

```text
╭───┬─────────────────────┬───────────────────────────────────╮
│ # │        name         │               value               │
├───┼─────────────────────┼───────────────────────────────────┤
│ 0 │ content-length      │ 786257                            │
│ 1 │ last-modified       │ Tue, 05 May 2026 16:21:59 GMT     │
│ 2 │ etag                │ "0:bff51:69fa1927:1046a108"       │
│ 3 │ content-type        │ application/pdf                   │
│ 4 │ content-disposition │ attachment; filename="report.pdf" │
│ 5 │ accept-ranges       │ bytes                             │
│ 6 │ date                │ Tue, 05 May 2026 16:22:42 GMT     │
╰───┴─────────────────────┴───────────────────────────────────╯
```

I think the "headers can not be override, if they are previously set" seems intended, but, in some cases, I really need the override capability.

Can we work on a compensation?

ask4fusora:
Closes #1564.

# Description

This PR implements a new `--inline` command-line flag to serve files with `Content-Disposition: inline` instead of the default `Content-Disposition: attachment`.

This allows web browsers to render supported files (such as PDFs, images, or text files) directly in the browser instead of forcing a download.

- With `--inline`:
  ```bash
  http head http://localhost:8080/sample.pdf
  ╭───┬─────────────────────┬───────────────────────────────╮
  │ # │        name         │             value             │
  ├───┼─────────────────────┼───────────────────────────────┤
  │ 0 │ content-type        │ application/pdf               │
  │ 1 │ content-disposition │ inline; filename="sample.pdf" │
  │ 2 │ ...                 │ [truncated]                   │
  ╰───┴─────────────────────┴───────────────────────────────╯
  ```
- Default (without `--inline`):
  ```bash
  http head http://localhost:8080/sample.pdf
  ╭───┬─────────────────────┬───────────────────────────────────╮
  │ # │        name         │               value               │
  ├───┼─────────────────────┼───────────────────────────────────┤
  │ 0 │ content-type        │ application/pdf                   │
  │ 1 │ content-disposition │ attachment; filename="sample.pdf" │
  │ 2 │ ...                 │ [truncated]                       │
  ╰───┴─────────────────────┴───────────────────────────────────╯
  ```

### 1. CLI Definition (`src/args.rs`)
- Added the `--inline` boolean flag to `CliArgs`.
- Added mapping to `MiniserveConfig::inline` in `src/config.rs`.

### 2. Single-File Serving (`src/listing.rs`)
- In `file_handler`, if `conf.inline` is enabled, we clone the file's original `ContentDisposition`, change its `disposition` to `DispositionType::Inline`, and apply it back using `set_content_disposition`.

### 3. Directory Serving (`src/main.rs`)
- In `configure_app`, if `conf.inline` is enabled, we configure `actix_files::Files` using `.mime_override(|_| DispositionType::Inline)`.

### 4. Custom Header Conflict Warning (`src/main.rs`)
- Added a warning if any custom `Content-Disposition` header via `--header` are specified, informing automatic default takes precedence.

### 5. Test Suite Refactoring & Integration Tests (`tests/fixtures/mod.rs` & `tests/serve_request.rs`)
- Refactored the `server` fixture in `tests/fixtures/mod.rs` to be a simple, 1-line wrapper around a new, generic `server_with_path` fixture. This allows any test to serve a specific path relative to the temporary directory, keeping the test suite completely DRY and backward-compatible.
  - **Reason:** I searched, and found that there is no tests about single-file serving.
- Added `inline_header_set_directory` in `tests/serve_request.rs` to assert that directory serving correctly sets the `Content-Disposition` header to `inline` when `--inline` is passed.
- Added `inline_header_set_single_file` in `tests/serve_request.rs` to assert that single-file serving correctly sets the `Content-Disposition` header to `inline` when `--inline` is passed.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -383,7 +383,7 @@ pub struct CliArgs {
     ///
     /// Example:
     /// miniserve --inline sample.pdf
-    #[arg(short = 'i', long = "inline", env = "MINISERVE_INLINE")]
+    #[arg(long = "inline", env = "MINISERVE_INLINE")]
     pub inline: bool,
 
     /// Visualize symlinks in directory listing
--- a/src/args.rs
+++ b/src/args.rs
@@ -376,6 +376,16 @@ pub struct CliArgs {
     )]
     pub header: Vec<HeaderMap>,
 
+    /// Serve files inline in the browser instead of triggering a download.
+    ///
+    /// Useful when serving files that can be rendered directly by the browser (e.g., PDFs,
+    /// images, or text files).
+    ///
+    /// Example:
+    /// miniserve --inline sample.pdf
+    #[arg(short = 'i', long = "inline", env = "MINISERVE_INLINE")]
+    pub inline: bool,
+
     /// Visualize symlinks in directory listing
     #[arg(
         short = 'l',
--- a/src/config.rs
+++ b/src/config.rs
@@ -164,6 +164,9 @@ pub struct MiniserveConfig {
     /// Shown instead of host in page title and heading
     pub title: Option<String>,
 
+    /// If enabled, serve files inline with `Content-Disposition: inline`
+    pub inline: bool,
+
     /// If specified, header will be added
     pub header: Vec<HeaderMap>,
 
@@ -369,6 +372,7 @@ impl MiniserveConfig {
             zip_enabled: args.enable_zip,
             dirs_first: args.dirs_first,
             title: args.title,
+            inline: args.inline,
             header: args.header,
             show_symlink_info: args.show_symlink_info,
             hide_version_footer: args.hide_version_footer,
--- a/src/listing.rs
+++ b/src/listing.rs
@@ -4,7 +4,14 @@ use std::path::{Component, Path};
 use std::time::SystemTime;
 
 use actix_web::{
-    HttpMessage, HttpRequest, HttpResponse, dev::ServiceResponse, http::Uri, web, web::Query,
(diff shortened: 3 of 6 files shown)


svenstaro:
Did you use an LLM for this? If so, disclose the model and harness.

ask4fusora:
For code: Frankly sire, no, I didn't use LLMs for this. All code is figured out myself, line by line, word by word, test by test.

For PR body and commit messages: I did have AI assisted. It's Gemini 3.5 Flash, within Zed Editor. Nevertheless, I did review every line it generated.

Hope this helps.

ask4fusora:
I don't want to dirty my loved softwares with AI nonsenses.
And I do want to learn something on the way, especially in system development.

Thus, this PR is made wholely by hand. The code may be bad or oversighted in some way, but I guarantee no AI was used.

So, I would appreciate if I can have your dedicated review. Thank you.

PR #1572, opened 2026-07-03
Serve extension-less text files inline instead of downloading

quodlibetor:
`actix-files` guesses Content-Type solely from the file extension, so extension-less files like LICENSE-MIT or Dockerfile fall back to `application/octet-stream` and browsers download them instead of showing them.

This fix: add a response middleware that, only when a successful response is `application/octet-stream`, sniffs the first 8 KiB of the served file: valid UTF-8 with no null bytes is converted to `text/plain` with `content-disposition: inline` so that browsers display them instead of prompting for a download.

Files with an already recognized type continue to use the old behavior.

(written with claude code, then hand edited/verified.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- next-header -->
 
 ## [Unreleased] - ReleaseDate
+- Serve extension-less text files (e.g. `LICENSE-MIT`) inline instead of forcing a download by sniffing file contents
 - Fix suffix generation for pastes in non-HTTPS env [#1562](https://github.com/svenstaro/miniserve/pull/1562) (thanks @pzhlkj6612)
 
 ## [0.35.0] - 2026-04-08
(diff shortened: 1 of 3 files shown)


svenstaro:
Thanks! Is this something that could better be done upstream, I wonder?

quodlibetor:
maybe? it would probably belong here: https://github.com/actix/actix-web/blob/c6249cc00c757083ae1b29a3c4e1748185f43071/actix-files/src/named.rs#L107-L125

mime_guess is explicitly file-extension only, and putting this in actix would add a file read in the load path of *all* actix static servers which feels likely to be riskier and more likely to be denied, but I could give it a shot.

svenstaro:
Well, I thought perhaps a parameter for that would be good and the default behavior would be the same as currently so it'd be less dangerous of a change. I'm not against this change in miniserve FWIW but I think having a discussion with upstream might be good to see their opinion on this.

quodlibetor:
Very fair, I opened an issue in actix to ask if they'd welcome it: https://github.com/actix/actix-web/issues/4134

quodlibetor:
fwiw I don't really know what the actix merge and release timelines are like, if you're happy in principle with this change I wouldn't resent you for merging this and reverting it when/if Actix improves/changes their api. also happy to wait and see how the conversation goes.

PR #1573, opened 2026-07-04
fix: prevent archive downloads from following escape symlinks

### LINKED ISSUE #1568: miniserve ZIP archive download follows symlink target outside served root even with `--no-symlinks`
# miniserve ZIP archive download follows symlink target outside served root even with `--no-symlinks`

### Summary

The ZIP archive download path still follows symlinks and includes the target file contents in the generated archive. This also happens when miniserve is started with `--no-symlinks`, even though that option is documented as hiding symlinks and preventing them from being followed.

I fetched `origin/master` on 2026-06-18 and tested the latest upstream commit:

`925fcfe2c64575a364580788749f5645fde0873e` (`2026-06-01T18:40:31+02:00`)

### Details

Relevant code path:

- `src/args.rs` lines `163-165`: `--no-symlinks` is documented as "Hide symlinks in listing and prevent them from being followed".
- `src/listing.rs` lines `424-428`: the archive-download handler passes `conf.no_symlinks` into `archive_method.create_archive(...)`.
- `src/archive.rs` lines `237-249`: ZIP generation calls `std::fs::metadata(entry_path)`, which follows symlinks, and then checks `entry_metadata.file_type().is_symlink()`. Because the metadata is for the target, not the symlink itself, this check does not detect the symlink.
- `src/archive.rs` lines `274-289`: the same `entry_path` is opened with `File::open(&entry_path)` and written into the ZIP as a regular file entry, so the outside target's contents are packaged.
- `src/archive.rs` lines `148-155`: the TAR path also follows symlinks by default through `tar_builder.follow_symlinks(!skip_symlinks)`.

Root cause:

The archive generator treats a path inside the served directory as safe without validating what the path resolves to. In the ZIP path, `std::fs::metadata()` and `File::open()` both follow symlinks, so a symlink entry inside the served root can cause miniserve to read and package a file outside the served root. The `skip_symlinks` check is ineffective for ZIP because it is checking followed metadata rather than symlink metadata.

Related CVE reference used for triage: `CVE-2024-12718` in `python/cpython`, as the same root-boundary pattern where archive/path handling must verify that resolved paths stay inside the intended root.

### Reproduction

This reproducer uses a fresh checkout and starts miniserve with `--no-symlinks`:

```bash
set -euo pipefail

git clone https://github.com/svenstaro/miniserve.git miniserve-repro
cd miniserve-repro
git checkout 925fcfe2c64575a364580788749f5645fde0873e
cargo build --locked --no-default-features

tmp="$(mktemp -d)"
cleanup() {
  kill "${pid:-}" 2>/dev/null || true
  rm -rf "$tmp"
}
trap cleanup EXIT

mkdir -p "$tmp/served"
printf 'MINISERVE-OUTSIDE-SECRET\n' > "$tmp/outside-secret.txt"
ln -s ../outside-secret.txt "$tmp/served/link_to_outside"

./target/debug/miniserve \
  --enable-zip \
  --no-symlinks \
  -i 127.0.0.1 \
  -p 18181 \
  -- "$tmp/served" > "$tmp/server.log" 2>&1 &
pid=$!

for _ in $(seq 1 50); do
  curl -fsS http://127.0.0.1:18181/ >/dev/null 2>&1 && break
  sleep 0.1
done

curl -fsS -o "$tmp/download.zip" 'http://127.0.0.1:18181/?download=zip'

unzip -l "$tmp/download.zip" | grep 'served/link_to_outside'
unzip -p "$tmp/download.zip" served/link_to_outside
```

### Expected Behavior

With `--no-symlinks`, archive generation should not follow a symlink inside the served directory. More generally, archive downloads should not include file contents whose resolved path is outside the served root.

### Observed Behavior

The ZIP archive contains the symlink path as a regular file entry, and reading that ZIP entry returns the contents of the file outside the served root:

```text
served/link_to_outside
MINISERVE-OUTSIDE-SECRET
```

I also verified the default TAR download path with `--enable-tar` and no `--no-symlinks`; the generated TAR contains `served/link_to_outside` as a regular file and `tar -xOf` returns `MINISERVE-OUTSIDE-SECRET`.

### Impact

If a served directory contains a symlink to a file outside the shared root, a remote downloader can request a generated archive and receive the outside file's contents. This is especially surprising for ZIP downloads because the behavior still occurs when the operator starts miniserve with `--no-symlinks`.

### Suggested Fix Direction

- In ZIP generation, use `symlink_metadata()` or `DirEntry::file_type()` to detect symlinks before following the path.
- When `skip_symlinks` is true, skip symlink entries before calling `metadata()` or `File::open()`.
- For both ZIP and TAR archive generation, resolve symlink targets and reject or skip any entry whose resolved path is outside the served root.
- Add regression tests for archive downloads where the served directory contains `link_to_outside -> ../outside-secret.txt`, including a ZIP test with `--no-symlinks`.

yoda77777:
## Summary

Fixes #1568.

ZIP archive generation used `std::fs::metadata()`, which **follows** symlinks. The subsequent `is_symlink()` check therefore never saw a symlink, so `--no-symlinks` was effectively a no-op for ZIP downloads. A symlink planted inside the served tree that pointed outside it would have its target file contents packaged into the ZIP.

### Changes

- Detect symlinks with `DirEntry::file_type()` (does not follow) before deciding whether to open/package an entry.
- When `--no-symlinks` / `skip_symlinks` is set, omit symlink entries entirely.
- Regardless of that flag, **never package file contents whose resolved path escapes the directory being archived** (ZIP, TAR, and TAR.GZ).
- Walk TAR trees ourselves (with the tar builder’s own `follow_symlinks` disabled) so the same containment rules apply; broken/unresolvable links are skipped instead of aborting the whole stream.
- In-root symlinks are still followed when symlinks are allowed (covered by a regression test).

### Test plan

- [x] `cargo test --test archive` — all 22 cases pass, including new regressions for escape symlinks with and without `--no-symlinks`
- [x] `cargo fmt --check` / `cargo clippy -- -D warnings`
- [x] Full unit + archive/auth integration tests pass locally (bind URL tests flaked on this machine due to unreachable Tailscale IPv6 addrs, unrelated to this change)

### Reproducer (from #1568)

With a served dir containing `link_to_outside -> ../outside-secret.txt` and `--enable-zip --no-symlinks`, `/?download=zip` previously included the secret contents under `link_to_outside`. After this change the escape link is omitted and no outside secret appears in the archive.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -73,6 +73,7 @@ tls = ["rustls", "rustls-pemfile", "actix-web/rustls-0_23"]
 [dev-dependencies]
 assert_cmd = "2"
 assert_fs = "1"
+libflate = "2"
 predicates = "3"
 pretty_assertions = "1.2"
 regex = "1"
@@ -80,7 +81,9 @@ reqwest = { version = "0.13", features = ["blocking", "multipart", "json", "rust
 reqwest_dav = { version = "0.3", default-features = false }
 rstest = "0.26"
 select = "0.6"
+tar = "0.4"
 url = "2"
+zip = { version = "8", default-features = false }
 
 [target.'cfg(not(windows))'.dev-dependencies]
 # fake_tty does not support Windows for now
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 use std::fs::File;
 use std::io::{Cursor, Read, Write};
 use std::path::{Path, PathBuf};
@@ -9,6 +10,43 @@ use tar::Builder;
 use zip::{ZipWriter, write};
 
 use crate::errors::RuntimeError;
+
+/// Whether `path` is equal to or nested under `root`.
+///
+/// Both paths are expected to already be canonicalized so that symlink
+/// components and `..` segments cannot fool a prefix check.
+fn is_within_root(path: &Path, root: &Path) -> bool {
+    path.starts_with(root)
+}
+
+/// Decide what to do with a symlink encountered while building an archive.
+///
+/// - When `skip_symlinks` is true, the entry is ignored entirely.
+/// - When following is allowed, only targets that resolve **inside**
+///   `canonical_root` are included. Targets that escape the archive root, and
+///   broken links that cannot be resolved, are omitted so a single bad entry
+///   cannot abort (or poison) the whole archive download.
+enum SymlinkAction {
+    /// Omit the entry from the archive.
+    Skip,
+    /// Follow the symlink; `resolved` is the canonical target path.
+    Follow { resolved: PathBuf },
+}
+
+fn symlink_action(entry_path: &Path, canonical_root: &Path, skip_symlinks: bool) -> SymlinkAction {
+    if skip_symlinks {
+        return SymlinkAction::Skip;
+    }
+
+    match entry_path.canonicalize() {
+        Ok(resolved) if is_within_root(&resolved, canonical_root) => {
+            SymlinkAction::Follow { resolved }
+        }
+        // Outside the archive root, or unresolvable (broken link / permissions):
+        // never package the target contents.
+        _ => SymlinkAction::Skip,
+    }
+}
 
 /// Available archive methods
 #[derive(Deserialize, Clone, Copy, EnumIter, EnumString, Display)]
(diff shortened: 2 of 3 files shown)


svenstaro:
Thanks for this! Do you think this could be resolved by adding onto the respective upstream packages?

quart27219:
Hi, I'm wondering if there is any plan for merging this fix.
It’s been almost a month since the PR was submitted, and I’m curious why it hasn’t been merged.

svenstaro:
@quart27219 It's 500 lines of changes and I'm doing this in my spare time. Also I'm waiting on @yoda77777 to get back to me on my question because I'd really rather have this upstream than in my codebase if possible.

Issue #1459, opened 2024-10-31
Feature request: Live reload


keybits:
Would you consider adding optional live reloading of files when they change?

I currently use [devd](https://github.com/cortesi/devd) for this. It's similar to miniserve in that it's a single binary for serving files over http. It has a live reload option which is very handy for local web development.

devd works fine, but development has stopped and miniserve has some features not available in devd. It'd be great to use just one tool.

svenstaro:
I suppose you don't only want to live reload the file tree but also the contents of file. I'm not really sure how that would work. We can't live reload everything and we'd need to inject something into pages to make it work?

keybits:
> we'd need to inject something into pages to make it work?

Yes, that's [how devd works](https://github.com/cortesi/devd?tab=readme-ov-file#livereload)

Since it's optional, it only happens when live reloading is enabled with sometthing like `-l` in the command invocation.

For filesystem watching devd uses it's own Go filesystem watching library [moddwatch](https://github.com/cortesi/moddwatch).
In the Rust world, I believe [notify](https://github.com/notify-rs/notify) would be a good option as it's battle-tested and cross plaform.

No problem if this is out of scope and would bring in too many dependencies.

svenstaro:
I'm not opposed to this if it doesn't introduce too much new code. If you like, you can throw together a quick PoC and I can take a look at that with no promises made as to its inclusion.

PR #1575, opened 2026-08-01
Add live reload option

Bnyro:
Hey, first of all, thanks for your work on `miniserve`! I've used it for at least 2 years now for testing websites and never looked back to Python's HTTP server.

This PR adds support for live reload as discussed in #1459. The idea is very similar to devd's behavior that was described in https://github.com/svenstaro/miniserve/issues/1459#issuecomment-2449482575. We basically spawn a WebSocket connection between the browser and Miniserve that notifies the browser each time a file changed. For CSS files, only the CSS is reloaded, not the whole page (again, inspired by devd).

The JavaScript injection is done in an actix middleware - I would prefer to hook into `actix-files`, e.g. something like `Files::override_file_handler`, but `actix-files` doesn't have any way to do it.

This is currently only implemented for serving directories, not individual files, because I had some issues with actix' router, i.e. I didn't find a clean way to register the actix routes for both without duplicating code. Anyways, I think that the feature is not that important for individual files and it's probably easy to fix (help appreciated).

Feedback appreciated!

Cheers


src/live_reload.rs
```diff
@@ -0,0 +1,136 @@
...
+) {
+    let (tx, mut rx) = mpsc::unbounded_channel();
+    let event_fn = move |res: notify::Result<notify::Event>| {
+        if let Ok(event) = res
+            && (event.kind.is_create() || event.kind.is_modify() || event.kind.is_remove())
+        {
+            let reload_event = if event
+                .paths
+                .iter()
+                .all(|p| p.to_string_lossy().contains(".css"))
```
Bnyro: If you wonder why we do `contains` instead of `ends_with` here: The [Helix editor](https://helix-editor.com) I use and many other editors don't modify the existing file when writing. Instead, they create a copy of the existing file, delete the original file, write the new file, and then delete the copy from the original file. As a result, it also changes `style.css.bak` files, which don't end with `.css`.

If you prefer to use `ends_with` that's understandable, just unfortunate for me then :)

src/live_reload.rs
```diff
@@ -0,0 +1,136 @@
...
+    let headers = resp.headers().clone();
+    let status_code = resp.status();
+    let content_type = resp.headers().get(CONTENT_TYPE);
+
+    if let Some(content_type) = content_type.and_then(|c| c.to_str().ok())
+        && content_type == ContentType::html().0
+    {
+        let body = body::to_bytes(resp.into_body())
+            .await
+            .map_err(|_| ErrorInternalServerError("failed to convert body to bytes"))?;
```
Bnyro, line 103: I don't think this error should ever happen, we could have unwrapped here I think - but I think it's better to handle it gracefully anyways.

svenstaro:
Hey, thanks for this. I'll take some time to review all of this since it's a lot of code. However, why did you choose to go with WebSockets as opposed to SSE (which we are already using)?

Bnyro:
> Hey, thanks for this. I'll take some time to review all of this since it's a lot of code. However, why did you choose to go with WebSockets as opposed to SSE (which we are already using)?
>
I didn't know/see that we already use SSE. I agree, SSE might be better because we don't need bidirectional communication for this. I'm open to switching the PR over if you want to, shouldn't be very difficult given that the file observing and JavaScript injection code is the same.

I think SSE is not used in the master branch yet, only in your PR: https://github.com/svenstaro/miniserve/pull/1487. So I'll also need to add the actix labs dependency here.

Bnyro:
Switched to SSE now, took longer than expected but works now.

src/live_reload.rs
```diff
@@ -0,0 +1,120 @@
...
+            let data = Data::new(event.to_string()).event("reload");
+            (
+                Ok::<Event, Infallible>(Into::<Event>::into(data)),
+                (watcher, rx),
+            )
+        })
+    });
+
+    // send keep alive to prevent browser from closing the connection
+    Ok(Sse::from_stream(stream).with_keep_alive(Duration::from_secs(5)))
```
Bnyro, line 55: No idea which keep alive duration makes sense, I just took 5 seconds because that's been in the actix web lab docs example.

